PR #662, opened 2012-05-15
bash: add inputrc

stefansaraev:


MindTooth:
I believe this to be a bit excessive.  Can't get word jumping to work properly.  Source for this `.inputrc`?

PR #2385, opened 2013-06-16
added mpd addon package

prg318:
Hello,

I have written an mpd (http://www.musicpd.org/) addon for openELEC.  By default, this addon will use the contents of "~/music" as the music library and will output audio to the default sound device (which can be changed by adjusting the mpd.conf in "~/.xbmc/userdata/addon_data/service.multimedia.mpd/config/mpd.conf".

MPD requires the libid3tag package which I have included in "packages/audio/libid3tag".

I would love to see this mpd addon in the openELEC add on repository so that users from all architectures that want to use mpd on their openELEC system can do so without needing to recompile everything.

I have submitted a previous pull request for this.  The previous pull request was not on a named branch, so I have opened a new pull request using a named branch.  I've closed the old pull request: https://github.com/OpenELEC/OpenELEC.tv/pull/2122

Let me know if you have any questions regarding this.  Cheers!

### DIFF
--- a/packages/audio/libid3tag/build
+++ b/packages/audio/libid3tag/build
@@ -1,0 +1,49 @@
+#!/bin/sh
+
+################################################################################
+#      This file is part of OpenELEC - http://www.openelec.tv
+#      Copyright (C) 2009-2012 Stephan Raue (stephan@openelec.tv)
+#
+#  This Program is free software; you can redistribute it and/or modify
+#  it under the terms of the GNU General Public License as published by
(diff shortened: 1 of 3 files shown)


Janghou:
MPD works fine on a RPI and RPI2. Great. Thx.

Unfortunately it does not play .mpc (musepack) or .ape (Monkey's Audio files). 
AFAIK for that it needs mpcdec and ffmpeg. 
Can that be added? 
( I tried to build myself, but it failed on a faad2: no package.mk file found)

PR #2581, opened 2013-09-13
Re-allow specifying architecture-specific patches (lost in 997dd43f8)

garbear:
The script tests for arch-specific patches, but the for loop no longer enumerates them.


stefansaraev:
I'll just close this :D

PR #2736, opened 2013-11-03
Added patch for sql_lex.h

peol:
This enables us to compile MySQL with server support (i.e. remove `--without-server` `--without-embedded-server`) in `package.mk`.

This is useful for people wanting to use the MySQL daemon on the Raspberry Pi for example.

However, this commit should be seen as a discussion topic, since the environment variable `YACC_HEXT_HH` seems to be set, but it doesn't build the .hh file (builds the regular .h file) so this patch pretty much works around that by hardcoding the `sql_yacc.h`-reference instead.


lrusak:
Is this still relevent?

peol:
I've stopped compiling OE myself as I moved the SQL database elsewhere. It did work the last time I tried, granted it was roughly a year ago.

I got no access to a build environment nowadays, I'm afraid.

PR #2798, opened 2013-12-04
create_installstick: umount all partitions and inform the user

LS80:
Small change to create_installstick with a way to ensure that all the partitions are unmounted, just in case there are more than 10 (admittedly, this is very unlikely!)

This also adds a message for each umount.


stefansaraev:
create_installstick is gone now

PR #2802, opened 2013-12-05
Support a USB storage partition in create_sdcard script

LS80:
This adds an option to the RPi create_sdcard script for setting up the storage partition on a USB stick, following the method at http://openelec.tv/forum/124-raspberry-pi/67688.

It also contains some other minor changes to improve the script.


stefansaraev:
create_sdcard is gone now

PR #3038, opened 2014-03-22
[RFC] disable suspend if booting from removable storage

stefansaraev:
suspend/resume is miserably broken (by design) when booting from removable storage (usb sticks). lets not allow xbmc to suspend/autosuspend

consoder the following scenario:
- user boots from usb stick
- user tries to suspend
- here usb host got reset. bye bye /dev/sda
- resume. /dev/sda is no more. bye bye /storage
- as a side effect - udevil would mount /dev/sda2 as /var/media/Whatever

with this commit, there will be NO "suspend" entry in xbmc shutdown menu. shutdown function CAN NOT be set to "suspend" and is forced as "shutdown"

needs testing on RPi, also need to ensure if it doesnt break something with netboot (this I can test myself next monday).

also, it has been reported several times that resume with NBD/ISCSI boot is broken. let me know if we need to disable suspend also for netboot. EDIT: suspend with nfs is quite fine after we decreased default retrans,timeo options, just some noticeable 5-10 sec delay on resume.

### DIFF
--- a/packages/sysutils/busybox/scripts/init
+++ b/packages/sysutils/busybox/scripts/init
@@ -588,6 +588,25 @@
 
   BOOT_STEP=final
 
+  # no suspend when booted from nonpersistent storage
+  STORAGE=$(cat /proc/mounts | grep " /sysroot/storage " | awk '{print $1}' | awk -F '/' '{print $3}')
+  if [ -n "$STORAGE" ] ; then
+    removable="/sys/class/block/*/$STORAGE/../removable"
+    if [ -e $removable ] ; then
+      if [ "$(cat $removable 2>/dev/null)" = "1" ] ; then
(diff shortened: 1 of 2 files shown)


Jan1503:
Hi and sorry for this late comment.
I've just installed Kodi on my new cubox (iMX6 build) and would like to use the suspend-function as well but unfortunately it's greyed out.

I'm a newbie on linux, tried your instructions here, but there is no /dev/.suspend_disabled-file so I assume the option should be enabled but it's not :) Maybe the cpu isn't capable of suspending?

Please give me a quick hint on how to enable suspend on this little box if possible.
Thanks and regards, Jan

stefansaraev:
your platform does not support suspend at all.

Issue #3111, opened 2014-04-18
systemd does not wait for /storage/.config/shutdown.sh


Grobwiefein:
Hi,

in the new OpenELEC release, systemd does not wait for /storage/.config/shutdown.sh to be fully executed. Something like

echo "Sleeping..." >> $HOME/msg
sleep 5
echo "Finished!" >> $HOME/msg

is interrupted during execution.

Is there a way to fix this behaviour? I’m using .config/shutdown.sh to switch off my VDR server remotely.

PR #3118, opened 2014-04-21
Modified before units of systemd in xbmc-(halt|poweroff|reboot) services.

Grobwiefein:
This commit should fix the bug described in #3111.


00joshi:
May it be that the problem prevails / resurfaced in "OpenELEC (official) Version: 5.0.8". Im trying to "cat ... | nc ...". Works when I execute the script but does not work when I poweroff.

PR #3172, opened 2014-05-04
Further improvements to wake on lan script

andreasmn:
Modified the wake on lan script to reduce waiting time if the share comes available before the wol_wait time is expired.
wol_wait is now used to set the max time we want to wait until we stop trying.


chewitt:
I'm closing this PR as it's ~19 months out of date

PR #3246, opened 2014-05-21
Avoid download an already retrieved source

Diesis:
Check existing MD5 to ensure the already downloaded source is not corrupt


chewitt:
I'm closing this down as last update is ~16 months ago. Thanks.

PR #3631, opened 2014-11-20
add the opportunity to disable wireless.

mrdominuzq:
as the title says.

i dont use wireless on my setup, so this gives me and other the opportunity to disable it.


mrdominuzq:
after thinking over this, i close this :)

PR #3633, opened 2014-11-21
Updating libftdi to libftdi1

phoefler:
This PR changes the rather old libftdi library to the actual new one.
all three dependencies of libftdi compile successfully with this commit, but I could only test lirc.
I don't have any LCD nor I have no glue what flashrom is for (it's not even used in Generic build if I see it correctly ;-) )


stefansaraev:
ping @sraue
+1 from me. if it doesnt break lcdproc. master only

sraue:
added (a bit different) in https://github.com/OpenELEC/OpenELEC.tv/commit/0e73cf3bdf4ad534a952d05fa830fd9438e08cbd - thanks much and sorry for the delay! :-)

PR #3656, opened 2014-12-02
projects/*/linux: enable crypto modules for NFSv4

Alphix:
These modules are necessary for supporting Kerberized
NFSv4 clients. They're pretty small (8K, 11K, 49K on my
system).

The actual addon with the necessary software tools is at:
https://github.com/Alphix/unofficial-addons/tree/nfs4-client


sraue:
added in https://github.com/OpenELEC/OpenELEC.tv/commit/b4bbefca6187546a235a9994cc270f70a658b3dd thanks much

PR #3657, opened 2014-12-02
[rfc] kodi crashlogs

stefansaraev:
on ubuntu, when kodi crashes, it dumps a nice crashlog. crashlogs help us and kodi devs to find out WHY..

the kodi.sh script included in this PR is a shameless rip off, cleaned up a bit ;)

I have tested on x86_64 and so far all good (simulating crash via kill -SEGV). but this needs testing. on RPi/Cuboxi-i with slow sdcard it might take quite lot of time to dump a 300+ MB core. also sh*t may happen on netbooted clients (but we shouldnt care much for those, anyway) EDIT: oh, and those people who install x86_64 on usb sticks, too...

please dont merge (yet), this is for after OE 5 and should not go to 5.0 branch.

@sraue any comments?

//cc @fritsch


stefansaraev:
now. after https://github.com/xbmc/xbmc/pull/5895 we have a problem. a big one, that will affect lot of stuff (not only this PR).

we will need:
http://sprunge.us/HJgV
or:
http://sprunge.us/YZEL

both tested and work. I'd prefer first one, but it's ugly... (XDG_DATA_HOME= )

the .xbmc -> .kodi move was enough pain. this xdg nonsense (sorry for the wording) will be even more pain.. so up to you. ping @sraue

stefansaraev:
ok, xbmc/5895 was reverted. IMO we are good to go. when (or if..) it is added back upstream, we will take care, it is unrelated to this PR.

@sraue any objections? this has been well tested and should be safe (for master only)

sraue:
yeah please commit if you are sure its ok, on master i will stick with helix for some time or readd support for building master or helix based on make options. master will become 5.2 soon (with kernel 3.18/19, systemd 218, python 2.7.9 and mesa 10.4)

Issue #2607, opened 2013-09-16
Xbox 360 wireless controller not working anymore (since OE 3.2)


bhabba:
Since I've upgraded from Openelec (ION x64 build) version 3.0.6 to 3.2 my Xbox 360 wireless controller doesn't work anymore. This is obviously a regression because there was a similar issue in the past which had been fixed in version 3.0 RC3 (see https://github.com/OpenELEC/OpenELEC.tv/issues/1580).

The controller is connected through the common USB receiver (third party replica of the official microsoft one): http://support.xbox.com/en-US/xbox-360/accessories/xbox-360-wireless-gaming-receiver-windows

musyne:
Hi,

It's still not working on a Raspberry Pi B+.
I have tried both 5.0 official and the build from @zzattack.
However, I tried 5.0 official on a x86 (NUC DN2820FYKH) and it works fine (although the lights keep blinking) so I guess the bug is specific to the ARM build.
Anything I can do to help investigate the issue?

bonds:
Awesome @zzattack, thanks for working on this! I'm excited about this ticket. :)

yene:
Hi,
I can confirm Xbox 360 controller does not work on Raspberry Pi 2 Model B. I also have wireless xbox 360 controller and a xbox one controller if you want me to do some testing.

jbreitbart:
I have a wireless and a wired XBox 360 controller on for my Pi2. If there is anything to test I would like to help!

lrusak:
@yene @jbreitbart No joystick support in Kodi on the RPi. Please search

PR #3672, opened 2014-12-07
Update valve's xpad rework patch

zzattack:
@lrusak [originally ](http://openelec.tv/forum/20-development-discussion/72352-openelec-test-build-xbox-360-controllers) included this patch which combines some of valve's steamos kernel work on the xpad driver, so that xbox 360 wireless controllers work a bit more nicely. His patch got outdated when xbox one controller support was added in 3.17 so the patch had to be redone.

I've removed https://github.com/zzattack/OpenELEC.tv/commit/8132b0a4f6390fe9324c2b1eeba69c4c08552274 because it kills support for chinese controllers with VID/PID 0xFFFF. Not sure why it this patch was originally added, but I actually own such a controller that needs this :P

In https://github.com/zzattack/OpenELEC.tv/commit/70c76826d1ed04ed503427c20876b8750332cae1 I've also first added a patch to update xpad to the 3.18 version so that hopefully the actual patch in https://github.com/zzattack/OpenELEC.tv/commit/437092d2d536e7d253e78bbb1d19f98e7c5eca03 will remain valid for just a tad longer.

Some positive test results were reported in #2607.


sraue:
@zzattack can you look at this for kernel 3.18 we now have included?

zzattack:
I can rebase again but the problem seems to be with the patch itself. Not
sure if the original author will fix it, but I spoke with Greg who
initially submitted the work to linux-input, and he will revisit the
patches soonish. Might be best to just wait for that. Whether or not the OE
patch in its current state is worth including for now, I'll leave for you
to decide. I'll rebase tomorrow at any rate.

sraue:
added in #3933

PR #3712, opened 2014-12-23
Add a project for WeTek Play TV box

codesnake:


stefansaraev:
news on this? I would like to start testing.
and btw projects/WeTek.Play/patches/linux/g18dtd.patch does not apply

codesnake:
Increased max CPU frequency to 1.5GHz.
@stefansaraev What kernel you are trying to apply the g18dtd.patch to? I have just rebuilt the whole project without any problems.

stefansaraev:
now after you updated its fine. thanks.

stefansaraev:
ouh. it seems linux-amlogic-3.10-b992de0 ships pre-compiled 32bit x86 arch/arm/boot/mkimage. that is not good on 64bit multiarch hosts.

codesnake:
@stefansaraev Currently WeTek Play project doesn't use mkimage at all. It appends device tree to the kernel image.

codesnake:
@stefansaraev Sorry, confused mkimage with mkbootimg.

stefansaraev:
:) well I have installed libc6-bin:i386 and fine. can live with it now. but it will be a problem in the future.

we may need this. http://sprunge.us/BCUQ

systemd-218 failed here without __USE_KERNEL_XATTR_DEFS defined.

codesnake:
@stefansaraev On Debian/Ubuntu you can also install the u-boot-tools package, which contains mkimage.

stefansaraev:
sure, that would work. but needs one more dependency in scripts/depends.
what about compiling tools/ from mainline u-boot? would that be too much work just for wetek play.

codesnake:
Sure, I will add the u-boot package for WeTek Play in future. The bad thing that it requires yet another toolchain to be able to compile ARC firmware.

stefansaraev:
for now #3778 is enough. I'll merge if you agree with it

stefansaraev:
could you please pick last 3 commits from https://github.com/stefansaraev/OpenELEC.tv/commits/wetek and we are ready to go (unless @sraue wants to rename the project?)

sraue:
yes i would rename it to "WeTek_Play" at least...

stefansaraev:
right. we may have problem with our settingsaddon/updates because of the dot. so better rename yep.

offtopic: I have fixed keyboard layouts in settings addon, but I'll push it after vpndrop... then 5.0 is frozen. and this PR should go in master only. not 5.0.

projects/WeTek.Play/options
```diff
@@ -0,0 +1,158 @@
...
+    PROJECT_CFLAGS=""
+
+  # LTO (Link Time Optimization) support
+    LTO_SUPPORT="yes"
+
+  # GOLD (Google Linker) support
+    GOLD_SUPPORT="yes"
+
+  # SquashFS compression method (gzip / lzo / xz)
+    SQUASHFS_COMPRESSION="lzo"
```
stefansaraev: any reason to not keep it gzip? gzip is a good size/speed compromise. lzo is slower

codesnake: No reason, I will change to gzip. Thanks.

stefansaraev: wait. I'll do some tests fitst ;)

stefansaraev: uhm. target image (SYSTEM) is ~ 10% smaller with gzip, no difference in boot times. with both lzo and gzip = ~ 13s +/- 100ms.

but looks like I am wrong. lzo is a bit faster at runtime.

with gzip:
```
Wetek:~ # echo 3 > /proc/sys/vm/drop_caches; dd if=/usr/lib/kodi/kodi.bin of=/dev/null
...
18597832 bytes (17.7MB) copied, 0.634658 seconds, 27.9MB/s
18597832 bytes (17.7MB) copied, 0.720829 seconds, 24.6MB/s
18597832 bytes (17.7MB) copied, 0.687909 seconds, 25.8MB/s
18597832 bytes (17.7MB) copied, 0.784210 seconds, 22.6MB/s
```

with lzo
```
...
18597832 bytes (17.7MB) copied, 0.542292 seconds, 32.7MB/s
18597832 bytes (17.7MB) copied, 0.549667 seconds, 32.3MB/s
18597832 bytes (17.7MB) copied, 0.583566 seconds, 30.4MB/s
18597832 bytes (17.7MB) copied, 0.558868 seconds, 31.7MB/s
18597832 bytes (17.7MB) copied, 0.617791 seconds, 28.7MB/s
18597832 bytes (17.7MB) copied, 0.594130 seconds, 29.9MB/s
```

so better keep lzo. I swear I got different results one year back when I last tested on x86. I must re-test that again.. ;)

sorry for the noise.

stefansaraev:
@codesnake FYI:

@sraue reported g18dtd.patch failed to apply on his buildhost (fine here but he has different distro different git/patch..)

it turned out to be some aml-specific files in their kernel with dos line endings. this is a big no-no. so I re-packaged and re-uploaded linux-amlogic-3.10-b992de0.tar.xz to sources.openelec.tv/devel

make sure to remove your sources/linux and convert dvbwetek.patch and g18dtd.patch to unix format (dos2unix) or pick from my wetek branch.

sraue:
added in https://github.com/OpenELEC/OpenELEC.tv/commit/b49c8b7aff28bfa0d76a3789afb87b7c2084c11e and renamed to WeTek_Play. thanks soooo much!!!! great work :-) now a bit finetuning :-) feel free to PR updates etc.

codesnake:
Thank you!

PR #3715, opened 2014-12-24
Intel: Fixes for HSW GT1

fritsch:
It seems the HSW GT1 (Pentium 3230 and so on) have a physical limit concerning the number of wm threads, which causes Render hang and media hangs.

Those patches need testing and should go into OE 5.1 first or 5.0.1 if we find some testfolks for this.

I have tracked those down with OpenELEC user zeeeh and the intel dev bwidawsky. Without those patches the system of zeeh crashes within some minutes.

PS: As I am not able to build anything here, please someone could verify that I created the Mesa patches directory, etc. correctly.

History: https://bugs.freedesktop.org/show_bug.cgi?id=87564


stefansaraev:
@fritsch status of this? is it safe?

fritsch:
Nope, note at all ... keep it for reference.

PR #3723, opened 2014-12-27
add vdr-plugin-epgfixer

glenvt18:
A very valuable plugin to deal with ugly epg data broadcasted by providers.
An example of that ugly epg:
an EIT event has short_text and description tags and both can be used as an event description.
But for some events only short_text is broadcasted, for some only description or the text is
splitted between them. It mostly depends on EIT table id type - present/following or schedule.
E.g "Champions League. Schalke - Real Madrid." may be shown in KODI for some events as 
"Schalke - Real Madrid." or  "Champions League." for other which is very annoying. This
particular issue can be fixed in a skin (all skins with PVR support) but there will still be
other epg errors - you never know, it depends on your provider. So, to my view, it's better to
give a user an opportunity to correct them with this plugin. It's small, depends only on
vdr and pcre and works fine for me.

Change PKG_URL in packages/multimedia/vdr-plugin-epgfixer/package.mk if you put it to http://sources.openelec.tv.

### DIFF
--- a/packages/addons/service/multimedia/vdr-addon/package.mk
+++ b/packages/addons/service/multimedia/vdr-addon/package.mk
@@ -24,7 +24,7 @@ PKG_ARCH="any"
 PKG_LICENSE="GPL"
 PKG_SITE="http://www.openelec.tv"
 PKG_URL=""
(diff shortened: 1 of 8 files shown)


stefansaraev:
thanks much and happy NY :)

PR #3728, opened 2014-12-29
linux: add patch for Geniatech T230 USB DVB-T/T2/C tuner

trsqr:
This patch adds support for Geniatech T230 USB DVB-T2/T/C tuner [0572:c688] that is also sold as August T210v2 or MyGica T230. It's currently the lowest cost DVB-T2 tuner that's working reliably. The driver is already in media_tree and will be in kernel 3.19. Before that the following changes are needed:

si2168: add support for inverting the TS clock
si2157: add support for the Si2148 tuner (uses same firmware as Si2158)
cxusb: add the device specific functions
em28xx: initialize si2168_config with zeroes before using


trsqr:
The patch has been confirmed working by users on RPi, x64 and i386 architectures: http://openelec.tv/forum/71-pvr-live-tv/74370-build-with-geniatech-t230-support

PR #3740, opened 2015-01-02
use default imageres and remove busydialogdelayms

jester-xbmc:
the default for imageres is 720, I suggest we increase to this res, should still be ok for 256mb Pi (really buy a new one for 2015...)
remove busydialogdelayms as it was removed since Gotham
http://kodi.wiki/view/Advancedsettings.xml#Recently_removed_tags

### DIFF
--- a/projects/RPi/kodi/advancedsettings.xml
+++ b/projects/RPi/kodi/advancedsettings.xml
@@ -3,11 +3,7 @@
   <showexitbutton>false</showexitbutton>
 
   <fanartres>720</fanartres>
-  <imageres>540</imageres>
-
-  <video>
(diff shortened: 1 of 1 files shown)


popcornmix:
Not sure the wiki is correct about busydialogdelayms:
https://github.com/xbmc/xbmc/blob/master/xbmc/settings/AdvancedSettings.cpp#L707
https://github.com/xbmc/xbmc/blob/master/xbmc/cores/dvdplayer/DVDPlayer.cpp#L628

jester-xbmc:
Ahh, thanks, I'll see if I can edit the wiki

PR #3741, opened 2015-01-02
ntfs-3g_ntfsprogs: fix gcrypt check in autoreconf

stefansaraev:
AM_PATH_LIBGCRYPT is undefined if libgcrypt is not (yet)
built, and we dont depend on gcrypt...


stefansaraev:
ping @sraue any objections?
building ntfs-3g is probably fine on regular builds, as it's compiled AFTER libgcrypt (pulled by afpfs-ng, libaacs, libbdplus, systemd etc..) but not if you do scripts/build ntfs-3g_ntfsprogs manualy after make clean. I noticed the error while compiling truecrypt (unofficial addons)

PR #3747, opened 2015-01-03
connman: permanently disable ipv6

### LINKED ISSUE #3746: Enabling IPv6 leads to network Flooding
I'm using OpenELEC 5.0 on an Intel Celeron NUC and on an older Zotac ZBox. As soon as I enable IPv6 (mode auto) my Network gets unusable as OpenELEC is flooding the net with DHCPv6 Requests like this one:
3794078	613.279746000	fe80::c23f:d5ff:fe65:fd5b	ff02::1:2	DHCPv6	120	Solicit XID: 0x2017bf CID: 000100011c3af7a6c03fd565fd5b 
This is reproducible, disabling IPv6 directly stops this behaviour.
Gateway is a FritzBox7390 with a Sixxs IPv6-Tunnel.

stefansaraev:
closes #3746

users who want to re-enable it, can override via .config/sysctl.d/

for OE 6.0 ipv6 should be permanently removed from kernel and OE settings addon.

### DIFF
--- a/packages/network/connman/sysctl.d/ipv6.conf
+++ b/packages/network/connman/sysctl.d/ipv6.conf
@@ -1,0 +1,1 @@
+net.ipv6.conf.all.disable_ipv6=1


chewitt:
IPv6 is not going to go away so we should understand and resolve the problem. Disabled by default is fine but IMHO removed completely is a wrong move.

stefansaraev:
k. then please do.

stefansaraev:
not needed after 01bf6480e8a0a13089f41ed28a9055295066baa0

PR #3751, opened 2015-01-04
Verify downloaded sources against a SHA256 checksum.

takeyourhatoff:
Sources were previously downloaded over HTTP and FTP then built without
validating their integrity. This is dangerous not only because the final build
may be manipulated by an attacker but also because during the build an attacker
could gain arbitary code execution during the configuration stage of most
packages on the machine doing the build.

In addition to increased security, this prevents package maintainers modifying
sources without bumping package numbers which could cause confusing bugs.

Checksums are stored in package.mk files in the variable $PKG_SHA256. I have
provided initial checksums for all packages which require it (excluding virtual
packages). Checksums can be automatically updated by running `$ make checksums`.
Which modifies in place all package.mk files, downloading files as appropriate.
New packages must have the PKG_SHA256 variable declared. A person adding a new
package may then use the `$ make checksums` command to fill it in automatically.


stefansaraev:
minus 1. we trust our sources. this would make my life harder ;)

takeyourhatoff:
It is not a matter of trusting your sources, somebody could manipulate the files in tranist.

vpeter4:
I think you are just little to paranoid.

MilhouseVH:
Can't help but agree on the paranoid comment. That's not to say this is a bad idea, I'm just not convinced it's required and can see it causing more headaches while not really solving any problem that currently exists (sure, there are _potential_ problems).

However if this is ever added, please include an option to disable __all__ package checksum checks - I frequently update local sources (mkpkg etc.) when self-building - overwriting the current tar/url/md5 - and don't want to then have to update the corresponding package.mk just so that my newer local source package isn't rejected.

takeyourhatoff:
MillhouseVH, files are only validated when they are downloaded. Once they are downloaded you are able to modify the sources without any complaints.

Regarding the paranoia. As this does not add any significant difficulties to the build process I believe it is a worthwhile change. The verification takes only seconds and is only run once files are downloaded from the network. I certianly did not expect the build script to download files over a plaintext connection and then execute shell scripts inside the downloaded packages.

MilhouseVH:
> MillhouseVH, files are only validated when they are downloaded. Once they are downloaded you are able to modify the sources without any complaints.

Thanks for the clarification.

vpeter4:
One option would be to have checksum optionally. If checksum is set then compare it and if not work as now.
Regarding data changed in transit: I don't see this important. Would really like to see one case when man in the middle changed data. For openelec I mean.
One better use case would be when file is removed from server and you get index.html for example. But even in this case build process is interrupted because the file extraction fails.

jimrollenhagen:
It's insane to think you can trust the network in this day and age.

AFAICT, these sources are used in the build process. OpenELEC ships binaries, presumably built by this build process. You're putting not only your build machines, but all of your users at risk by trusting a plaintext connection to provide the code you expect.

takeyourhatoff:
> Regarding data changed in transit: I don't see this important. Would really like to see one case when man in the middle changed data. For openelec I mean.

You would be lucky to ever find out it happened.

vpeter4:
> You would be lucky to ever find out it happened.
Doesn't this contradicts your first post a little?

> Why would the checksum ever not be set?
Because when doing/testing some new package I really don't want to mess with checksum also. Check MilhouseVH comment.

stefansaraev:
@jimrollenhagen 
> It's insane to think you can trust the network in this day and age.

insane or not. I do. if you cant trust your network or your isp. you have a bigger problem than this.

> AFAICT, these sources are used in the build process. OpenELEC ships binaries, presumably built by this build process. You're putting not only your build machines, but all of your users at risk by trusting a plaintext connection to provide the code you expect.

speaking of that. releases are hosted on openelec servers, too. theoreticaly, man in the middle could "force" users to download (or update to) some bad build with god knows what inside.. again this pr changes exactly nothing but makes my life harder.

@takeyourhatoff you missed a change to scripts/get. forgot to git add ?

takeyourhatoff:
> speaking of that. releases are hosted on openelec servers, too. theoreticaly, man in the middle could "force" users to download (or update to) some bad build with god knows what inside.. again this pr changes exactly nothing but makes my life harder.

Automatic updates are downloaded without authentication? I will patch that too, because yes, that is an even bigger problem.

> @takeyourhatoff you missed a change to scripts/get. forgot to git add ?

Github just does not display all changed files because there are so many of them, I assure you that it is changed: https://github.com/takeyourhatoff/OpenELEC.tv/blob/verify-sources/scripts/get#L71 .
I recomend you look at this on your machine rather than on the Github website.

stefansaraev:
> Automatic updates are downloaded without authentication? I will patch that too, because yes, that is an even bigger problem.

and how "authentication" should be done ?

what would you do for people who download release tarbals manualy ?

takeyourhatoff:
> and how "authentication" should be done ?

It would be helpful to serve your website over https for starters but I was thinking about embeding a public key in the source code and having all updates signed by the corasponding private key.

> what would you do for people who download release tarbals manualy ?
You could post SHA256 sums on your website which hopefuly you start serving over HTTPS.

stefansaraev:
> You could post SHA256 sums on your website which hopefuly you start serving over HTTPS.

we use mirrorbrain. https is not going to happen.

takeyourhatoff:
How would you feel about signing relases for updates? I think this is very important.

stefansaraev:
I would not sign at all. neither I would rewrite settings addon to check signatures. and we dont do major-to-major upgrades anyway, for a good reason. users do that manualy. 99% of the users dont care for sha1sums even if they are published somewhere. ssl on (most) OE mirrors wont happen.

and well that would require some changes to the server infrastructure, too. at least switching to ssl for update.openelec.tv. I am not going to do that just because few people are too much paranoid. sorry

and well. remember the diginotar incident? ssl is not panacea.

takeyourhatoff:
It is understandable that you feel as though ssl is impossible right now. I am offering to modify the settings addon to check signitures. Signing would be automatic as part of the normal build process.

MilhouseVH:
What about custom builds, will they have to be signed and whose public key will be used to authenticate them when upgrading from a release build to a custom build?

stefansaraev:
> It is understandable that you feel as though ssl is impossible right now. I am offering to modify the settings addon to check signitures. Signing would be automatic as part of the normal build process.

it doesnt "fix" the original problem (mitm).

> What about custom builds, will they have to be signed and whose public key will be used to authenticate them when upgrading from a release build to a custom build?

correct.

vpeter4:
Don't complicate things which works fine now.

takeyourhatoff:
@MilhouseVH I guess I would only verify updates for `official` builds. Would that solve that problem?

> it doesnt "fix" the original problem (mitm).

It goes a long way. Sure there is the potential for a users first download to be modified and the validation checks to be stopped, but as it stands now, anybody at a priviliged point on the network of a user of OpenELEC can trigger an update at any time and own the device.

takeyourhatoff:
Shall we move discussion of the update mechanism here: https://github.com/OpenELEC/service.openelec.settings/issues/57 and continue to disuss the patch for downloading sources?

fritsch:
Here is my suggestion (for releases only):
Every build we do officially, can be signed with the official openelec
private key. Every user can verify this with our public key. Then we know
at least which versions the user is running.

2015-01-05 13:21 GMT+01:00 takeyourhatoff <notifications@github.com>:

> Shall we move discussion of the update mechanism here:
> OpenELEC/service.openelec.settings#57
> <https://github.com/OpenELEC/service.openelec.settings/issues/57> and
> continue to disuss the patch for downloading sources?
>
> --
> <https://github.com/OpenELEC/OpenELEC.tv/pull/3751#issuecomment-68701114>.
>



-- 
                   Key-ID:     0x1A995A9B
                   keyserver: pgp.mit.edu

chewitt:
There are currently 438 packages in the OE build-system. To create an authenticated image requires us to validate all of those sources; one un-auththenticated package effectively compromises the security of the system. The nature of the project and how we obtain and use sources (which in-turn is dependent on how a smörgåsbord of upstream projects distribute themselves) currently prevents us from achieving an authenticated build. As such I think this is something to be mindful of for the future but it is not something we currently need to concern ourselves with. Building signing for distribution is different and that I would support; as per comments made in the other thread.

takeyourhatoff:
In case what I said before wasn't clear, the signing provides the
authentication property. By signing a release, you are asserting that it
was created by you. Not necessarily that you have personally examined all
upstream sources,which of course is not feasible. What this then achieves
is the possibility to provide an automatic update facility that is not
susceptible to man in the middle attacks. It would still be possible for a
package maintainer to release bad code which you do not spot and therefore
gets into the official build, but if my first patch is applied it would be
difficult for a man in the middle to change the source a package maintainer
is sending you.

These two improvements do not add perfect security by any means, but
prevent a wide range of serious attacks which have been carried out in the
past. As an example, malware called finfisher used the  unauthenticated
apple iTunes automatic update mechanism to infect users machines (
http://m.spiegel.de/international/germany/a-799259.html). This is the exact
problem that faces OpenELEC and release signing would protect your users
from this attack. A media centre may not seem like a lucrative target for
attackers but at the very least it will be part of the users local network
and may become a stepping stone for a more serious attack.

takeyourhatoff:
On 5 Jan 2015 20:37, "Christian Hewitt" <notifications@github.com> wrote:
>
> I think you need to separate the process of securing builds we create for
distribution from the informal process of users (and staff) self-building
something from sources without inconvenience. Here's my suggestion:
As I said before, the checks only apply when files are being downloaded. I
can't imagine a situation where the checks on that would need to be
disabled. That said, effectively what happens when make check is run is
that the files are downloaded without authentication, and then the
checksums in the package.mk files are regenerated according to the newly
downloaded files. Ideally these checksums would then be corralled with the
same command run by other developers, especially for a final release.
>
> If an upstream source package is distributed with checksums the package.mk
should contain a URL in a PKG_CKSUM variable that is used for obtaining the
checksum. In most cases it will be the same as PKG_URL but with an extra
suffix. The build-system should verify the checksum against the file
downloaded before caching the source file and checksum file (with
appropriate .suffix, not all files use MD5 and the scheme should be made
obvious) and the build process should halt in the event of a mismatch. If
PKG_CKSUM is null or missing from package.mk verification is skipped and an
MD5 sum of the file downloaded is stored as we already do. This approach
places the burden of checksum creation upstream on the original publisher
without impact to developers, checksums are federated with original sources
over many servers which further reduces risk of compromise to supply-chain,
is backwards compatible, and the ability to locally fake a checksum is
retained.
I don't understand why you would have a url to download the checksum as
opposed to including the checksum in the first place. More importantly, if
the checksum were download it too could be manipulated in transit and would
defeat the purpose of the whole thing. Again, not quite understanding the
need to be able to 'fake a checksum'.
>
> If a package is created and hosted by the OE team (as we do for a number
of git based sources) we should assume responsibility for publishing a
checksum with the package. It's an easy addition to the mkpg scripts to
generate a SHA256 hash and copying two files to the sources server instead
of the current one file is an easier habit to learn.
Same problems as before, If the package can be modified in transit, and the
checksum is transmitted over the same channel, then the checksum can only
be used for detecting accidental corruption.
>
> Improved assurance of packages in the supply-chain for signed builds is
achieved through a one-time removal of cached packages on the build server;
from that point forward we are building with more verified packages than
before. Self-builders or developers can repeat the same should they have
any concerns about package authenticity. Over time packages with PKG_CKSUM
will increase in quantity until a majority are being verified for signed
release builds.
>
> I'm not going to lose too much sleep about finfisher style attacks as the
actors behind those campaigns won't be too obstructed with any
countermeasures we put-up, but I've investigated several targeted
waterholing and supply-chain attacks in the past 18-months so it's not
beyond comprehension. I'd still support parking this for a while as I
consider sorting the forums out with our limited project resources to be a
higher priority for the project, but if you consider the number of active
OE hosts and current rate of growth (we added ~29k systems since Christmas
Eve) we have a node count that rivals some of the larger crimeware botnets
so it's not something to be ignored forever.

MindTooth:
Self explanatory.  :+1: from me.

With todays security issues, this should be a must.  And at least when we know for a fact that OpenELEC is used to view, and or download "pirated" content.  Don't want MPAA/RIAA snooping on us would we?

fritsch:
Lol - for those it's enough to have your IP in the torrent tracker ... the
rest is told by your provider.

2015-01-24 15:23 GMT+01:00 Birger J. Nordølum <notifications@github.com>:

> Self explanatory. [image: :+1:] from me.
>
> With todays security issues, this should be a must. And at least when we
> know for a fact that OpenELEC is used to view, and or download "pirated"
> content. Don't want MPAA/RIAA snooping on us would we?
>
> --
> <https://github.com/OpenELEC/OpenELEC.tv/pull/3751#issuecomment-71319114>.
>



-- 
                   Key-ID:     0x1A995A9B
                   keyserver: pgp.mit.edu

MindTooth:
Whatever.. I stand by my comment.

stefansaraev:
your comment does not make sense. this PR has nothing to do with pirated content and so. neither with MPAA/RIAA.
and we dont care for pirated content or whatever OE users want to download. if pirating is illegal in their country, it is illegal and they are on their own. zero fucks given if they choose to break the law. period. and again this pr has nothing to do with pirated content.

I stand my comment too, this PR is a no-go as-is. if @sraue wants to sign release tarbals/images - sure that would be fine, we dont need a PR for that.

MindTooth:
I see my point used an extreme example.  But checking sources is always good.  Wether for injections, tampering, etc.  Just a safety layer for checking that you get what you expect.

SmileyJames:
+1 for merge.

eklavya:
I am just a user but seeing the attitude of the maintainers towards security being "we couldn't care less", I think I should be looking towards (better secured) alternatives like osmc. Of course it's a community effort and nobody owes anyone anything but at least being concerned about the basic security of your users is warranted. This seems to be the only open source project I have come across where security is not even in the agenda.

I apologize if I have offended someone as that was not my intent. Whether the PR is required or not I can't comment but the unwillingness shown here to even try doesn't inspire confidence and is frankly immature.

lrusak:
@eklavya and does osmc or anyone else verify all there packages used with a checksum?

It's not that we aren't security minded, we are just somewhere in the middle. OpenELEC really is secure if you leave ssh access disabled and even then you can just disable password logins. We give you the option make it secure. Remember that this is a mediacenter distribution, not a fileserver, webserver, desktop OS or otherwise.

eklavya:
@lrusak osmc is based off of debian and it uses the same update/install infra, so yes it's all verified against checksums.

You are correct openelec is a mediacenter distribution but it's an internet enabled distribution. If your build contains malware a LOT of people will be affected as it's a very very popular distribution. Let's say you are very careful while building the images and no malwares went through. You also started providing images over https (you really really should). You still have to worry about updates.

My point is, it's not a nice to have, it's the bare minimum.

stefansaraev:
@eklavya

> You are correct openelec is a mediacenter distribution but it's an internet enabled distribution

no mediacenter should  be DIRECTLY exposed to internet (or via dmz). one who cares about security, knows this very well.

> If your build contains malware a LOT of people will be affected as it's a very very popular distribution

I am more concerned about malware, that can be distributed via 3rdparty kodi repos

> Let's say you are very careful while building the images and no malwares went through.

we are.

> You also started providing images over https (you really really should).

read above why it's no. I wont repeat myself. 

> My point is, it's not a nice to have, it's the bare minimum.

in current form. this has -1 from me. read above why. that wont change.

lrusak:
@eklavya Yes OSMC is built off debian, but they do compile their own packages too. None of these packages are signed either (for example the linux kernel and Kodi).

The source is there, go have a look ;)

eklavya:
@stefansaraev well, good luck. In this day and age if someone says a media consuming device should not be connected to internet they are kidding themselves. I fail to understand why that particular mirror provider is the only option when even github gives you free release management. It's good if you don't want to repeat yourself, the problems will go away magically, I am sure.

@lrusak oh?  I assumed since it's based on debian. Thanks for taking the time to investigate. I will take a look. They should also do something about that. I really like it and I wish openelec could get more serious about this. However the style of the answers of some contributors here reflect a sense of elitism and "we know better" attitude which won't help anyone.

stefansaraev:
> should not be connected to internet they are kidding themselves

read, carefuly, what I wrote.

> I fail to understand why that particular mirror provider is the only option when even github gives you free release management

that must be some joke.

you didnt read the whole thing above, did you ?

> github gives you free release management

that seems to be one more joke.

so, what are you waiting for, take over this PR and make it acceptable so it can be merged.

eklavya:
@stefansaraev yes mighty lord, you are right. Forgive this weak one. It's obviously all a joke. But your highness the best joke was yours, about me wasting my time with a PR here. I will stop wasting your precious time right now lord.

fritsch:
No a place for polemic spamming of user email boxes.

PR #3764, opened 2015-01-09
[RFC] allow addons to ship profile.d scripts

stefansaraev:
this was requested by @ksooo

most of the potential issues I can think of are fixed already in bf42102

addons are not allowed to override anything that is already set via system profile.d/ scripts.

however, a rogue addon could "sleep 100000" thus effectively preventing stuff sourcing /etc/profile from starting. but addons can do other bad things as well, anyway. if one does such sh** he should be beaten.

pls dont merge. lets discuss it first and then decide if we want it.

### DIFF
--- a/packages/mediacenter/kodi/profile.d/00-addons.conf
+++ b/packages/mediacenter/kodi/profile.d/00-addons.conf
@@ -16,6 +16,13 @@
 #  along with OpenELEC.  If not, see <http://www.gnu.org/licenses/>.
 ################################################################################
 
+# addons profile.d/*.profile
+for config in /storage/.kodi/addons/*/profile.d/*.profile; do
+  if [ -f "$config" ] ; then
+    . $config
+  fi
+done
+
 oe_setup_addon() {
   if [ ! -z $1 ] ; then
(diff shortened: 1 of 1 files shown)


ksooo:
:+1:

vpeter4:
I would do very generic approach. Even allow such bad things. Why? Because it can be handy.

ksooo:
Sorry, don't quite get what you mean. Is your statement a pro or a con?

vpeter4:
It is +1. But seo talks about bad things addon can do. If it does it's addon's fault. From what I see in the patch it allows this. So it is ok for me.

ksooo:
@vpeter4 Ah. Thanks for clarifying.

stefansaraev:
@vpeter4 this one of the bad things that can interfere badly with the core system. thats why I mention it ;)

stefansaraev:
@sraue for signoff. this is tested by me, works, and safe (for now)

ksooo:
Tested by me as well. Works just fine.

ksooo:
Thx.

 Viele Grüße
Kai Sommerfeld.

PR #3767, opened 2015-01-10
Add sleep before  creating filesystem

joda17:
In some computers mkfs try to creating filesystem, before kernel update partition table.


stefansaraev:
thanks.

PR #3773, opened 2015-01-11
drop pulse

stefansaraev:
pulse support in openelec is nonexistent. even for selfbuilders. I re-added it for testing. I remove it now

will merge in 24 hours. if one wants it back - he should send PRs with fixes.


HamiltonCompany:
OK we discussed this again last year.  Our company has used Openelec with pulse audio 5 for nearly 1 year in a production environment and it is nearly flawless with the new Kodi AE.  I know you guys really don't want to support bluetooth but your built in bluetooth support is also flawless when pulse is compiled in.  This is in fact the only way to support bluetooth audio in OE.  Plus pulse audio 6 now includes support for 7.1 channel streams (which was a major drawback before).  

So to sum up why you should throw pulse back in:
*Great support in Kodi
*Only OE support for Bluetooth Audio (much better than the old ALSA hack)
*Our company will donate a sizable donation =)
*With some direction I will update the package to Pulse 7.0 and test it.

lrusak:
@HamiltonCompany Pulseaudio still lacks HD audio support (TrueHD, DTS-HD) so it is still a blocker in my opinion. If you want to ping me in the IRC we can talk about setting up a place in the forum where you can provide builds with pulseaudio.

chewitt:
@lrusak It's a topic that I plan to raise at the next formal team discussion.

HamiltonCompany:
Haven't done IRC for 10 years or more.  Lemme go did up a client. =)

vpeter4:
I would like to have support for BT speaker and from what I see PA is a must. At least with BlueZ 5. Seems it can't be done with ALSA. Am I correct?
But is it possible to have ALSA as default package and PA as addon or some sort of side daemon? So we can still have stable OpenELEC with ALSA and PA for some special uses?

fritsch:
You can use the dbus API (there is a special module to ignore devices for PA) to tell pulse to not attach itself to hdmi / analog / whatever devices and only use it for the bluetooth output.

With a little hack to Kodi (one liner) it can use pulse / alsa device next to each other. That way you could manage BT with PA und keep everything else as is.

lrusak:
@fritsch care to elaborate?

fritsch:
What detail is missing?
Am 16.10.2015 4:12 vorm. schrieb "Lukas Rusak" <notifications@github.com>:

PR #3774, opened 2015-01-11
new package: giflib

stefansaraev:
giflib:host required for kodi's I*** new texturepacker
giflib:target may be required later, for kodi

### DIFF
--- a/packages/graphics/giflib/package.mk
+++ b/packages/graphics/giflib/package.mk
@@ -1,0 +1,41 @@
+################################################################################
+#      This file is part of OpenELEC - http://www.openelec.tv
+#      Copyright (C) 2009-2014 Stephan Raue (stephan@openelec.tv)
+#
+#  OpenELEC is free software: you can redistribute it and/or modify
+#  it under the terms of the GNU General Public License as published by
+#  the Free Software Foundation, either version 2 of the License, or
+#  (at your option) any later version.
+#
+#  OpenELEC is distributed in the hope that it will be useful,
+#  but WITHOUT ANY WARRANTY; without even the implied warranty of
+#  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+#  GNU General Public License for more details.
+#
+#  You should have received a copy of the GNU General Public License
+#  along with OpenELEC.  If not, see <http://www.gnu.org/licenses/>.
+################################################################################
+
+PKG_NAME="giflib"
+PKG_VERSION="5.0.5"
+PKG_REV="1"
+PKG_ARCH="any"
+PKG_LICENSE="OSS"
+PKG_SITE="http://giflib.sourceforge.net/"
+PKG_URL="$SOURCEFORGE_SRC/giflib/$PKG_NAME-$PKG_VERSION.tar.bz2"
+PKG_DEPENDS_HOST="zlib:host"
+PKG_DEPENDS_TARGET="toolchain zlib"
+PKG_PRIORITY="optional"
+PKG_SECTION="graphics"
+PKG_SHORTDESC="giflib: giflib service library"
+PKG_LONGDESC="giflib: giflib service library"
+
+PKG_IS_ADDON="no"
+PKG_AUTORECONF="yes"
+
+PKG_CONFIGURE_OPTS_HOST="--disable-shared --enable-static"
+PKG_CONFIGURE_OPTS_TARGET="--with-sysroot=$SYSROOT_PREFIX"
+
+post_makeinstall_target() {
+  rm -rf $INSTALL/usr/bin
+}


MilhouseVH:
@seo, there's a problem with the TexturePacker changes, somewhere, most likely this package, which doesn't respect the frames per second of GIFs, and instead results in them being animated as quickly as possible (causing very high CPU load, not to mention turning the GIFs into a visual mess).

For instance, [this](https://github.com/popcornmix/xbmc/commit/9424c71973be32369fbba70d4ec752f49d0d921f) GIF is animated rapidly in OpenELEC, rather than at a sedate 2 frames per second.

I've tested the above GIF in current Kodi master on Ubuntu 14.10, with TexturePacker changes (PR5436 and PR6170), and the GIF is displayed correctly at 2 fps, so the problem seems to be somewhere in the OpenELEC changes required to build current Kodi master - these are:

#new package: libsquish
https://github.com/stefansaraev/OpenELEC.tv/commit/d6f4bebf108093c59c91d74688c35688deced826
#Use our own texturepacker
http://nmacleod.com/public/oebuild/patches/00_texturepacker.patch

Can you confirm if you see similar GIF behaviour on x86 with OpenELEC master + Kodi master and the above patches? I'm seeing it on R-Pi.

stefansaraev:
not sure on x86_64, back then I was using a very customised build. I dont remember having the behaviour you reported. ATM I think it's fine on wetek + kodi master

MilhouseVH:
I guess it's time I rattled off a Generic/x86 build... been a few years since I tried one... :)

MilhouseVH:
Same behaviour as R-Pi on OpenELEC Genric/x86_64 (Revo 3700) - GIF ignores frame rate, plays out too fast.

Same GIF, same Kodi, on Ubuntu 14.10, plays the GIF correctly.

Steps to reproduce:

a) 1. Patch skin.confluence with aforementioned busy.gif
2. Replace kodi-logo.png with busy.gif in 720p/Home.xml - this should place the GIF on the home screen, making it easier to see.
3. Make a custom build with the modified skin.

Alternatively:
b) 1. Clone skin.confluence, changing addon id and name
2. Copy busy.gif to the new skin.confluence/media folder
3. Replace kodi-logo.png with busy.gif in 720p/Home.xml - this should place the GIF on the home screen, making it easier to see
4. Change to the modified skin

The correct result in both cases (as it is with Ubuntu) should be a spinner animating at 2 frames per second - the indicator should take 6 seconds to rotate once around the dial.

The incorrect result (as it is with OpenELEC on x86 and R-Pi) is a spinner that rotates faster than 2 frames per second. Typically I'm seeing the spinner complete one rotation approximately every second, which would would suggest at least 12 frames per second.

Prior to this TexturePacker change, the GIF displayed from Textures.xbt would display correctly (at least, on a RPi), however it seems that "unpacked" GIFs when read from the media folder *would* display too fast, so it seems that GIFs were always partially broken in OpenELEC. Now, they seem to be completely broken, even when read from Textures.xbt. Or there's some other voodoo going on here that I haven't been able to work out, but certainly GIFs are pretty broken, and now they're worse. :)

I will contact @Memphiz to see if rebasing (5180)[https://github.com/xbmc/xbmc/pull/5180] will be of any use to OpenELEC.

Memphiz:
are you using giflib5 for sure?

Memphiz:
i mean on the texturepacker side...

stefansaraev:
ok. just tested on wetek + Isengard:
- busy.gif accessed via kodi filemanager - fine but not animated, no clue it is supposed to be shown as static image
- busy.gif in /storage/.kodi/addons/skin.confluence/media + a change to DialogBusy.xml - fine
- busy.gif in Textures.xbt + a change to DialogBusy.xml - fine.

```
This is a GIF - lets load it via libgif...
busy.gif
    frame    0                                ARGB * (32,32 @ 4096 bytes)
    frame    1                                ARGB * (32,32 @ 4096 bytes)
    frame    2                                ARGB * (32,32 @ 4096 bytes)
    frame    3                                ARGB * (32,32 @ 4096 bytes)
    frame    4                                ARGB * (32,32 @ 4096 bytes)
    frame    5                                ARGB * (32,32 @ 4096 bytes)
    frame    6                                ARGB * (32,32 @ 4096 bytes)
    frame    7                                ARGB * (32,32 @ 4096 bytes)
    frame    8                                ARGB * (32,32 @ 4096 bytes)
    frame    9                                ARGB * (32,32 @ 4096 bytes)
    frame   10                                ARGB * (32,32 @ 4096 bytes)
    frame   11                                ARGB * (32,32 @ 4096 bytes)
```

so, whatever breaks it in your builds, it's nothing related to this PR (which is just a preparation for Isengard), as it look gif rendering is "broken in general"  for you

and let's better move the discussion to irc.

MilhouseVH:
Seems to be fixed by the addition of https://github.com/xbmc/xbmc/pull/6285, even when displaying non-packed GIFs, which I can't really explain...

PR #3782, opened 2015-01-12
[addons/boblight] - add boblight-aml - boblight client for amlogic devic...

Memphiz:
This adds a new client for amlogic based SOCs which have kernel support for the amvideocap driver.

Basically this results in a new binary boblight-aml which can be used as boblight client which is active as soon as a video is played with aml hw acceleration.

e.x. call it like:

boblight-aml -s <boblightdserverip> -p 100

This starts the client with priority 100 - so you can use the normal boblight addon for having the static boblight in the menu (it has prio 128) and as soon as a video starts the boblight-aml takes over and drives the leds - after movie playback the static light is in charge again.

### DIFF
--- a/packages/addons/service/multimedia/boblightd/patches/boblightd-2.0.5-add_aml_client.patch
+++ b/packages/addons/service/multimedia/boblightd/patches/boblightd-2.0.5-add_aml_client.patch
@@ -1,0 +1,632 @@
+diff --git a/src/Makefile.am b/src/Makefile.am
+index 9ba5381..c5a4dc6 100644
+--- a/src/Makefile.am
++++ b/src/Makefile.am
+@@ -7,6 +7,7 @@ AM_CFLAGS =\
+ 	 -g
+ 
+ bin_PROGRAMS = boblightd \
++	boblight-aml \
+ 	boblight-constant
+         
+ 
+@@ -19,6 +20,11 @@ endif
+ 
+ endif
+ 
++boblight_aml_SOURCES  = clients/boblight-aml/boblight-aml.cpp
++boblight_aml_SOURCES += clients/boblight-aml/flagmanager-aml.cpp
++boblight_aml_SOURCES += clients/flagmanager.cpp
++boblight_aml_SOURCES += util/misc.cpp
++
+ boblight_v4l_SOURCES = \
+ 	clients/boblight-v4l/boblight-v4l.cpp \
+ 	clients/boblight-v4l/flagmanager-v4l.cpp \
+diff -urPp src/clients/boblight-aml.cpp src/clients/boblight-aml/boblight-aml.cpp
+--- /dev/null	Thu Jan  1 00:00:00 1970
++++ b/src/clients/boblight-aml/boblight-aml.cpp	Thu Jan 15 10:24:16 2015
+@@ -0,0 +1,491 @@
++/*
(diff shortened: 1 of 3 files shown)


Memphiz:
e.x. for wetek.play this is needed in addition for getting the needed /dev/amvideocap0 device:

https://github.com/codesnake/OpenELEC.tv/pull/36

Memphiz:
nearly forgot - kudos to @davilla who did the initial work on this and pointed out the amvideocap api which was added to current amlogic kernels :)

davilla:
:)

stefansaraev:
thanks!

we may need some check in bin/boblightd.start  if /dev/amvideocap0 exists and then start boblight-aml, otherwise start regular boblightd/boblight-x11 right ?

packages/addons/service/multimedia/boblightd/patches/boblightd-2.0.5-add_aml_client.patch
```diff
@@ -0,0 +1,462 @@
...
++boblight_aml_SOURCES  = clients/boblight-aml/boblight-aml.cpp
++boblight_aml_SOURCES += clients/boblight-aml/flagmanager-aml.cpp
++boblight_aml_SOURCES += clients/flagmanager.cpp
++boblight_aml_SOURCES += util/misc.cpp
++
+ boblight_v4l_SOURCES = \
+ 	clients/boblight-v4l/boblight-v4l.cpp \
+ 	clients/boblight-v4l/flagmanager-v4l.cpp \
+diff -urPp src/clients/boblight-aml.cpp src/clients/boblight-aml/boblight-aml.cpp
+--- a/src/clients/boblight-aml.cpp	1970-01-01 01:00:00.000000000 +0100
```
stefansaraev: build fails because file created in src/clients not src/clients/boblight-aml. bad c/p ?

should be --- /dev/null

same for flagmanager-aml.cpp and flagmanager-aml.h

or pick http://sprunge.us/SfDj

Memphiz:
args - sorry - updated ... 

well checking for /dev/amvideocap0 would be ok i guess - but i am not sure if the device is named the same on all amlogic platforms ... thats why i added an -d option to boblight-aml for beeing able to specify the device (where amvideocap0 is default) - but in general a check for this device would be good yeah :)

Memphiz:
added the change to boblight.start - untested though ... because there is no conf file for boblight-aml this would only work when we want to use the local boblightd ... maybe i change it so it supports a config file in addition where the user could point it to an remote boblightd (thats the setup i am using atm).

davilla:
amvideocap0 should be fine

Memphiz:
don't merge yet - still working on using the xml config from the boblight addon if found ...

Memphiz:
Ok i am done with it now. It has now 2 secs timeout if it couldn't be connected and in case it found a settings.xml from the boblight (client) addon it uses its settings (hostip/port for the daemon to connect to, and all the boblight settings from the "movies" category).

stefansaraev:
thanks. if you dont have something else to add I will bump boblightd addon and rebuild it.
should we merge #3798 now or wait for for second opinion  -> @codesnake ?

Memphiz:
nothing i can think of for the boblight-aml client :) (using it happily since 2 days ...)

PR #3783, opened 2015-01-13
mkimage enhancements

stefansaraev:
- projects are allowed to set name for target kernel (RPi/WeTek_Play use kernel.img)
- minors

@sraue @MilhouseVH could you confirm this still produces valid images for RPi/x86, just in case ?

### DIFF
--- a/projects/RPi/options
+++ b/projects/RPi/options
@@ -147,3 +147,6 @@
 
   # build and install 'FishBMC' Visualization (yes / no)
     KODI_VIS_FISHBMC="no"
+
+  # kernel image name
+    KERNEL_NAME="kernel.img"
--- a/projects/WeTek_Play/options
+++ b/projects/WeTek_Play/options
@@ -160,3 +160,6 @@
   # extra build dependeices
     EXTRA_DEPS="mkimage"
     EXTRA_DEPS_PKG="u-boot-tools"
+
+  # kernel image name
+    KERNEL_NAME="kernel.img"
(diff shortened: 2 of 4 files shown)


MilhouseVH:
I'll certainly take a look, but probably not for a couple of days as I'm a bit swamped right now. Initial scan of the changes doesn't frighten me, which is a good sign...

MilhouseVH:
What about the instances of "kernel.img" at [line 360](https://github.com/stefansaraev/OpenELEC.tv/blob/mkimage/scripts/image#L360) and 365? Don't they need ${KERNEL_NAME} too?

stefansaraev:
I didnt touch it because it is noobs specific.

MilhouseVH:
I've never made an image, to be honest.

Wouldn't the BOOT_IMAGE kernel parameter also need to change to match $KERNEL_NAME?

stefansaraev:
no. it has nothing to do with scripts/mkimage

MilhouseVH:
Kinda confused then as to what is achieved by this change.

stefansaraev:
well not much. its to avoid "special handling" and adding more and more code to mkimage for such "special projects" like wetek play, maybe others in future. imagine a device that expects kernel stored as boot.img - only one line is to be added to project's options. 
atm I cant build and test all supported projects, thats why I ask here :)

MilhouseVH:
OK.

Then my only question is why not also use $KERNEL_NAME to somehow update linux.*.conf in terms of the value used for BOOT_IMG, although that's perhaps more intrusive a change.

With this change, isn't it possible that will be able to create an image that is using a custom kernel name, but when you update the system from a tar file (even if it's the same source, and same custom kernel namel) the actual kernel name written to the boot device will revert to kernel.img as this is defined by BOOT_IMG.

stefansaraev:
on  WeTek its done in devicetree. I am not goind to mess with that.

PR #3784, opened 2015-01-13
IMX: Revisit coherent_pool

fritsch:
- This should go to 5.0
- The bump is again needed for usb dual tv tuner card - Evolve Venus

Btw. it's not good to nail such values into the kernel config, cause they are not overwritten properly by user supplied argument.


fritsch:
ping @vpeter4 @stefansaraev @firmwarethatisjusttoolargeandsucks

stefansaraev:
you sure those are not overwriten?
if so +1.

vpeter4:
Those parameters are added: https://github.com/OpenELEC/OpenELEC.tv/blob/47138b261c0dc299b7ec695e874733adf9df4354/projects/imx6/patches/linux/linux-007-config_cmdline.patch

fritsch:
Yeah, with the above code the hard coded kernel cmd_line overwrite the user supplied ones, cause they are appended and come last therefore.

stefansaraev:
bootloader arguments should be appended to default cmdline. not default cmdline to bootloader args. CONFIG_CMDLINE_EXTEND (should?) take care of that.

PR #3785, opened 2015-01-13
projects/imx6/patches/linux: prepend kernel config cmdline to bootloader...

vpeter4:
... arguments from uEnv.txt


stefansaraev:
thanks. I'll take care tomorrow, I have to move coherent_pool back where it was but its too late here :)

PR #3789, opened 2015-01-14
update to ffmpeg-2.5.4 like xbmc and update patches

mrdominuzq:
update to ffmpeg 2.5.3 and update patches to match xbmc-ffmpeg branch :)

actually they are only on 2.5.2, i can change it to that if it's necessary 

added the disable-sdl like https://github.com/xbmc/xbmc/commit/e0d346d7719844659e95e798ccccd9dee358960f
hope it's correct ?

havent tested the build, but it build fine :)


stefansaraev:
this should not be merged before kodi is bumped to 15 in master.
there is a good reason we use exactly the same ffmpeg version kodi uses.

mrdominuzq:
okay :)

fritsch:
Wait, please keep open. We will update xbmc to 2.5.3 next chance (when time permits).

PR #3791, opened 2015-01-14
Allow unique build directories by appending project-defined suffix

MilhouseVH:
With this change, adding the property BUILD_SUFFIX to a project options file (or on the command line) would allow different build directories to be used without having to modify OPENELEC_VERSION, which has implications elsewhere (ie. where source tarballs are downloaded from etc.).

So for instance if I want to rattle off a quick RPi Helix build, without trashing my existing master build (`build.OpenELEC-RPi.arm-devel`) I could specify `BUILD_SUFFIX=helix` and an alternate build directory named `build.OpenELEC-RPi.arm-devel-helix` would be created instead.


vpeter4:
I don't get the purpose of this. If you get new build folder then you obviously change some package also? So you already messed system. Or did I miss something fundamental?

MilhouseVH:
You might be applying local patches in the project folder, and trying different builds. Having one single `build.OpenELEC.RPi.arm-devel` build directory is... restrictive. There's no need to change any packages for a completely different build directory to be useful, it can all be done with patches. You may just be changing a project option.

stefansaraev:
ok from me if it makes your life easier.

sraue:
sometimes it makes sense i think, dont know a example actually, but i think i missed this the one or other day too :-)

PR #3793, opened 2015-01-14
Enable binary patching

MilhouseVH:
This pull request allows binary patches to be applied during the OpenELEC build process.

As one example, pull request [PR6201](https://github.com/xbmc/xbmc/pull/6201) contains a binary patch (image file, test.png, in commit [8438b740](https://github.com/Montellese/xbmc/commit/8438b740f24cd3085b711c8e85f4204001f67ff9)) that would otherwise not be possible to apply on top of OpenELEC without repackaging kodi..


stefansaraev:
ok from me. @sraue ?

PR #3794, opened 2015-01-15
wetek_play: Add full support for WeTek internal DVB-T/T2, DVB-C, DVB-S/S2 and ATSC tuners

codesnake:
Update kernel patches and configuration for full support of WeTek internal DVB-T/T2, DVB-C, DVB-S/S2 and ATSC tuners.


stefansaraev:
thanks :)

PR #3798, opened 2015-01-15
Wetekfixes

Memphiz:
Those contain the following fixes for the wetek.play - this is all kernel related:
1.  pixelclock for 23.976Hz mode was wrong
2. yamaha amp (and maybe others) couldn't lock audio in 1080p24 mode (tried it with 44k1, 48k and 96k - none worked before - all work with my patch)
3. allow 1.5ghz for the meson6 cpu (there was still a clamp to 12000 internally)
4. fix race condition in ge2d (which made boblight-aml stuck after some time of usage)
5. add the amvideocap device to the devicetree (needed for boblight-aml).


codesnake:
Well done @Memphiz!
@stefansaraev I'm unable to test the PR now, but I think @Memphiz already did, and you can merge it.

PR #3803, opened 2015-01-17
[RFC] Wetek config changes

sraue:
sync WeTeks kernel config with our other projects, use eventlircd instead AMRemote
@codesnake @stefansaraev @memphiz


sraue:
all our projects using the same options basically, why not here too?

stefansaraev:
amremote allows using ANY remote with wetek's ir. so you are going to disable that ?

stefansaraev:
would be nice if you squash all kernel config changes to one commit after you address comments, so I'll have another look. thanks.

codesnake:
After testing I see several problems:
* After turning on the box first two splash screens (WeTek and OE) are not shown, just no signal. Most likely caused by some changes in kernel config
* Power button on remote doesn't work. Most likely because of disabled amremote
* Some buttons on the remote mapped differently after the PR

packages/sysutils/eventlircd/udev.d/98-eventlircd.rules
```diff
@@ -64,6 +64,14 @@ SUBSYSTEMS=="acpi", ATTRS{hid}=="LNXPWRBN", \
   ENV{eventlircd_enable}="true", \
   ENV{eventlircd_evmap}="power.evmap"
 
+SUBSYSTEMS=="input", ATTRS{name}=="key_input", \
```
stefansaraev: gotcha! this one fixes the hardware power button not working. you can push it directly to master ;)

sraue:
updated @memphiz @stefansaraev @codesnake without the eventlircd stuff (will do a PR later maybe)

@codesnake if i enable "CONFIG_BTRFS_FS=y" the first (wetek) splash will not be shown, can you take a look at this?

codesnake:
OK, no problem, I'll take a look why there's a problem with splashes.

projects/WeTek_Play/linux/linux.arm.conf
```diff
@@ -2981,7 +3071,7 @@ CONFIG_SWITCH=y
 # CONFIG_EDAC is not set
 CONFIG_RTC_LIB=y
 CONFIG_RTC_CLASS=y
-CONFIG_RTC_HCTOSYS=y
+# CONFIG_RTC_HCTOSYS is not set
```
stefansaraev: any reason for disabling this? :)

projects/WeTek_Play/linux/linux.arm.conf
```diff
@@ -1043,20 +1062,20 @@ CONFIG_AM_WIFI_SD_MMC=y
...
-CONFIG_RTL8188EU=m
-CONFIG_RTL8188EU_MP=m
-CONFIG_SDIO_DHD_CDC_WIFI_40181_MODULE=y
+# CONFIG_RTL8821AU is not set
+# CONFIG_RTL8812AU is not set
+# CONFIG_RTL8192CU is not set
+# CONFIG_RTL8192DU is not set
+# CONFIG_RTL8188EU is not set
+# CONFIG_RTL8188EU_MP is not set
+CONFIG_SDIO_DHD_CDC_WIFI_40181_MODULE=m
```
stefansaraev: this is on-board hardware, I would keep it built-in. @codesnake what's your opinion?

projects/WeTek_Play/options
```diff
@@ -111,13 +111,13 @@
...
   # Modules to install in initramfs for early boot
-    INITRAMFS_MODULES="softcursor bitblit font fbcon"
+    INITRAMFS_MODULES=""
 
   # additional drivers to install:
   # for a list of additinoal drivers see packages/linux-drivers
   # Space separated list is supported,
   # e.g. ADDITIONAL_DRIVERS="DRIVER1 DRIVER2"
-    ADDITIONAL_DRIVERS=""
+    ADDITIONAL_DRIVERS="RTL8192CU RTL8192DU RTL8188EU"
```
stefansaraev: these 3rdparty drivers are horrible. see what happens on the rpi front.. sure, can be added if you want them so much, but later. lets fix the remaining issues first before adding stuff we dont/cant support at all. my 2 cents.

stefansaraev:
for everything else than my comments above, ack from me.
@codesnake for an opinion on rcu/iosched changes

projects/WeTek_Play/options
sraue: this i added because i removed the intree drivers (see your comment above). This in-tree drivers are basically the same drivers (+ some more). we have in this way a better control with patches which ads new product ids and patches etc...
All of the RTLxxxxXU drivers (with "U") are drivers for USB hardware.

projects/WeTek_Play/linux/linux.arm.conf
sraue: no, my fault... will change this back later this evening

projects/WeTek_Play/options
stefansaraev: hehe. my commend above is about CONFIG_SDIO_DHD_CDC_WIFI_40181_MODULE 

> This in-tree drivers are basically the same drivers (+ some more). we have in this way a better control with patches which ads new product ids and patches etc

we dont have control at all but ok ;)

codesnake:
@sraue Unfortunately I haven't found why splash cannot be shown after enabling CONFIG_BTRFS_FS. Quite strange. All other changes at first sight don't cause any problems.

projects/WeTek_Play/linux/linux.arm.conf
stefansaraev: ah. when CONFIG_SDIO_DHD_CDC_WIFI_40181_MODULE is built-in it has no change to load the nvram config and firmware. unless it's not in initrd... maybe...

PR #3804, opened 2015-01-17
ffmpeg: update to ffmpeg-2.4.6

stefansaraev:
should be merged and backported to 5.0 after xbmc/xbmc#6217 goes in


sraue:
can we have the single patches back and the wtv index patch please?

stefansaraev:
> can we have the single patches back

why? it's easier as one patch. and it comes from kodi. officialy.

> and the wtv index patch please

ah my fault. re-added.

PR #3805, opened 2015-01-17
Add PKG_SOURCE_NAME option

ghollingworth:
This patch adds a PKG_SOURCE_NAME option to the package.mk file

This allows you to download a decorated url from the PKG_URL but not use the decoration, an example of this is:

PKG_URL="https://api.github.com/repos/FiveNinjas/Skin/tarball/$PKG_NAME-$PKG_VERSION?access_token=biglongapikeyyougetfromgithub"
PKG_SOURCE_NAME="$PKG_NAME-$PKG_VERSION.tar.gz"
 
unpack() {
  mkdir -p $BUILD/$PKG_NAME-$PKG_VERSION
  tar xfz sources/$PKG_NAME/$PKG_NAME-$PKG_VERSION.tar.gz --directory $BUILD/$PKG_NAME-$PKG_VERSION
  mv $BUILD/$PKG_NAME-$PKG_VERSION/FiveNinjas*/* $BUILD/$PKG_NAME-$PKG_VERSION/
}

I had to insert the unpack because github doesn't name the enclosed directory in the right format

### DIFF
--- a/scripts/get
+++ b/scripts/get
@@ -33,8 +33,8 @@ if [ -n "$PKG_URL" ]; then
     PACKAGE="$SOURCES/$1/$SOURCE_NAME"
     PACKAGE_MIRROR="$DISTRO_MIRROR/$PKG_NAME/$SOURCE_NAME"
     [ "$VERBOSE" != "yes" ] && WGET_OPT=-q
-    [ "$PKG_SOURCE_NAME" != "" ] && OUTPUT_OPTION="-O $SOURCES/$1/$PKG_SOURCE_NAME"
-    WGET_CMD="wget --passive-ftp --no-check-certificate -c $WGET_OPT -P $SOURCES/$1 $OUTPUT_OPTION"
+    [ -n "$PKG_SOURCE_NAME" ]  && WGET_OUTPUT_OPTION="-O $SOURCES/$1/$PKG_SOURCE_NAME"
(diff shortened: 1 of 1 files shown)


vpeter4:
Usually in such cases file is manually downloaded, repacked and put on openelec sources. Using script tools/mkpkg/mkpkg_some_package.

And such approaches has been proposed in the past and not accepted. Like #3698.

ghollingworth:
Yeah I expected something like that.  Problem with rebuilding etc is that the package is not under the same license and it makes it difficult to create an automated build system (i.e. to automate the build when the github sources change)

Looks like I'm not the only person looking for a solution though, how about a post_wget method to handle such a case by the package?

Gordon

vpeter4:
It doesn't matter how it is handled - no change like this was accepted in the past. I also proposed downloading directly from a git for example.
But maybe you would have more luck.

stefansaraev:
-1

ghollingworth:
Thanks for the comments I've updated to reflect...

Happy to receive useful comments or any other ways of achieving this if this is not the right mechanism.

chewitt:
@ghollingworth - after consulting with @sraue - If you are still interested in adding this please clean-up or rebase the commits. If not, please close out the ticket. Thanks.

ghollingworth:
I no longer have a requirement for it, I'm happy to close

PR #3809, opened 2015-01-18
[WeTek_Play] - add led driver for the blue powerled

Memphiz:
As the title says. This adds support for switching on and off the blue led on the WeTek.Play via sysfs led class (with the usual triggers which are available like timer, heartbeat, sdio, gpio and so on).

Led can be accessed in /sys/class/leds/wetek:blue:powerled

Only the blue led can be controlled. The red one seems to be hardwired (so you can only switch to Blue or to Red with this).

@codesnake for review ...


Memphiz:
I have added a patch which enables the cpu temp sensor. Its a bit WIP because atm it doesn't give correct readings because of:

https://github.com/codesnake/linux-amlogic/blob/master/drivers/amlogic/input/saradc/saradc.c#L356

Someone with the specs manual has to fix the adc to celsius conversion ...

codesnake:
Looks good. Have not tried, though.

Memphiz:
WIP - i will add a led-remote-trigger which allows us to blink the led whenever a key on the remote control is pressed ... (long live open source - i found that code laying around on another github repo already ;) ).

Memphiz:
Ok ledtrigger is done. With this PR and this little script (started from .config/autostart.sh for example) you get this behavior:

script: http://pastebin.com/GLaCEgk6
behavior: http://youtu.be/T8pgqdoFeK8

PR #3811, opened 2015-01-18
systemd service ordering fixes

stefansaraev:
those are systemd service ordering fixes backported from master

needs testing.

should probably fix sshd issue reported on forums (and other random services that depend on tmpfiles) not started due to tmpfiles-setup failing


stefansaraev:
please dont merge yet. I cant verify if 5.0 is affected. if its affected - sshd wont start once a few reboots. or more stuff may fail randomly after first install, but I cant reproduce, must be confirmed first..

PR #3814, opened 2015-01-19
fs-resize: move as systemd target

### LINKED ISSUE #3812: Auto resize is never called on master
I compiled the auto expandable image for RPi, branch master a few hour ago (commit https://github.com/OpenELEC/OpenELEC.tv/commit/f4c40cfaa6b1c77ac1f69ae2dc9328ef5690400b )
Auto-resize never happens. The storage partition stays at 34MB and the .please_rezise_me get never removed. Kodi is launched without reboot.

I think that it also affects other projects.

There is a systemd SKIP message about ordering cycle at boot, but I'm not sure if it has something to do with this bug.

stefansaraev:
closes #3812


stefansaraev:
tested backup/restore. tested fs-resize. tested hard/soft reset. all working.
@Kivutar also confirms fs-resize works.

PR #3815, opened 2015-01-19
vdr-satip: update to vdr-satip-1.0.2

CvH:
less important CA related updates


CvH:
sry fu up pull request

PR #3820, opened 2015-01-20
IMX6: Don't override user set colors

fritsch:
Default is 24 bit. 16 bit was set for performance reasons, but image quality is quite bad.


fritsch:
Please also pick for 5.0 branch

Issue #3763, opened 2015-01-09
OpenELEC 5.0 - MP4 Playback Issues (slowdown, choppy playback)


sonicnkt:
Hi,
im testing the new OE5.0 now for several days on my AMD Kabini System and even tho the system crashes and random reboots of 4.2.X are gone i came across several Mp4 files that have playback issues (some youtube videos as well using youtube addon).

The playback is verry choppy and audio is out of sync.
It also slowsdown the whole kodi system if you play these files in the background and navigate through the menu. (FPS dropped from 60 to 5-8) and CPU Usage shows 100% on one core 

These files play fine on my PC using VLC or Kodi so it is a OE5.0 related Problem or atleast OE on AMD Kabini.
I tested playback using VDPAU acceleration aswell as without, no change.

XBMC-Log of the playback:
http://pastebin.com/hV0C6PMA

Here is a short test file uploaded by lordalibaski on the forums:
https://www.sendspace.com/file/ws9o23
Codec Info:
http://pastebin.com/tRsUPD6E

fritsch:
Better:
https://dl.dropboxusercontent.com/u/55728161/OpenELEC-Generic.x86_64-5.0.0-fritsch2.tar
?

do a full backup of your data before. This build does not have all dvb-c /
dvb-t drivers and is for testing purpose only.

2015-01-09 13:20 GMT+01:00 sonicnkt <notifications@github.com>:

> Hi,
> im testing the new OE5.0 now for several days on my AMD Kabini System and
> even tho the system crashes and random reboots of 4.2.X are gone i came
> across several Mp4 files that have playback issues (some youtube videos as
> well using youtube addon).
>
> The playback is verry choppy and audio is out of sync.
> It also slowsdown the whole kodi system if you play these files in the
> background and navigate through the menu. (FPS dropped from 60 to 5-8) and
> CPU Usage shows 100% on one core
>
> These files play fine on my PC using VLC or Kodi so it is a OE5.0 related
> Problem or atleast OE on AMD Kabini.
> I tested playback using VDPAU acceleration aswell as without, no change.
>
> XBMC-Log of the playback:
> http://pastebin.com/hV0C6PMA
>
> Here is a short test file uploaded by lordalibaski on the forums:
> https://www.sendspace.com/file/ws9o23
> Codec Info:
> http://pastebin.com/tRsUPD6E
>
> --
> Reply to this email directly or view it on GitHub
[...]

PR #3822, opened 2015-01-22
Radeon: disable RADEON_GEM_GTT_WC for Kabini and Aruba

fritsch:
This fixes some performance regressions on Kabini and Aruba. Sadly it does not solve anything when also applied to E-Series.

Should be picked until upstream knows what's going on: https://github.com/OpenELEC/OpenELEC.tv/issues/3763#issuecomment-70944593


fritsch:
Push to 5.0 - then at least Kabini will work again for sw decoding.

PR #3831, opened 2015-01-23
IMX: Disable console blank

fritsch:
After discussion with @smallint and @wolfgar - this should be added as default.


fritsch:
Please pick on 5.0 also

PR #3838, opened 2015-01-25
wetek_play: set scaling_max_freq=1512000

stefansaraev:
lost after switch to wetek_play.dtd
//cc @codesnake


codesnake:
Agree :)

PR #3840, opened 2015-01-26
WT: Backport fullscreen dirty region fixes PR6090

fritsch:
This makes 1080i50 working without a/v sync issues on WetekPlay. Skips are highly reduced also when opening the codec screen.


stefansaraev:
gotcha. confirmed.

PR #3850, opened 2015-01-29
Fix divide by zero crash

pinhao:
When ds3000_set_frontend is called with 0 symbol rate the module would crash with a divide by
zero error.
Added simple symbol rate validation.
Tvheadend was triggering this bug and crashing openelec.


stefansaraev:
thanks much. are you the author of this patch?

pinhao:
Yes

stefansaraev:
did you send it upstream ?

pinhao:
No because I don't exactly were I should submit it.
I did send an email about the problem to the author of the drive, but I didn't send the patch because I didn't have one at the time.

stefansaraev:
this should be it https://patchwork.linuxtv.org/project/linux-media/list/

I'll do build test and maybe merge later today. 

//cc @fritsch for 2nd opinion.

pinhao:
which one? the inner print?
I've added both because the first one was already there I just added more detail and the second one was to to give info about the range to someone enabling the debug flag.
But yes, they are a bit redundant.

sraue:
is kernel 3.17 affected too?

pinhao:
Yes, I think the bug exists since the driver was added.
It's weird that it wasn't detected or fixed before. For my case it crashed when tvheadend tuned to frequency 0 and symbol rate 0.
Maybe something changed in the call stack and before the request was invalidated before crashing.

fritsch:
WTH would any sane software tune to this value?

PR #3854, opened 2015-01-30
drop afp support

stefansaraev:
afpfs-ng is unmaintained for like 6 years now, atm broken and not used in openelec. time to go. comments?


bjornwiberg:
I Really think you should consider to use AFP, at least on RPi because the speed compared with Samba is dramatically better on AFP. I now use OpenElec 4.1.2, just to maintain my speed between my RPi and my TimeCapsule.

PR #3857, opened 2015-02-01
wetek_play: avl6211: let userspace knows we support dvbs2

stefansaraev:
... at least vdr/wirbelscan is confused and thinks there is no dvb-s2 support


stefansaraev:
//cc @codesnake

stefansaraev:
got ack from Sasha_sr

PR #3863, opened 2015-02-03
AMD Fix APU performance regression

fritsch:
That bug was hitting us very long. After talking again with Alex Deucher (agd5f) it was shown that https://bugs.freedesktop.org/show_bug.cgi?id=88658 is this cause.

It was introduced between 10.3.1 and 10.3.2 Mesa via: http://cgit.freedesktop.org/mesa/mesa/commit/?h=10.3&id=64c2bdc334ba472603b1e7cd2c3046cfbce285b6

Thanks much to @macman88 for bisecting


fritsch:
Confirmed by user!

PR #3864, opened 2015-02-03
Amd bisect master

fritsch:
See: https://github.com/OpenELEC/OpenELEC.tv/pull/3863


fritsch:
This needs testing, btw. Don't push yet. The 5.0 stable PR linked above is save - I tested it.

PR #3867, opened 2015-02-04
Mesa: Fix the performance regression in an upstream conform way

fritsch:
Upstream suggested workaround for the AMD performance regression.

Please keep open until I have positive feedback from the testing thread. The "Revert" worked, but would produce other issues Michel Daencer told.


fritsch:
User reported, that the fix is fine.

fritsch:
Hint @fritsch also push to ontop of what we have in master.

PR #3878, opened 2015-02-06
remove create_installsrtick* / create_sdcard

stefansaraev:
time to remove it. people should use images.

ping @sraue


CvH:
Maybe add the necessary tools (Win32DiskImager.exe ...) to write the images to usb stick ? 
I think it is quit impossible for a lot people to do this.

stefansaraev:
sure I can do, but I dont know if their license permits redistributing. must check.

CvH:
At least add a readme with the necessary instructions (where to dl the tools etc) to the package, GPLv2 by the way.

A setup.exe would be the better way, but I think nobody want to waste his time on that :)

vpeter4:
Ask wetek nicely to get source code for their installer.

stefansaraev:
guys guys, stop it. setup.exe or whatever does not make sense in .tar releases, those will be ONLY for updating.

CvH:
We got OE*.tar and OE*-efi.img.gz, both contain (as soon as merged) nothing to install/create USB Stick for OE, neither exist a easy setup tool to download. 
For an average user is is absolutely not clear what to do. You need to find and read a wiki to create a usb stick - that is even less user-friendly as it was before. To write an image to an usb drive is not a common task for 99% of the people.

stefansaraev:
sure. thanks for the suggestion. please send a PR to stefansaraev:installsticks

CvH:
I just wanted to point out that YOU remove the only user-friendly way to install OE to gain whatever.
Maybe I misleading the goal of this distribution, but I thought it should be easy and user-friendly.

stefansaraev:
I am curious how 99% of ubuntu users know how to burn a CD or usb from iso.
as said, tars will be for updating. whether (and where) we write instructions how to burn an usb, is another story. those tools have no place in tars. and anyway create_installstick.bat/exe is broken already for lot of users on windows 8+, are you going to fix it  ?

chewitt:
I've been asking to remove all the scripts and pseudo-exe cruft for a while .. this is great. We currently have 4+ methods of installing which drives user confusion on which is correct (or best, which is a non conversation I'm bored of having) and the batch version frequently doesn't create working USBs. We need a README.txt file and revised installation instructions on the wiki and website to go with this change but that's a minor thing (as I presume this targets the release after 5.0.x series). The WeTek installer is Windows-only so won't work on my Mac (and it appears to be closed-source).

vpeter4:
WeTek installer is not closed source because it uses some library (GPL, free, ...). I have been told we can get the sources.

lrusak:
I have revised the instructions on the wiki for some time now. It really seems clear enough for me, people just need to be able to read.

http://wiki.openelec.tv/index.php/HOW-TO:Installing_OpenELEC/Writing_The_Disk_Image

stefansaraev:
@lrusak thanks much.

minors:
- Download the current s/release archive/release image/
- linux users may also need to umount first before burning (gnome may be configured to automount for example)

stefansaraev:
@sraue merge clock ticking. will do after one week. if no objections of course

PR #3888, opened 2015-02-08
init: fsck -y when automatic repair fails

stefansaraev:
too much rpi whinners on forums. looks like most of our users give zero fuck about manualy repairing

quick and dirty. needs testing.

pls dont merge

### DIFF
--- a/packages/sysutils/busybox/scripts/init
+++ b/packages/sysutils/busybox/scripts/init
@@ -403,28 +403,62 @@
     /bin/busybox reboot
   }
 
+  force_fsck() {
+    echo "Filesystem corruption has been detected"
+    echo "To prevent an automatic repair attempt continuing"
+    echo "press any key or power off your system within the next 120 seconds"
+    echo ""
+    read -t120 -n1
+    # The exit status is 0 if input is available
+    # The exit status is greater than 128 if the timeout is exceeded
+    if [ "$?" -ne "0" -o "$?" -gt "128" ] ; then
+      echo "repairing filesystem.."
+      echo ""
+      /sbin/fsck -T -M -y $RUN_FSCK_DISKS
+      FSCK_RET="$?"
+      if [ "$(( $FSCK_RET & 8 ))" = 8 ] ; then
+        # fubar
+        echo "Forced fsck failed. Your system is broken beyond repair"
+        echo "Please re-install OpenELEC"
+        echo ""
+        echo "Press enter to shutdown now"
+        echo ""
+        read fubar
+        poweroff
+      fi
+      do_reboot
+    else
+      echo "shutting down..."
+      sleep 5
+      sync
+      poweroff
+    fi
+  }
+
   check_disks() {
     if [ "$RUN_FSCK" = "yes" -a -n "$RUN_FSCK_DISKS" ]; then
       progress "Checking disk(s): $RUN_FSCK_DISKS"
         /sbin/fsck -T -M -p -a $RUN_FSCK_DISKS > /dev/null 2>&1
         FSCK_RET="$?"
 
-        case "$FSCK_RET" in
-          0) # no errors found
-            progress "no filesystem errors found, continuing..."
-            ;;
-          1) # filesystem errors corrected
-            progress "filesystem errors corrected , continuing..."
-            ;;
-          2) # reboot needed
-            echo "filesystem repaired, reboot needed..."
-            do_reboot
-            ;;
-          4) # errors left
-            error "could not repair filesystem, dropping to debug shell, try to run 'fsck' manually"
-            do_reboot
-            ;;
-        esac
+        # FSCK_RET is the bit-wise OR of the exit codes for each filesystem that is checked.
+        if [ "$(( $FSCK_RET & 8 ))" = 8 ] ; then
+          # operational error
+          force_fsck
+        elif [ "$(( $FSCK_RET & 4 ))" = 4 ] ; then
+          # errors left
+          force_fsck
+        elif [ "$(( $FSCK_RET & 2 ))" = 2 ] ; then
+          # reboot needed
+          echo "filesystem repaired, reboot needed..."
+          do_reboot
+        elif [ "$(( $FSCK_RET & 1 ))" = 1 ] ; then
+          # filesystem errors corrected
+          progress "filesystem errors corrected , continuing..."
+        elif [ "$(( $FSCK_RET & 0 ))" = 0 ] ; then
+          # no errors found
+          progress "no filesystem errors found, continuing..."
+        fi
     fi
   }
 


chewitt:
I'd handle this differently. Put a message on screen that says "Filesystem corruption has been detected. To prevent an automatic repair attempt continuing press any key or power off your system within the next 60 seconds" ..and if no input within that time we roll dice and fsck -y the offending volume. This way we are still pausing to warn users (which is the right thing to do, and for those in-the-know it's a chance to opt-out) but we are also automating the process for the aforementioned whining little shits who reckon the update broke their pi/box/whatever.

stefansaraev:
@chewitt good idea! I'll do

stefansaraev:
updated. still needs testing

sraue:
yes i would wait a bit more

stefansaraev:
now 120s. would be nice if you can simulate fs corruption on rpi/rpi2 and test it.

stefansaraev:
no rpi users interested ?

InuSasha:
I can test tomorrow.

InuSasha:
Testreport:
start with good storage: kodi up and running
start with destroyed storage: debug shell
start with corrupted storage:
 - without user interaction: repair run and rebooted after 2min -> kodi up and running
 - with user interaction: immeditly interupted -> power down

stefansaraev:
thanks.

> start with destroyed storage: debug shell

what you mean by destroyed? no ext4 superblock, you did dd if=/dev/zero of=/dev/mmcblk0p2 ? this PR wont fix that. we cant fix that ;)

InuSasha:
destroyed means zeroed the first 2MB.
This was only a regression test, to be sure the does not change behavior, or result in a crazy state.

stefansaraev:
updated. now handling exit code 8 from fsck (operational error). thats returned for example when sb is invalid

```
m11x ~ # dd if=/dev/zero of=/dev/sdb2 bs=1024 count=1 seek=1
m11x ~ # fsck -T -M -p -a /dev/sdb2 
fsck.ext2: Bad magic number in super-block while trying to open /dev/sdb2
....
is corrupt, and you might try running e2fsck with an alternate superblock:
e2fsck -b 8193 <device>
```

in this case, recovery is possible from backup sb.

@sraue note the "bit-wise OR exit code" comment, our (current) implementation is a bit wrong ;)

InuSasha:
Failed Tests:
- start with destroyed storage: reboot loop
   * boot -> wait 2min -> failed fsck -> reboot -> wait 2min -> failed fsck... 

Successful Tests:
- start with good storage: kodi up and running
- start with corrupted storage:
   * without user interaction: repair run and rebooted after 2min -> kodi up and running
   * with user interaction: immeditly interupted -> power down

stefansaraev:
> Failed Tests:
> - start with destroyed storage: reboot loop
>  * boot -> wait 2min -> failed fsck -> reboot -> wait 2min -> failed fsck... 

expected. and perfectly fine :)

InuSasha:
really?

not check the return code of fsck, and echo:
Storage Disk dramaticly damaged, i can't help you!!!!
Press any key to shutdown

stefansaraev:
well. I'll think about it tomorrow. but if its broken beyond manual repair, it requires reinstall. at least dropping debug shell wont help you anyway

InuSasha:
i think we can't fix every user problem.
debug shell drop, was never my intention. but a user information, that he has to act (like "reinstall oe") is good, imho. a little bit less whiners ;)

stefansaraev:
sure ^ :D

InuSasha:
- start with destroyed storage: carzy error message :D

PS:
I don't know, what you are smoking, but i want it, too :dancers:

stefansaraev:
@sraue this is ready to be merged in master. any objections ?

fritsch:
Can you change the god commit message by doing: %s/god/nullptr/ ?

stefansaraev:
already rebased. no crazy error messages.

sraue:
if it works, merge

stefansaraev:
well. tested by 3 people now. 
remember. it doesnt fix the root cause of rpi2 fs corruption as intended only for master, please dont pick for stables. there the root cause should be fixed instead.

MilhouseVH:
Really sorry I wasn't able to test this sooner, had a lot of things on my plate that didn't include potentially destroying filesystems... :)

However, I've found a problem.

The Pi1 `/flash/cmdline.txt` is configured as follows:
```
boot=/dev/mmcblk0p1 disk=/dev/sda1 ssh noram fastboot quiet
```
But when booting with PR3888, it constantly detects a filesystem problem (I've checked both filesystems in Ubuntu 14.10, and they're clean as a whistle).

I modified `init` so that the output of the `fsck` in `check_disks` is visible, which reveals the following:
```
CP437: Invalid argument
fsck.fat 3.0.27 (2014-11-12)
fsck.ext2: No such file or directory while trying to open /dev/sda1
Possibly non-existent device?
/dev/mmcblk0p1: 39 files, 27085/494612 clusters
```
and `FSCK_RET` is 8.

It appears that `/dev/sda1` is not available to be checked by the time `fsck` is being run.

If I add `debugging progress break=all` to `cmdline.txt` then I never see a problem - so this looks like a timing issue with USB devices.

I've tested the following proof-of-concept patch, which works:
```
diff --git a/packages/sysutils/busybox/scripts/init b/packages/sysutils/busybox/scripts/init
index f69f2f7..128dc8f 100755
--- a/packages/sysutils/busybox/scripts/init
+++ b/packages/sysutils/busybox/scripts/init
@@ -437,6 +437,17 @@
 
   check_disks() {
     if [ "$RUN_FSCK" = "yes" -a -n "$RUN_FSCK_DISKS" ]; then
+
+      while [ : ]; do
+        DISKS_MISSING=
+        for d in $RUN_FSCK_DISKS; do
+          [ ! -b $d ] && DISKS_MISSING="$DISKS_MISSING $d"
+        done
+        [ -z "$DISKS_MISSING" ] && break
+        progress "Waiting for disk(s)$DISKS_MISSING to become available..."
+        sleep 1
+      done
+
       progress "Checking disk(s): $RUN_FSCK_DISKS"
         /sbin/fsck -T -M -p -a $RUN_FSCK_DISKS > /dev/null 2>&1
         FSCK_RET="$?"
```
With this patch, I see the following when `debugging progress` is enabled, and the system boots successfully:
```
### Loading kernel modules ###
### Waiting for disk(s) /dev/sda1 to become available... ###
### Waiting for disk(s) /dev/sda1 to become available... ###
### Waiting for disk(s) /dev/sda1 to become available... ###
### Checking disk(s): /dev/mmcblk0p1 /dev/sda1 ###
### no filesystem errors found, continuing... ###
etc.
```
Unless there is a better solution to this problem it may be desirable to enhance the POC patch with some sort of timeout (10 seconds?) in the `while` loop, and just fail completely when the timeout is exceeded as this would suggest one or both of `boot`/`disk` are not available, which is clearly catastrophic.

stefansaraev:
not going to work because we support LABEL= and UUID=
this never worked so I prefer to remove it.
could you test http://sprunge.us/fUhT
thanks.

MilhouseVH:
I'm sure that will work (currently building), but it's going to greatly lessen the benefit of this change - anyone using SD+USB on a Pi will not have either of their disks checked, not even /dev/mmcblk0p1, because the USB devices are never available in time.

What about waiting only for /dev devices, or is there a way to convert LABEL/UUID to a device?

stefansaraev:
> 'm sure that will work (currently building), but it's going to greatly lessen the benefit of this change - anyone using SD+USB on a Pi will not have either of their disks checked, not even /dev/mmcblk0p1, because the USB devices are never available in time

/dev/mmcblk0p1 is irrelevant. it never breaks, if it breaks, you have a problem :)
see above (edited) I dont care much for non-persistent storage, fsck never worked here. I'd push this immediately if it works (but I cant test atm, maybe tomorrow) as it fixes a regression for those who use usb sticks for /storage

> What about waiting only for /dev devices, or is there a way to convert LABEL/UUID to a device?

via blkid. overcomplicated, I am not going to do that.

MilhouseVH:
Your patch works fine, many thanks.

stefansaraev:
thanks much for reporting and testing it! fixed.

weedy:
I wish you guys would consider pulling this into stable for Pis.

I hit reboot in yaste and my pi rebooted a few seconds later. This felt normal, and I did not see a lot of disk activity when it quit. The update from 4.x to 5.0.6 looked like it went smoothly but went it reset I had filesystem error. Pulling the card and running fsck it was some random crap pulled from imdb or tvdb.

Had this patch been shipped for all Pis I would have stayed on my couch instead of marching up and down 2 flights of stairs.
Please think of the lazy people, we like being fat.

stefansaraev:
> I wish you guys would consider pulling this into stable for Pis.

not going to happen. sorry

julegatti:
Why isn't it possible for Raspberry PIs?

chewitt:
@julegatti because this is an all-round hideous band-aid for the problem (the kind that will cause as much trouble as it cures) and pi developers are making positive progress on addressing the real root cause. Patience grasshopper..

julegatti:
ok! I thought it will never be fixed.
have you take a look at [this theory](http://raspberrypi.stackexchange.com/a/31391/26602)?

PR #3894, opened 2015-02-10
GCC 4.9 has support for armv7ve

Kwiboo:
GCC 4.9 added support for armv7ve, use it for cortex-a7 and cortex-a15

### DIFF
--- a/config/arch.arm
+++ b/config/arch.arm
@@ -32,12 +32,9 @@
       SIMD_SUPPORT="no"
       ;;
     cortex-a7|cortex-a15)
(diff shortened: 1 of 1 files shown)


MilhouseVH:
https://gcc.gnu.org/ml/gcc-patches/2013-12/msg01783.html

```
Armv7ve is basically a armv7-a architecture profile with Virtualization Extensions.
```

Is this likely to be of much benefit to OpenELEC?

Kwiboo:
It will not benefit other then less build logs about mismatch between march and mtune.

PR #3905, opened 2015-02-11
curl: Describe why we want it stable

fritsch:
@CvH


stefansaraev:
I hope he is happy now. I'll pick for master.

CvH:
@fritsch
yea great, at least some kind of benefit for your wasted time with the curl bump. 
These descriptions should be standard if these kind of information/bugs are available, should safe some time at the next bump.

fritsch:
They also should be standard when bumping versions within a stable release.

stefansaraev:
@CvH there is almost noo benefit for his and my wasted time. only white hair. but thanks for your comments :)

btw, where have you been in past xx weeks ?

there will be no next bump

PR #3910, opened 2015-02-12
linux: Update saa716x PCIe interface chipset patch

vpeter4:
(driver support for TechniSat SkyStar 2 eXpress)

based on https://git.kernel.org/cgit/linux/kernel/git/torvalds/linux.git/commit/?id=a2ea5561173f5c2c14e6050b261d225acd99fa08


stefansaraev:
thanks. but is it really good idea to re-introduce a peviously hidden symbol ?

[see](https://git.kernel.org/cgit/linux/kernel/git/torvalds/linux.git/commit/?id=a2ea5561173f5c2c14e6050b261d225acd99fa08) the changes to technisat-usb2 ...

PR #3912, opened 2015-02-12
Add Technotrend S2-4600 support (using Liplianin v37)

kenzodeluxe:


CvH:
Instead of adding the old frontend code wouldn't it be safer to just attach the "new" mainline m88ds3103 and m88ts2022 frontends to the dw2102? 
They are official at kernel in difference to the outdated liplianin code which is also unmaintained since years.

kenzodeluxe:
You are right; I just ported the stuff I was using on my Ubuntu box to Openelec and made some changes to not conflict with current mainline modules. The next step is using the m88* code, however I do not have the time to look into this right now. Consider this a "hack" to make S2-4600 work with Openelec.

CvH:
I tied my self to do this about a week ago on top of your ubuntu implementation, it compiles and get detected but throws a error at the end "!dvb-usb: recv bulk message failed: -110" 
Due to the lack of proper coding skills I´m unable to fix that, also not sure if the whole code make sense. Maybe it helps you.
http://pastebin.com/c7iGXF9i

kenzodeluxe:
Won't be able to look into that for a few days now, but I've seen those errors as well on boxes that do work. Are you saying that the DVB part doesn't work at all when you're seeing the error?

dhead666:
I've got two units sitting in my drawer without use, thought I might get around to try and fix this myself (I'm not a developer but I'm trying to learn more about drivers) but didn't got to it yet, so I'm grateful for your efforts here though I have two remarks:

1. I don't think openelec's devs would agree to merge such patch and maintain it when it can pushed to upstream, I suggest to post on the [linux-media mailing list](http://vger.kernel.org/vger-lists.html#linux-media) or on linuxtv channel at freenode and try to get help cleaning this patch and getting it merged upstream and only then PR the update patch here so the device will be supported in openelec 'till a major update kernel release will be out and then it could be dropped.

2. Don't PR from a master branch, this always get messy, create a separate feature branch on your repo and push from there a PR.

kenzodeluxe:
Thanks for your input, dhead666 - I think you're right and therefore will close this request. Those who might want a short term solution for their S2-4600 devices can look into this patch, and those who can wait shall wait for a nicer and better solution. I believe CvH's patch should be looked at very closely as it appears to be the way cleaner solution than mine, and I will do so as time permits.

CvH:
I have mailed TT about the support of S2-4600.

>Derzeit sind unsere Entwickler damit beschäftigt, die Treiber der TT-connect S2-4600 im Kernel zu integrieren. Wenn dieser Vorgang abgeschlossen ist, werden wir auf unserer Homepage eine dementsprechende Information herausgeben."

An developer is working at the integration - if avaible, they release it at the homepage !

Sounds good :)

kenzodeluxe:
That's perfect - thanks, CvH, for relaying that information.
Looking forward to _real_ support in the kernel coming from those who should really know how to integrate this.

trsqr:
I'm actually working on adding the TT S2-4600 support in the kernel. Shouldn't be that much of a job. I'll be travelling for the next week or so, but I'd guess before end of March the driver should be ready...

trsqr:
The driver is ready and submitted on to the linux-media mailing list: https://www.mail-archive.com/linux-media@vger.kernel.org/msg86437.html

CvH:
:+1: are you doing a pr for oe master and 5.0 (if 5.0 receiving those prs)?

trsqr:
@CvH Don't know if it could be in 5.0. I'll see, if there's interest, I can make a PR though. Otherwise, I'll just refer people to your media-build build of OE5 :-) That is, as soon as it hits media_tree / media_build...

trsqr:
For those interested, there's a current media_tree with the TT S2-4600 patches applied on top of it:
https://github.com/trsqr/media_tree/tree/s2-4600

trsqr:
Wrote a PR #4029 for this.

PR #3917, opened 2015-02-15
ffmpeg: add service_type patch

dhead666:
Adding new feature from ffmpeg master, setting service type value in mpegts stream.
I'm not sure if this feature will be included in next 2.5.x release or only in 2.6.

Specifically this is useful with [tvheadend's custom mpegts input](https://tvheadend.org/projects/tvheadend/wiki/Custom_MPEG-TS_Input), one can pipe through ffmpeg an internet radio stream to tvheadend as mpegts stream and by setting the correct service_type (digital_radio / 0x02) the stream will be auto recognize as radio channel in tvheadend.

here's an example for tvheadend's pipe input (might be geo-restricted)
```
pipe:///usr/bin/ffmpeg -loglevel fatal -i http://s4ifl.castup.net:1935/990310134-123.flv/_definst_/playlist.m3u8 -vn -acodec copy -metadata service_provider=IBA -metadata service_name=GIMEL -f mpegts -mpegts_service_type digital_radio pipe:1
```

If the scope of use seems to the developers too limited then I will drop it.

p.s. I might add that the examples in [tvheadend's wiki](https://tvheadend.org/projects/tvheadend/wiki/Custom_MPEG-TS_Input) that uses `-tune zerolatency` depends on libx264 which isn't available on openelec (I can send a separate pr for adding it).

### DIFF
--- a/packages/multimedia/ffmpeg/patches/ffmpeg-mpegtsenc-service_type.patch
+++ b/packages/multimedia/ffmpeg/patches/ffmpeg-mpegtsenc-service_type.patch
@@ -1,0 +1,137 @@
+From 5a04c938ebd232cec8e2319207aa58d9609db516 Mon Sep 17 00:00:00 2001
+From: dhead666 <myfoolishgames@gmail.com>
+Date: Mon, 2 Feb 2015 11:56:52 +0200
+Subject: [PATCH] libavformat/mpegtsenc: allow to set service_type in sdt
+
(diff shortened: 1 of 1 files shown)


dhead666:
Closing, v2.6 which will include this feature will be out soon.
https://ffmpeg.org/pipermail/ffmpeg-devel/2015-March/169506.html

PR #3929, opened 2015-02-20
Change the default settings level to "advanced"

lrusak:
OpenELEC is designed as a living room appliance and most user use it as such. This means that most of the users will be using HDMI audio and passthrough.

This allows the passthrough options to be available immediately and avoid people from asking stupid questions about how to enable passthrough audio.

### DIFF
--- a/packages/mediacenter/kodi/config/guisettings.xml
+++ b/packages/mediacenter/kodi/config/guisettings.xml
@@ -2,4 +2,7 @@
   <debug>
     <screenshotpath pathversion="1">$HOME/screenshots/</screenshotpath>
   </debug>
+  <general>
+    <settinglevel>2</settinglevel>
+  </general>
 </settings>


stefansaraev:
hiya, I am not the only one who thinks the whole kodi settilgslevel thing is meh.

https://github.com/stefansaraev/OpenELEC.tv/commit/0a8667f3398c5c1969afd47e422ce90f99f49915
https://github.com/stefansaraev/OpenELEC.tv/commit/380cd764cee0a03313f4bd0381eadfe275a6f202

MilhouseVH:
It's kind of a shame, but yeah, I think this will prove beneficial in the long run...

NedScott:
Wouldn't it be smarter to just make a PR to change those audio settings to "standard" in XBMC master? There is nothing specific to OpenELEC for this. Kodi, on all platforms, is designed as a living room appliance, dedicated to that task. The settings levels are not set in stone, and we're very open to people's input if they feel that certain settings should be exposed on standard.

Otherwise, this defeats one of the main benefits of having settings levels, which is to allow users to filter out the less common and/or advanced settings.

da-anda:
changing settings level to advanced by default in core is not an option IMO. One should not have to mess with "low level" stuff. Rather we should think if some settings should be moved into another level OR we add some first-run wizard (f.e. like on the Alienware Alpha) where they ask in 4 simple steps to 1) select desired resolution, 2) do some overscan compensation if needed, 3) select preferred audio device (where we could ask for passthrough) and 4) do an update check.

Another question would be - why don't we enable passthough by default if it's what's used by most?

PR #3931, opened 2015-02-21
Updated ouya controller patch

lrusak:


stefansaraev:
thanks. on frm me. @sraue do you want this in ?

sraue:
yes i want

lrusak:
@sraue I didn't do arm configs because you won't be able to use the controller in Kodi on those platforms anyway.

sraue:
@lrusak then it should be disabled, not tested the patch yet but i thing after merge the build asks... would be nice if kodi would support controllers on arm plattforms, maybe another thing where libinput is usefull?

lrusak:
Added to RPi's configs as they already have the other controllers set as well.

PR #3933, opened 2015-02-21
xpad rework for 3.19

lrusak:
Unfortunately I still do not own an xbox controller, so I am unable to test. It does compile though.

I am just putting this here if someone wants to test. hint hint @zzattack

I wish we could just have a decent xbox driver :(


stefansaraev:
thanks. ok from me. @sraue do you want this in ?

sraue:
yes

zzattack:
I'll start a build and let you know how it works, and whether it the issues from https://github.com/OpenELEC/OpenELEC.tv/pull/3672 are gone

zzattack:
Seems no different from #3672 

imag0688

zzattack:
Should we revert this? Maybe the user space driver can be used instead?

sraue:
@zzattack when you can see this error?

zzattack:
When rebooting the panic hit rate is over 50%. Cold boot about 10% for me.

stefansaraev:
then this should be removed. and we wait for those patches to hit mainline. @lrusak ?

bhabba:
Hi@all

I'm very happy to see your efforts in integrating the xpad LED fixes into the official OE builds. Concerning the kernel panic my observation with @zzattack's [build (the 2nd one)](https://github.com/OpenELEC/OpenELEC.tv/issues/2607#issuecomment-68019067) is that the kernel only panics under very special circumstances which I described [here](https://github.com/OpenELEC/OpenELEC.tv/issues/2607#issuecomment-66147500).

@zzattack you say the panic hit rate in your setting is over 50% when rebooting. Do you manually power off/on the controller during reboot? Because I could only run into the kernel panic **under the following three very specific circumstances:**

1. When I initiated a warm reboot and powered off the controller at a very specific moment (after openelec was shut down but before the bootloader came up).
2. When I shut down my HTPC (but didn't disconnect it from current) and afterwards powered it on again. *In this situation the kernel panic can be avoided by manually powering off the controller after the shut down of the HTPC (normally the controller stays on after a shut down).*
3. After cold boots I only managed to run into the kernel panic when I had powered on the controller before the HTPC (but after the HTPC and wireless receiver were connected to the current).

In all other situations there was no kernel panic. All of the described behaviour was completely reproducible. Of course others might experience a different behaviour... I have a Zotac ZBOX ID41 which delivers power through its USB ports when connected to the current but powered off (I believe this fact matters!).

lrusak:
I'm not really sure what to do.

We have three options as I see it.

1) We ship the stock xpad driver (which I believe works but is broken
because the lights blink)

2) we ship the xpad driver used from steamos (which I believe @zzattack
said is broken in other ways)

3) we incorporate the user space driver xboxdrv as an addon.

I could look into making the xboxdrv into an addon in the unofficial repo,
however, I am not able to test it.

Opinions @sraue and @stefansaraev

stefansaraev:
whatever you think is best. I would not waste time on this (my 2c)

bhabba:
To be clear: The xpad driver in OE 5.0.4 (which is the stock xpad driver I assume) indeed works perfectly except the blinking LED.

zzattack:
@lrusak: how about "4) fix the issues from 2) ourselves"?
I'll take another stab at it. Perhaps I'll have more luck this time.

zzattack:
@bhabba: I didn't touch the controller when rebooting. It remained on, but I'm not sure if the receiver did too. 
You say you can only trigger the KP by turning the controller on at a specific moment. It seems that for me this moment may not be all that specific.
I don't use the official 360 wireless receiver, and my NUC doesn't power USB periphals when it's shut down. As you suggested, I agree that it's not unthinkable these differences could affect the hitrate of the panic.

blobfish:
Hello,
I became disillusioned with my xbox controller and the lights. I created a patch from the stock kernel xpad.c to the latest steamos xpad.c. Using that patch during build has cured my controller problems. No more: blinking lights, triggers causing havoc, sometimes volume doesn't work....probably some more I can't think of right now. I was able to cause 1 kernel panic at startup using bhabbas' procedures. Couldn't repeat after about 15 tries. I have a zotac box with an official wireless receiver.

This post is probably noise, but wanted to share my experience. Thanks

nickaardo:
Hey guys,
OE 5.0.6, the Xpad driver still works perfectly except the LED.
Please don't abandon this issue :( An xbox 360 controller is by far the best remote for Kodi :/

chemstar:
it seems valve has fixed 360s led problem
http://www.phoronix.com/scan.php?page=news_item&px=Xbox-Wireless-Linux-4.2

maybe it can be backported to kernel 4.1

zzattack:
doesn't seem like anything we haven't already tried was changed, so I doubt
this solves the panics we encounter, but at least there's finally some
progress

PR #3939, opened 2015-02-22
Revert "projects/WeTek_Play/linux: enable CONFIG_AML_NAND_KEY"

stefansaraev:
This reverts commit 4fa2adb25820d5a076ea71a92a447b7841c682e4.

this fixes ethernet mac address changing to some nonsense

please, never revert my commits where I disabled something. especialy if it affects network, without pinging me

62e3640621cdff016f91087be7958f09045a6472


stefansaraev:
hum, this looks to be silently fixed in aml kernel, now after read_mac_from_nand() it has the original mac address. ping @codesnake ?

or is it because I've had kernel without AML_NAND_KEY runing, so correct value was already written to the nand, and it would (again) break miserably on new installs ?

in all cases,I would like to have the following in:
```
diff --git a/drivers/amlogic/ethernet/am_net8218.c b/drivers/amlogic/ethernet/am_net8218.c
index aec2980..250d9d7 100755
--- a/drivers/amlogic/ethernet/am_net8218.c
+++ b/drivers/amlogic/ethernet/am_net8218.c
@@ -926,9 +926,6 @@ static int aml_mac_init(struct net_device *ndev)
        printk("--1--write mac add to:");
 
        data_dump(ndev->dev_addr, 6);
-#ifdef CONFIG_AML_NAND_KEY
-       read_mac_from_nand(ndev);
-#endif
        printk("--2--write mac add to:");
        data_dump(ndev->dev_addr, 6);
        write_mac_addr(ndev, ndev->dev_addr);
```

I dont trust this ethernet driver.

codesnake:
No, there were no fixes in kernel for that. The problem with AML_NAND_KEY is that it affects NAND partition layout and is required for booting from NAND. I'll try to fix that in kernel.

stefansaraev:
I am fine with AML_NAND_KEY as long the ethernet driver dont do nonsense(see above)

should I merge this for now ?

codesnake:
Yes, you can merge this now. I'll return the AML_NAND_KEY back after I fix ethernet driver.

stefansaraev:
thanks ;)

PR #3945, opened 2015-02-23
linux: implement limited color range for amd dce6+ asics (agd5f)

fritsch:
This implements Limited Color Range for AMD gpus. This is a long awaited feature via: https://bugs.freedesktop.org/show_bug.cgi?id=83226

It needs testing as I currently don't have access to AMD hardware.

### DIFF
--- a/packages/linux/patches/3.19/linux-999.10-amd-limited-color-range.patch
+++ b/packages/linux/patches/3.19/linux-999.10-amd-limited-color-range.patch
@@ -1,0 +1,273 @@
+From 7b3d1747b3345ecc7b1d49b7b920b7ada5e1eb36 Mon Sep 17 00:00:00 2001
+From: Alex Deucher <alexander.deucher@amd.com>
+Date: Mon, 23 Feb 2015 10:11:49 -0500
+Subject: [PATCH 1/2] drm/radeon: add an output_csc property
+
+This adds the drm property for output csc (e.g.,
+transform of output display colorspace).  Currently
+only common ones (TV RGB, BT.601, BT.709) are supported,
+but bypass and tv rgb are really the only useful ones at
+the moment.  Additionally we could expose a user adjustable
+matrix in the future.
+
+This commit just adds the property.  The hw support will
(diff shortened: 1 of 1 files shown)


fritsch:
jenkins build this please

sraue:
jenkins is out of order :-) you have built and tested?

fritsch:
Nope, I only picked it and applied it to 3.19 - don't have Generic build
env set at the moment. If you have capacity to do a build for that one - I
think we get some AMD users to test it.

2015-02-23 21:38 GMT+01:00 Stephan Raue <notifications@github.com>:

> jenkins is out of order :-) you have built and tested?
>
> --
> <https://github.com/OpenELEC/OpenELEC.tv/pull/3945#issuecomment-75626629>.
>



-- 
                   Key-ID:     0x1A995A9B
                   keyserver: pgp.mit.edu

sraue:
ok, will build tomorrow if i am able, if you are faster merge

fritsch:
Alex posted another patch (on top) but that needs way more backports from his experimental tree. We are good to go at the moment with those two.

We get all this work with 4.0 kernel.

sraue:
can you rebase to kernel 3.19.1, or should i do byself?

fritsch:
I am not fully sure which way we should go here ... the above patches only support setting the values via e.g. xrandr - Will OpenELEC 5.2 be shipping with 3.19? or do you already plan with 4.0?

fritsch:
Btw. you can just pick those as they are - no need to rebase.

sraue:
i think 5.2 will be still based on 3.19
i only asked because you prefer if i merge via the github webgui, we use 3.19.1 now so i have to merge manually

sraue:
added in https://github.com/OpenELEC/OpenELEC.tv/commit/9a57d80a114fd0fafd506bdb1ac0e55937c81336

fritsch:
Here is an example on howto use it: 

xrandr --output HDMI-0 --set output_csc tvrgb

PR #3946, opened 2015-02-24
lcdproc: don't build parallel

MilhouseVH:
Builds sometimes fail trying to cp files that don't yet exist when using more than -j1.


stefansaraev:
thanks, could you please move the pre_make_target block before post_makeinstall_target for better readability

MilhouseVH:
Done.

Issue #3503, opened 2014-09-25
STK1160 support does not work


tociek:
Hi,

STK1160 support does not work in 4.1.6. I do have /dev/video0, but when I start hyperion with the grabber, I get a lot of "frame too small errors"

Frame too small: 686552 != 691200
Frame too small: 687444 != 691200
Frame too small: 686964 != 691200
(...)

lsusb and dmesg output can be found here:
https://github.com/OpenELEC/OpenELEC.tv/commit/d66494b9b4a998ac49fd23a7f316a09584b47b9c

Let me know if I can deliver some more details. I can also send you STK1160 for testing if needed.

Jacek

slovdahl:
Is there any progress on this issue? It would be awesome to have working STK1160 support out of the box! OpenELEC 5.0.0 and 5.0.1 on my 1st gen model B RPi isn't able to detect the video grabber at all, nothing shows up in dmesg or lsusb.

stefansaraev:
there is none. if you find out how to fix, please send us a PR. 
my opinion is still we shouldnt support this hardware at all. it has been added, I disagreed, then it didnt work and whoever added it doesnt care

tociek:
Hi, there is a solution developed by Endo - he added frame scaling to the driver. 

Here is the patch: img.lightberry.eu/download/linux-049-STK1160-addFramescaling.patch

I agree that it would be great to have it included in official OE release (even better in official linux kernel)

dalehamel:
i can confirm the patch provided by @tociek works in OE 5.0.3, which is what the latest version of rasplex is based on. I recommend merging it

dalehamel:
@stefansaraev here's the patch that fixes the issue, verified working against OE 5.0.3 and master branch kernel (3.19)

tociek:
Guys, is the patch in 5.0.8? it doesn't look like it - scaling does not work.

mmodler:
Is there any reason why the patch hasn't made it into currect openelec? I have the frame to small issue.

Is there any chance the patch gets applied?

PR #3948, opened 2015-02-24
Add stk1160 framescaling patch to RPi and RPi2

dalehamel:
This adds support for stk1160 framescaling patch, which we can confirm solves https://github.com/OpenELEC/OpenELEC.tv/issues/3503

Confirmed that this solves for our own issue https://github.com/RasPlex/RasPlex/issues/350

I would suggest adding this as patch in packages/linux, but I have only tested this on RPi1 and RPi2 so that's why I'm only adding the patches to these projects.

FYI I'm also working on an actual package.mk for hyperion, but since it requires Qt libs as dependencies it involves fetching some blobs from raspbian. Happy to contribute this, and OpenELEC.settings addon patches once completed.

@stefansaraev @sraue @fritsch  for review 

cc @tociek

### DIFF
--- a/projects/RPi/patches/linux/linux-02-STK1160-addFramescaling.patch
+++ b/projects/RPi/patches/linux/linux-02-STK1160-addFramescaling.patch
@@ -1,7 +1,28 @@
-diff -uNr a/drivers/media/usb/stk1160/stk1160-core.c b/drivers/media/usb/stk1160/stk1160-core.c
---- a/drivers/media/usb/stk1160/stk1160-core.c	2014-11-20 15:40:34.000000000 -0700
-+++ b/drivers/media/usb/stk1160/stk1160-core.c	2014-11-26 05:04:33.096241000 -0700
-@@ -418,6 +418,7 @@
+From edf8977b62ae86859cccf3d0ea4883333ea4d138 Mon Sep 17 00:00:00 2001
+From: Dale Hamel <dale.hamel@srvthe.net>
+Date: Tue, 24 Feb 2015 10:31:04 -0500
+Subject: [PATCH] Add framescaling support to stk1160
+
+Implements frame scaling support for stk1160 to support format changes instead of a static frame size.
+
+This is effectively a dumb sampling, and could perhaps benefit from being an averaging instead.
+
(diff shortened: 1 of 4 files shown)


fritsch:
Please send this patch upstream where the driver is originally maintained. If it's already coming from there - please make sure to keep the original author.

I will be afk for the rest of the week, so won't have time until sunday.

sraue:
@dalehamel where this patch comes from and who could maintain this if we update the kernel? if this will be maintained and will be committed upstream i have no problem with including this

dalehamel:
@fritsch @sraue are you more concerned with stability, or the maintenance burden of a new patch?

The only code touched here is related to the stk1160 anyways - in terms of maintenance the worst situation you end up in is the patch stops working, in which case it's in the same state it is now. If the patch stops applying, it can just be removed. Since it's a kernel module, it's not loaded unless requested by the user anyways.

Is it the OpenELEC policy not to accept kernel patches, and only take them from upstream? I'll track down the author and try to get the patch mainlined. If I make progress I'l update this PR.

dalehamel:
I've gotten ahold of the author of the patch, and I have submitted it the original authors of the stk1160 driver on his behalf with permission.

For the time being, you can consider me to be the maintainer of this patch if that helps. I've read it and understood that it implements framescaling (a feature listed as a TO DO by the original author) by doing a "dumb sampling" of the image for efficiency reasons.

Let's leave this PR open until I get comments from the original author and reviewers of the stk1160 driver, as I may update the patch with their recommendations etc.

-Dale

fritsch:
Please redo the patch, add the original author. We cannot take code snippets, where it is not fully clear they come from. It needs documentation in the patch, including email adress.

dalehamel:
@fritsch is this updated format more acceptable?

knowncolor:
@tociek I don't think this is currently included in the latest OE, since I am getting the "Frame too small" messages. It doesn't look like it's included in the repo either (https://github.com/OpenELEC/OpenELEC.tv/tree/master/projects/RPi2/patches/linux). Do you by any chance have an OE image for RPi2 with this patch installed? Will save me installing linux on my laptop and compiling from scratch...

PR #3949, opened 2015-02-24
New package - libinput

lrusak:
I'm really doing this as a test. I'm not looking to merge this anytime soon.
Fedora is making libinput default in F22 and I wanted to try it.

This isn't really needed now as it is aimed for touchpads, but it can take care of everything else too. 
It will be needed for the wayland transition.

My testing consisted of moving a mouse and typing on a keyboard. Nothing fancy. It other people have other devices, like usb remotes and whatever, please test them.

~~I patched out mtdev as it is a dependency of libinput. It is just a multitouch driver and I figured it isn't needed. It can always be added later.~~

libevdev needed to be built shared because of leaky symbols. There may be a better way to deal with this.
see, https://bugs.freedesktop.org/show_bug.cgi?id=82292

I put libinput in the wayland directory because that's how it is on freedesktop.
see, http://cgit.freedesktop.org/wayland/libinput/

Yes, it does work, http://sprunge.us/NhBd
```
[3121428.819] (**) flirc.tv flirc: Applying InputClass "libinput"
[3121428.819] (II) Using input driver 'libinput' for 'flirc.tv flirc'
```


sraue:
great :-) hope kodi could use libinput too in the future

lrusak:
Updated and added dependency, mtdev

PR #3951, opened 2015-02-25
New script - scripts/uninstall

lrusak:
Just thought I would share this.

It's helpful when building new packages to test different config options (shared, static). It allows easy removal of packages from the toolchain and sysroot.

Basically just scripts/build but opposite. Not sure why this doesn't exist already but maybe I'm just missing something.

```
PROJECT=RPi ARCH=arm ./scripts/uninstall parted
```
```
PROJECT=RPi ARCH=arm ./scripts/uninstall parted:host
```

### DIFF
--- a/scripts/uninstall
+++ b/scripts/uninstall
@@ -1,0 +1,105 @@
+#!/bin/bash
+
+################################################################################
+#      This file is part of OpenELEC - http://www.openelec.tv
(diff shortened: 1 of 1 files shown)


stefansaraev:
ok from me if it works and makes your life easier ;)
however, I wouldnt recommend using it on major toolchain changes.

ping @sraue

PR #3956, opened 2015-02-27
Add support for installing OE into NAND memory

codesnake:
* Changed NAND partition layout. Now it is about of 3GB available for user data.
* Kernel and system image now both reside as KERNEL and SYSTEM files on system partition that uses UBIFS filesystem. Init script was slightly modified to allow mounting of UBIFS partitions.
* It makes possible to build a ZIP installation package to install OE into NAND from recovery using the following command: ARCH=arm PROJECT=WeTek_Play make amlpkg

Later I will add to the bootloader an ability to import the uEnv.txt file as in Cuboxi.


stefansaraev:
thanks. I have some questions:

1. what happens with aml ethernet driver
2. is it (still) possible to install onto sdcard after this PR. you removed bootargs patch
3. I dont like having 5MB+ blobs in git. @sraue should those be packaged and downloaded at build time.

sraue:
thanks much for this PR, i will look in this asap, but maybe i can do after weekend

> is it (still) possible to install onto sdcard after this PR. you removed bootargs patch
this should be still possible

> I dont like having 5MB+ blobs in git. @sraue should those be packaged and downloaded at build time.
yeah i would create a package for such binary stuff (the same for the wetek dvb driver) i will change this after testing and looking in this PR

codesnake:
@stefansaraev 
1. I checked and get_aml_key_kernel() in read_nand_from_nand() always returns -22, so it never rewrites the MAC address. Correct me if I'm wrong.
2. Yes it is possible. Bootloader now passes the required kernel params. Also as I noted I'll add uEnv.txt reading shortly.

stefansaraev:
> 1. I checked and get_aml_key_kernel() in read_nand_from_nand() always returns -22, so it never rewrites the MAC address. Correct me if I'm wrong.

well, dunno, we've had this aml ethernet issue before, did you test with brand new device ? 
however, I'd recommend using the patch from https://github.com/OpenELEC/OpenELEC.tv/pull/3939#issuecomment-75432692 
I dont trust this driver. I'd want it fixed so it doesnt read nonsense from nvram.,

> 2. Yes it is possible. Bootloader now passes the required kernel params. Also as I noted I'll add uEnv.txt reading shortly.

cool. with updated u-boot or would it work with older (stock) u-boot ? I did not check default uboot env

codesnake:
1. That kernel is not for WeTek Play only. There are also Meson8, Meson8b and MesonM8n. Removing a line is not a fix. Personally I never had problems with MAC address. But If you feel safer with the patch, of course you can add it. But instead of just removing the read_mac_from_nand() call, you can just uncomment the //#undef CONFIG_AML_NAND_KEY line above.
2. Of course tt will not work with Android u-boot. It should be changed as well.

stefansaraev:
>That kernel is not for WeTek Play only. There are also Meson8, Meson8b and MesonM8n. 

we keep it as OE specific patch. for wetek_play project. we have 17 patches, one more doesnt hurt.
and well, I had the mac address change problem once, and at least two kodi devs, too.

> Of course tt will not work with Android u-boot. It should be changed as well.

o_O. whatever that means ;)
so, after merging this, do we require users with existing SD installs to do something (update?), manualy?

codesnake:
Users of existing SD installs should have either install the OE with updated u-boot into NAND, or wait when Android firmware with updated u-boot will come out. If you can propose another solution to have single build (update TAR) for both NAND and SD-card installations, given the fact that they should get different kernel params, I'll be glad to hear from you.

stefansaraev:
I have none, but I fear it is not good idea to force people install to nand, I personaly want to have "dual boot" :)
and. when we break something and the box dont boot after update, it's not as easy to recover as with SD install. I broke my installation 3 times in the past month, once with absolutely unrelated udevil change :)

so we wait. do you know when updated android will be released?

sraue:
added in https://github.com/OpenELEC/OpenELEC.tv/commit/5adb752317f4ba4e23e03cfa63c9e4bd65b51823 thanks much

PR #3960, opened 2015-02-27
IMX: Linux: Backport AVM Fritz 2 Stick support - requested by users

fritsch:
Sadly, we cannot upgrade our imx kernels, here is a backport of the Fritz WLAN v2 driver. Cherry-picked changes.


fritsch:
These patches were verified working by user ph1 on OpenELEC IRC

PR #3967, opened 2015-03-01
RPi2: build also module sp2

trsqr:
Module sp2 was not built for RPi2. The module is required for TechnoTrend CT2-4650 CI USB tuner.

This issue does not exist on 3.17 or 3.19 kernels, so only RPi2 was affected.


trsqr:
Confirmed working by an end user:
http://openelec.tv/forum/81-dvb-c-support/67950-technotrend-ct-4650-support?start=24#133719

stefansaraev:
yay. missing autoselect in 3.18 ?

PR #3969, opened 2015-03-02
Update Package - xf86-video-nvidia 346.59

lrusak:
It's time to update it. 

This is the latest in the long lived branch

http://www.nvidia.com/Download/driverResults.aspx/83686/en-us


chewitt:
nvidia driver branch changes should not be done within an OE release series as it will change support for cards that are currently working, i.e. 346.47 in OE v5.0.6 would drop support for ION 1st gen cards that work fine with 340.76 in OE v5.0.5. I appreciate we are starting to see users with 4k sets and GTX9XX cards that need the newer driver, but there are significantly more active users with older working cards than users with bleeding edge budgets. We also need to think about nvidia driver changes in context of 32-bit support, as if/when we drop i386 (and the legacy driver it contains) we will also drop support for a large number of older, common, and perfectly usable nvidia cards.

stefansaraev:
ok for master only

@chewitt thats not our problem. those users can run some older release or switch to ubuntu. we cant and we wont support their hardware forever.

lrusak:
@chewitt this was only intended for master ;)

vpeter4:
Simple solution: 2 images. One with older driver and one with newer. Both can share whole build folder just final image would be different. Or even one image with both drivers inside.
If OpenELEC would care for "older users"...

fritsch:
Basically we can stop caring - when the vendor that got a whole lot of
money from those users has stopped caring already.

2015-03-02 20:14 GMT+01:00 Peter V. <notifications@github.com>:

> Simple solution: 2 images. One with older driver and one with newer. Both
> can share whole build folder just final image would be different. Or even
> one image with both drivers inside.
> If OpenELEC would care for "older users"...
>
> --
> <https://github.com/OpenELEC/OpenELEC.tv/pull/3969#issuecomment-76786658>.
>



-- 
                   Key-ID:     0x1A995A9B
                   keyserver: pgp.mit.edu

vpeter4:
I don't agree if hw is fine and sw is fine. Just need to build it. It's not the same as story with Broadcom CrystalHD card which is not supported in sw either.
But this is just my opinion and doesn't want to start any big discussion.

stefansaraev:
no. it's exactly the same story as broadcom chd. if vendor doesnt care, why should we, seriously ?

vpeter4:
Seriously? Because OpenELEC handles customers better than Nvidia. Or at least it could and it should if conditions are suitable. Like if driver still compiles/works when kernel version changes. If it doesn't it's different story.
But with approach you took openelec is the same ****** company/organization/whatever like many others out there. Which should not happen in my opinion (which doesn't count anyway).

chewitt:
If nvidia didn't care about users they wouldn't continue to bump 17x.xx era drivers to bring Xorg support to cards sold 10+ years ago, but they do. I consider them to be a responsible vendor, and despite all the fuss made about it being a closed-source and could-use-a-diet driver it is by-far the most stable and least troublesome driver we use. The sole annoying thing is that their release cycle has resulted in a situation where 304.xx, 340.xx and 347.xx are required for OE to support a broad range of their cards that are in active daily use. Our attitude towards users on some aspects of hardware support (in particular wireless drivers .. despite their noted charms) is increasingly aggressive, disrespectful and not aligned with the OE's goal of being a simple and n00b friendly distro.

fritsch:
Support needs to be done by real humans. And wireless drivers, especially from the "usual suspects" need ajusting, porting to our latest and greatest kernels we use ... which is a big nightmare.

So - now we should release 3(!) 386 builds to support old hardware? Additionally reintegrated fglrx legacy version for the poor AMD 4200 Users?

We need to shrink our "ideal world" to what we can afford. Writing walls of text is one thing, but really caring about bugs (hardware failures we cannot reproduce, cause hw is long not existent anymore) or alone keeping our features in a working state (*wink* at conman and all that related nightmare) is just not doable at the very moment.

Therefore: If we support something, we do it correctly. If we cannot afford it - we don't do it.

stefansaraev:
@vpeter4 we dont have customers. we dont sell hardware, we do it in our spare time. we dont have competitors and we can afford to be s*** company/organization.

@chewitt then we dont remove i386 but instead rename it to Legacy and "support" it for next 30 years. but I dont want to. you support it, you take care of the addons, you take care to fix legacy driver when new kernel/xorg is out and we want it bumped (because sometimes nvidia lags for months). I am sure @vpeter4 will help, too

MilhouseVH:
Not working with Revo3700 (ION2) - just getting "failed to start xorg". This is a clean build of master, with this PR. Current master without this PR is working perfectly.

http://sprunge.us/JDYi

image

lrusak:
It even tells you that they have moved support to the legacy driver
```
[    7.322368] NVRM: The NVIDIA Second Generation ION GPU installed in this system is
[    7.322368] NVRM:  supported through the NVIDIA 340.xx Legacy drivers. Please
[    7.322368] NVRM:  visit http://www.nvidia.com/object/unix.html for more
[    7.322368] NVRM:  information.  The 346.47 NVIDIA driver will ignore
[    7.322368] NVRM:  this GPU.  Continuing probe...
```
So what are we going to do? Not support any new cards that come out?
We can keep using the nvidia-legacy driver in the i386 project for now. Then when time comes to drop i386 all together we can decide what we want to do.

MilhouseVH:
Going to be a shame to kill off perfectly serviceable hardware because of this one issue. Hopefully there's a solution that can mean a few more years of ION2, without too many headaches...

vpeter4:
Totally agree with @MilhouseVH. And it seems I don't see the real problem which is? Just because we move to higher driver we can't use old hw anymore? Still don't see why not provide 2 builds. I see no problem of doing this. At least not at this time. For future no one knows.
@stefansaraev: maybe we don't have customers but we do have "customers" in some other way. Hope you don't thread your real customers at work in a same way as this. But that's another different story.

sraue:
we have actually only 2k Generic i386 users which dont run on x86_64 hardware, so it makes sense to drop the i386 support with OE-6.0 (possible in a half year). Instead this version we really could create 64bit legacy builds, which shares the addons with the normal generic builds. I dont like much this solution, like @stefansaraev also dont does... we cant do anything if a vendor stops the support for systems which are still "perfect" to run OpenELEC, but this is now known for more then a year... but i think a 64bit legacy build starting with OE-6.0 we maybe can maintain for the next 2 years or so (if there will be help from ppl to test this regulary).

MilhouseVH:
I have an interest in keeping ION2 viable for a bit longer (it's my main lounge system) so I'm happy to help out with testing as and when necessary.

lrusak:
Is it not possible to include both nvidia drivers and just use the nvidia-legacy driver as a fallback if the nvidia driver fails?

stefansaraev:
> Is it not possible to include both nvidia drivers and just use the nvidia-legacy driver as a fallback if the nvidia driver fails?

absolutely nope. overlapping pci ids. I dont wan this again.

fritsch:
Reminds me of fglrx and fglrx-legacy .... yeah, was a real nightmare back at that time.

lrusak:
How about this (When i386 gets dropped first of course) lrusak/OpenELEC.tv@389796420ff65bb59fa1f10915d8b7dfc1b5ee60

lrusak:
Rebased and updated to 346.59

Still waiting on #4075

PR #3977, opened 2015-03-03
linux: fix TechnoTrend CT2-4400v2 clock for RPi2

trsqr:
RPi2 was still having no support for TechnoTrend CT2-4400v2 (which was introduced in kernel 3.19). This small patch enables gapped clock for CT2-4400 and CT2-4400v2, so that both of them work properly.


trsqr:
Too many platforms with different kernels... Happy when all of them run 3.19!

stefansaraev:
thanks.

> Too many platforms with different kernels... Happy when all of them run 3.19!

that will never happen ;)

PR #3985, opened 2015-03-07
linux: Use legacy turbo (fixes BYT full freeze)

fritsch:
This fixes: https://bugs.freedesktop.org/show_bug.cgi?id=88012 and https://github.com/OpenELEC/OpenELEC.tv/issues/3726

I wait until pushing a fix for master - as we can see this way, what upstream decides to push in their stable branches.


fritsch:
Here is a Version 5.0 branch based testbuild: https://dl.dropboxusercontent.com/u/55728161/OpenELEC-Generic.x86_64-devel-20150307090204-r20103-g030e2a8.tar

PR #3987, opened 2015-03-07
kodi.sh: Do not overwrite kodi_crashlog.log, keep older crashlogs.

ksooo:


stefansaraev:
we can end up with thousands of crashlogs. need something to clean up old..

MilhouseVH:
Seems like this should have a limit, either based on number of logs (keep last 12?) or date (keep a weeks worth?) I had a system go into a crash loop unnoticed for several hours the other day - I'd have had hundreds of crash logs with this PR.

stefansaraev:
```
rm -f $(ls -1t $CRASHLOG_DIR/kodi_crashlog_*.log | tail -n +11)
```

sorta.

ksooo:
Kodi has more than one scenario where it bombs. ;-) Some (rare) crashes do actually occur if I'm not at home and my family uses the system. Collecting all the different(!) crashlogs gives me the chance to analyze them all at some point in time, not just the latest.

ksooo:
Now keeps only youngest 10 crashlogs.

ksooo:
@stefansaraev btw, where do the exit code definitions used in kodi.sh come from?

stefansaraev:
@ksooo https://github.com/xbmc/xbmc/blob/master/tools/Linux/kodi.sh.in#L159

stefansaraev:
I am fine with this now (10 crashlogs). @MilhouseVH ?

MilhouseVH:
10 is good, I'm OK with that. Thanks both.

PR #3997, opened 2015-03-12
linux: enable all buttons on TiVo Slide Pro remote

foresto:
The linux kernel has supported the TiVo Slide remote control for some time, but does not recognize the USB ID of the newer Slide Pro. This patch adds the missing data structures so the newer remote will be recognized by the driver, thereby allowing the TiVo, LiveTV, and Thumbs Up/Down buttons to be mapped with a hwdb file.

### DIFF
--- a/packages/linux/patches/linux-010_tivo_slide_pro.patch
+++ b/packages/linux/patches/linux-010_tivo_slide_pro.patch
@@ -1,0 +1,57 @@
+From 716d6bc9f73bd366d4b64b473055af79810d67fc Mon Sep 17 00:00:00 2001
+From: Forest <web11.forest@tibit.com>
+Date: Thu, 12 Mar 2015 23:43:12 -0700
+Subject: [PATCH] hid: enable all buttons on the TiVo Slide Pro remote
+
+The linux kernel has supported the TiVo Slide remote control for some time,
+but does not recognize the USB ID of the newer Slide Pro. This patch adds
+the missing data structures so the newer remote will be recognized by the
+driver, thereby allowing the TiVo, LiveTV, and Thumbs Up/Down buttons to be
+mapped with a hwdb file.
+
+Signed-off-by: Forest <web11.forest@tibit.com>
+---
+ drivers/hid/hid-core.c | 1 +
+ drivers/hid/hid-ids.h  | 1 +
+ drivers/hid/hid-tivo.c | 1 +
+ 3 files changed, 3 insertions(+)
+
+diff --git a/drivers/hid/hid-core.c b/drivers/hid/hid-core.c
+index 1d0f2b6..722a925 100644
+--- a/drivers/hid/hid-core.c
++++ b/drivers/hid/hid-core.c
+@@ -1980,6 +1980,7 @@ static const struct hid_device_id hid_have_special_driver[] = {
+ 	{ HID_USB_DEVICE(USB_VENDOR_ID_THRUSTMASTER, 0xb65a) },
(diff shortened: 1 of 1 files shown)


stefansaraev:
the patch should be valid git-formatted patch (with author (you) and date)
other than that, ok from me. thanks

foresto:
@stefansaraev, what's your favorite way to make a git-formatted patch for files that aren't tracked by git?

foresto:
Thanks, @stefansaraev. I reformatted the patch. How does it look now?

stefansaraev:
perfect :)

ping @sraue

sraue:
@foresto can you move the patch in the versioned subdir of the patches? i am sure it will not apply with all our kernel versions (3.10, 3.14,3.17-19), ist this something you want in our 5.0 builds too?

foresto:
I think it might very well apply with all those kernel versions, but I'd be happy to move the patch to another directory if you like. Which directory do you suggest?

By the way, I submitted my patch upstream last night.

And yes, I would like to see it in the OpenELEC 5.x builds too. That would allow me to use my remote without having to patch and rebuild every time you folks release an update. :)

sraue:
if you can test a Wetek build and imx6 build (if it builds), would be nice

foresto:
I've been using it with a CuBox-i for a few days now, so I think it's safe to say it's already tested with imx6.

I'm trying to get a successful wetek play build right now. I can see that my patch applies, but the overall build fails somehwere in libsystemd. It's probably unrelated to my patch. I'll git pull and try again.

foresto:
Just to be sure I'm running the build correctly, is this the right command line?
PROJECT=WeTek_Play ARCH=arm make

sraue:
yes its right, simply be sure it applies, no need for a full build, if it applies for 3.10,14,17-19 then its ok there, tell me and i will merge this patch

foresto:
How do I tell the build system which kernel version to use?

stefansaraev:
applies fine on all supported projects. when it breaks (or is added upstream) we move it to project specific dirs.

foresto:
Thanks, @stefansaraev. Do I need to do anything in order to have it included in the next 5.x release?

sraue:
no, i will add this (and will see if it applies for all kernel versions there)

sraue:
and thanks much :-)

foresto:
Cheers!

stefansaraev:
@sraue as said, it applies, I tested before merging.

foresto:
For the record: My patch has been accepted on a development branch of the linux kernel, so assuming it makes it through Linus, we ought to see it upstream in linux 4.0 or 4.1.

stefansaraev:
very nice @foresto if accepted in -next, it will be in 4.1 (too late for 4.0)

foresto:
Looks like it's in 4.0 rc5. :)

foresto:
It has just been added to the linux 3.12 stable tree, too.

Lusephur:
Is this now working with linux 4.3? 
I've compiled openelec from the master branch (os version 7) and not all buttons are working. Has the patch been reverted?

foresto:
It's working fine on OpenELEC 6.0 official. Note that this patch only supplies the kernel part of the picture. To actually use the extra buttons on this remote, you still have to map the key codes, and then assign those to functions in Kodi.

Instructions here:
http://kodi.wiki/view/TiVo Slide

Please take any further configuration discussion to the forum; it doesn't belong in this pull request.
http://forum.kodi.tv/showthread.php?tid=79616

Lusephur:
It is working now.  The fault lay with myself and the changes needed in the hwdb file.  The first two lines had to be changed, as per the updated wiki entry.

PR #4000, opened 2015-03-12
busybox: enable all features for 'find' applet

### LINKED ISSUE #3999: OpenELEC 5.0.5 only has a striped down version of BusyBox
Hello,
i am currently using OpenELEC 5.0.5 for RPi1 and RPi2. i tried to rename hundrets of video and audio files accros the folders, to get rid of annoying parts of the filenames...
e.g.:
from "title (360p).mp4" to "title.mp4" and
from "title (256bps).mp3" to "title.mp3".

for that i loged in via SSH to OpenELEC and i tried some one-liner i found on the internet to do the job.
but i found out, that options/feature of commands of busybox on OpenELEC are not available.

on OpenELEC the busybox version is
```BusyBox v1.22.1 (2015-02-27 20:34:02 CET) multi-call binary.```
i have the same version on an other device installed
```BusyBox v1.22.1 (2014-01-20 ...) multi-call binary```
there the examples from the internet are working on that device.
for example, because of spaces in the filename, i have to use the ```-print0``` option for command ```find```, with out that option, the following command ```xargs -0``` will struggle.
on OpenELEC, the command ```find``` does not understand that ```-print0``` option.
so i called ```--help``` of the command ```find``` and i saw, that the number of supported options are different - even the versionnumbers of thier busybox are equal.

on freenode i got information that the busybox version in OpenELEC i a stripped down version of busybox by choice. that is the reason, why commands and its options won't work on OpenELEC.

on the freenode channel i got suggestions to use command in an other way to reach the same result, but even then there was no success.

as last workaround i downloaded the official binary of busybox from
http://www.busybox.net/downloads/binaries/1.21.1/ (unfortunately there is no 1.22.1 binary for download available yet)
(i used **busybox-armv7l** on RPi2) and used that version as substitution for the old feature stripped busybox version of OpenELEC.
it was working more or less easy.

i can't beleave, that busybox was stripped down in features, because of stability of OpenELEC.
i think the reason is to shrink down the size of busybox for systems with less RAM.
isn't it possible to include a full featured version of busybox to official OpenELEC for systems, that have enough RAM?
on ma RPi2, the busybox binary version of OpenELEC is 0.5MB, the full featured official busybox binary version i donloaded is 1.1MB.
i see no reason to stripp down busybox only to have 0.6MB more RAM available on systems that have enough RAM.
ok, i think all the most features of busybox will very rarly be used, but if you need a feature, and you see no other way how to realize, then you really whish to have all the features.

for tests, i tried things like:
```find /media/ -type f -print0 | xargs -0 echo``` (should work without errormessage)
```find /media/ -type f -exec echo {} \;``` (should work without errormessage)
```find /media/ -type f -printf "%f\n"```
```find /media/ -type f -print "%f\n"```

maybe i gave up too early to find a proper one-liner for the stripped down busybox version of OpenELEC

EDIT: tested against busybox_vm on http://www.busybox.net/live_bbox/live_bbox.html
```find . -type f -print0 | xargs -0 echo```
```find . -type f -exec echo {} \;```
should work (simply list filenames that find is giving).

stefansaraev:
closes #3999

master only

will push in 24 hours if no objections.


fritsch:
+1 though we are a very "cut down distribution" - as we have a shell, we
should not limit that by removing useful tools.

2015-03-12 21:51 GMT+01:00 Stefan Saraev <notifications@github.com>:

> closes #3999 <https://github.com/OpenELEC/OpenELEC.tv/issues/3999>
>
> master only
>
> will push in 24 hours if no objections.
> ------------------------------
> You can view, comment on, or merge this pull request online at:
>
>   https://github.com/OpenELEC/OpenELEC.tv/pull/4000
> Commit Summary
>
>    - busybox: enable all features for 'find' applet
>
> File Changes
>
>    - *M* packages/sysutils/busybox/config/busybox-target.conf
>    <https://github.com/OpenELEC/OpenELEC.tv/pull/4000/files#diff-0> (22)
>
> Patch Links:
>
>    - https://github.com/OpenELEC/OpenELEC.tv/pull/4000.patch
>    - https://github.com/OpenELEC/OpenELEC.tv/pull/4000.diff
>
> --
> <https://github.com/OpenELEC/OpenELEC.tv/pull/4000>.
>



-- 
                   Key-ID:     0x1A995A9B
                   keyserver: pgp.mit.edu

PR #4002, opened 2015-03-13
kodi: update to kodi-14-e7ba06f

wezley98:
Update OpenElec 5.0.x to use Kodi 14.2 RC

### DIFF
--- a/packages/mediacenter/kodi-theme-Confluence/package.mk
+++ b/packages/mediacenter/kodi-theme-Confluence/package.mk
@@ -17,7 +17,7 @@
 ################################################################################
 
 PKG_NAME="kodi-theme-Confluence"
-PKG_VERSION="14-085163e"
+PKG_VERSION="14-e7ba06f"
 PKG_REV="1"
 PKG_ARCH="any"
 PKG_LICENSE="GPL"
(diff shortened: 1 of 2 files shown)


fritsch:
Thx - do all our custom patches still apply? Anything that can be dropped?

stefansaraev:
wait for @sraue I am sure he already has it packaged and pushed to his local tree.

wezley98:
Just running the build process now, am I correct in assuming that the build process would notify me on completion if anything failed to apply?

Note: Quite new to the project but do have spare time if there is anything that needs to be worked on, Mainly a php dev for coding skills, but I want to contribute something back to the project as I use it every day.

fritsch:
Spare time is always good to have. Make yourself familiar with our buildroot. That way you will be able to provide (un)official addons. Furthermore there are a lot of users with very small questions on IRC - also welcome here to help.

wezley98:
Will do.

Looks like this build fails to apply @ OpenELEC.tv/projects/RPi2/patches/kodi/kodi-001-helix_rpb_backports.patch

http://sprunge.us/jYeY

stefansaraev:
added

PR #4009, opened 2015-03-14
wetek_play: linux: fix patch order

stefansaraev:
for now just cosmetics, but wrong patch order may (and will) cause issues in future.

@codesnake do you mind renaming those ?


sraue:
+1 no problem for me

codesnake:
@stefansaraev The order should not change, because I just removed some of the patches, but didn't rename them. But you are right they require a cleanup.

stefansaraev:
current order
000 - 10 - 100 - 120 - 20 - 30 ... 70
(10,20,xx applied before 100, 1xx)

should be: 
000 - 010 - 020 - .... - 100 - 120 .. 

:)

sraue:
also would be nice to use 9xx instead or so, maybe we need some other patches from our ald 3.10 kernel (used in our 3.2.x builds) i think about some remote/dvb drivers which will be maybe requested in the future (see https://github.com/OpenELEC/OpenELEC.tv/tree/openelec-3.2/packages/linux/patches/3.10.24 )

PR #4020, opened 2015-03-19
Libaacs cleanup

mrdominuzq:
compiles fine under subdir and remove  "unrecognized options: --disable-examples, --disable-debug"

also i dont think the patch is needed.


packages/multimedia/libaacs/package.mk
```diff
@@ -30,22 +30,14 @@ PKG_SHORTDESC="libaacs: a research project to implement the Advanced Access Cont
 PKG_LONGDESC="libaacs is a research project to implement the Advanced Access Content System specification. This research project provides, through an open-source library, a way to understand how the AACS works. This research project is mainly developed by an international team of developers from Doom9."
 
 PKG_IS_ADDON="no"
-PKG_AUTORECONF="yes"
+PKG_AUTORECONF="no"
```
stefansaraev: humh. why. we always do autoreconf when possible.

mrdominuzq:
isn't only needed if configure.ac has been modified ?

stefansaraev:
generaly yes. but in the past, some sources come with fucked up generated configure. 

ping @sraue

sraue:
we also replace config.guess and config.sub generally which needs a autoconfig... not sure if its still needed this way, maybe for some very old packages or if new architectures are added...

whats the problem with enabled autoreconf for this package?

mrdominuzq:
i haven't tried with autoreconf enabled, because when configure.ac wasn't modified without the patch, i thought autoreconf wasn't needed :)

sraue:
@mrdominuzq can you try and enable if it works? if not, paste the error here

mrdominuzq:
Yes :)

mrdominuzq:
https://www.diffchecker.com/vlviw1am

output with/without autoreconf

sraue:
looks good and dont breaks, so enable autoconf again and update this PR, then we will commit

mrdominuzq:
Okay :) Will do it sunday or monday. But just to nr clear, We Walt to enable autoreconf on target packages ?

sraue:
yes we do this, except there is no autoconf support (cmake, no configure....), or its broken (very rarly) for all packages, also host packages

mrdominuzq:
will keep that in mind :) thanks

mrdominuzq:
squashed :)

PR #4024, opened 2015-03-20
VAAPI: Bump libva and libva-driver-intel

fritsch:
libva: Our cherry-picked patch can be removed has gone upstream
libva-driver-intel: a lot of bug fixes upstream - though we still need to keep the deint fixes for BYT, IVB, SNB - was too late for them to get into 1.5.x


stefansaraev:
push?

PR #4030, opened 2015-03-21
linux: eMMC: Don't initialize partitions on RPMB flagged areas.

stefansaraev:
this fixes an issue on some baytrail boxes with integrated emmc, where installer hangs / wont detect the target drive.

on that hardware parted and blkid would hang for minute or two, trying to access /dev/rpmb
tested by me and @fritsch on an "wintel" box.

would be nice if RPi dev can check if this doesnt break something.

//cc @MilhouseVH 

EDIT: @MilhouseVH I dont know if RPi/2 has/needs /dev/rpmb. but if it boots with this patch, all fine.


MilhouseVH:
@seo: Tested on both RPi and RPi2, no problems booting with this PR (they don't have /dev/rpmb).

stefansaraev:
thanks @MilhouseVH

PR #4033, opened 2015-03-22
[xpad] update steamos patch with urb reuse fix

zzattack:
Another attempt at fixing xpad for wireless controllers, this time including the patch from https://github.com/ValveSoftware/steamos_kernel/pull/14 

Just tested it, still panics.
Maybe also needs the patch from http://www.spinics.net/lists/linux-input/msg31446.html


zzattack:
Could you revert this for the time being?

PR #4034, opened 2015-03-22
scripts: add support for PKG_ADDON_PROVIDES

sraue:


stefansaraev:
for some reason. this did not work for me (isengard), but it doesnt hurt. if you want it - push it.

addon devs should be pinged. @lsellens

PR #4042, opened 2015-03-26
Intel: Fix EDID overriding (makes audio work)

fritsch:
This should fix a long standing issue when users want to boot with their AVRs disabled or resume from suspend and end without audio. 

It's perhaps worth to integrate a working EDID file into OE for easy usage. Was only verified with kms_helper for now.

Edid (Full HD + All bitstream audio codecs): https://dl.dropboxusercontent.com/u/55728161/edid-samsung-marantz.bin


bkuhls:
Works great when AVR+TV (HDMI: PC -> AVR -> TV) are powered off during boot using i3-4330T CPU with a custom edid extracted from my system, all AVR capabilities are available in Kodi!

Now I only have to revert a CEC-related commit so I can power on my TV hours after boot and still have a working CEC-connected remote control, now I have 30s ;)
https://github.com/opdenkamp/xbmc/commit/85d5bd459e157bb4b2650f4bf0a1dda1b962a545

fritsch:
@bkuhls Thanks for your feedback!

Big thank you goes to @jnikula for implementing it. If you have something on your amazon wish list - feel free to post a link :-)

fritsch:
If it applies, it can go to OE 5.0.x - as it is used by a minority and was also CCed to stable-kernel.

PR #4051, opened 2015-03-28
Update package.mk

hgomez:
libssh: change download url, red server is currently offline, the 0.6.4 source code can be downloaded from their git server.


stefansaraev:
ty. could you please change commit message to something more clear. like "libssh: change download url"

stefansaraev:
however. it works now.

PR #4052, opened 2015-03-29
[build] add BUILD_SUFFIX to the IMAGE_NAME, too.

InuSasha:
actually, different builds (with different changes) on the same commit are overwriting the previous files in target/.
(spezially on non-devel builds)

testet on master and 5.0.6 with release and noobs builds

Sascha


stefansaraev:
ok from me. doesnt hurt. @sraue ?

MilhouseVH:
Would it be clearer/more flexible if this were called TARGET_SUFFIX, although I appreciate you may always want to set this var as well as BUILD_SUFFIX, however there may also be times when you only want to set one but not the other, or even use different values for TARGET and BUILD suffixes.

InuSasha:
This is a solution, too.
A third way, is to add an OPENELEC_VERSION_SUFFIX.
-> replaces (actual) BUILD_SUFFIX and IMAGE_SUFFIX, additional it is also in version number in the running system.

If you want the OPENELEC_VERSION_SUFFIC or IMAGE_SUFFIX, tell me. I make some tests and PR.

Sascha

stefansaraev:
IMAGE_SUFFIX looks better. but I am fine with whatever makes your life easier ;)

InuSasha:
I switched to IMAGE_SUFFIX. It was tested with master and 5.0.8 in combination with release and noobs target.
The OPENELEC_VERSION_SUFFIX is implemented at an other branch (https://github.com/InuSasha/OpenELEC.tv/tree/features/build_openelec-version), if you are interested.

stefansaraev:
IMAGE_SUFFIX is good. and imo enough. thanks.

PR #4065, opened 2015-04-01
Enable use of systemd-journal-gatewayd

lrusak:
I'm not sure of how practical this will be or if it will be more of just a "cool" thing

systemd-journal-gatewayd allows viewing the journal from a remote host using http. ie,
```
http://<openelec-ip>:19531
```

This allows you to browse the entire journal or journal entries from specific units.

You can also format the url to get specific information
```
http://<openelec-ip>:19531/entries?boot&PRIORITY=3
```
This will get journal entries for the current boot that are log priority 3

This uses socket activation via systemd-journal-gatewayd.socket so systemd-journal-gatewayd does not run until the specific port is requested.

see, http://www.freedesktop.org/software/systemd/man/systemd-journal-gatewayd.html
and, http://www.freedesktop.org/software/systemd/man/systemd.journal-fields.html

### DIFF
--- a/packages/sysutils/systemd/package.mk
+++ b/packages/sysutils/systemd/package.mk
@@ -23,7 +23,7 @@ PKG_ARCH="any"
 PKG_LICENSE="GPL"
 PKG_SITE="http://www.freedesktop.org/wiki/Software/systemd"
 PKG_URL="http://www.freedesktop.org/software/systemd/$PKG_NAME-$PKG_VERSION.tar.xz"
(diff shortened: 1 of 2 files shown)


stefansaraev:
:-1: 
no!

MilhouseVH:
-1

It's somewhat cool, but crazy to include as standard as it's just another potential security nightmare waiting to happen IMHO. If you want remote log access, set up remote ssh access, or use netcat etc.

If it could be made into an unofficial addon then sure, I'd give it a +1, although I couldn't see myself ever using it.

stefansaraev:
I just close this ;)

PR #4069, opened 2015-04-05
Wayland display server support

lrusak:
This is just a place holder to show what is currently possible and not possible when using a pure wayland display server on OpenELEC. I have no interest testing xwayland as there is no benefit to OpenELEC.

**Working:**
- [x] Transparent startup with Kodi auto starting.
- [x] Intel graphics
- [x] AMD graphics
- [x] Nvidia graphics (using nouveau, the nvidia blob doesn't support wayland at this time)

**Not Working:**
- [ ] VAAPI (VAAPI is supported on wayland, but isn't yet supported through Kodi)
- [ ] VDPAU (no wayland support)
- [ ] Refresh rate switching (no wayland equivalent to xrandr yet)

**Other Details:**
- Weston is being used here, though using a different compositor would be ideal there isn't one that currently fits the needs of OpenELEC (ie, orbment, velox, etc)
- Only for x86_64 build
- Need the following PR for vaapi on wayland. xbmc/xbmc#1302

I will continue updating this PR as support improves :+1: 

**Updates:**

2015-05-08: Update to wayland/weston 1.7.91
https://dl.dropboxusercontent.com/u/21742150/wayland/OpenELEC-Generic.x86_64-20150508-wayland.tar

2015-06-05: Update to wayland/weston 1.8.0
https://dl.dropboxusercontent.com/u/21742150/wayland/OpenELEC-Generic.x86_64-20150605-wayland.tar


stefansaraev:
nice nice! :)

tapir:
The ideal would be that Kodi becomes a wayland compositor itelf. Do you know if anything like that is in the works?

lrusak:
@tapir I don't think anyone is working on such a thing (yet). I think once libweston is split from the weston tree it will be much easier for people to develop their own compositors.

see, http://www.phoronix.com/scan.php?page=news_item&px=wayland-libweston-1.9

PR #4071, opened 2015-04-05
rework live mode and clean up

stefansaraev:
as title says

@vpeter4 sorry, this has to go.


vpeter4:
Progress require sacrifice too :)

PR #4073, opened 2015-04-05
remove nbd support

stefansaraev:
nbd is documented nowhere. lets see if someone still use it

it is probably broken, anyway, at least I am sure suspend/resume is broken like with iscsi..


stefansaraev:
ping @opdenkamp do you still need this ?

opdenkamp:
hey, I'm not using it right now but I'm using nfs instead for the moment. it does feel slower to boot though, probably because of the extra overhead, but I haven't measured boot times.

any reason for removing this, because I wanted to switch back to using nbd anyway, just needed to update some scripts after reinstalling servers.

stefansaraev:
hehe. removing nbd is the first (second, CIFS is gone) step to completely remove netboot.

nah. I am kidding :)

I thought it's not used at all, as it's not documented on OE wiki, and I have never seen anyone using it/asking questions about it. 

however. it does no harm. and as you (as the original submitter of the work) want to switch back to nbd. I am not touching it.

note. suspend/resume is broken with nbd/iscsi, I am not going to ever think about fixing it ;)

opdenkamp:
thanks, I'm not using suspend/resume. probably doesn't work with nfs either ;-)

PR #4075, opened 2015-04-06
New Project - Nvidia-Legacy

lrusak:
Now that i386 is dropped we need to update the nvidia driver to support new cards, but we also need to support ION machines (for now).

This is basically a skeleton of the Generic project with changes to the graphics drivers and the add-ons url (to use the same url as the Generic project)

With this then #3969 can be pulled and everyone can be happy :+1:


stefansaraev:
@sraue what you prefer. this or https://github.com/OpenELEC/OpenELEC.tv/pull/3969#issuecomment-82737784

stefansaraev:
no bikesheding. but as this is nvidia specific, what about renaming it to Nvidia-Legacy or so ? @sraue so users are not confused ;)

chewitt:
I agree to that.. nvidia-legacy removes confusion. rebase!

lrusak:
Updated

PR #4076, opened 2015-04-06
add 2sf audio decoder addon

notspiff:
This adds a 2sf audio decoder addon.

I consider this a training run, I got another 10 or so on the way, but want to see if this is okay.
I have built it for x86_64 (Generic) and arm (RPi2).

### DIFF
--- a/packages/mediacenter/kodi-binary-addons/audiodecoder.2sf/package.mk
+++ b/packages/mediacenter/kodi-binary-addons/audiodecoder.2sf/package.mk
@@ -1,0 +1,48 @@
+################################################################################
+#      This file is part of OpenELEC - http://www.openelec.tv
+#      Copyright (C) 2009-2014 Stephan Raue (stephan@openelec.tv)
+#
+#  OpenELEC is free software: you can redistribute it and/or modify
+#  it under the terms of the GNU General Public License as published by
+#  the Free Software Foundation, either version 2 of the License, or
+#  (at your option) any later version.
+#
+#  OpenELEC is distributed in the hope that it will be useful,
+#  but WITHOUT ANY WARRANTY; without even the implied warranty of
+#  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+#  GNU General Public License for more details.
+#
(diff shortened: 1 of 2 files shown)


stefansaraev:
thanks. I'll do quick build test later today.

is this going upstream? atm we use mkpkg_kodi to package everything ;)

notspiff:
upstream doesnt want to carry it.

packages/mediacenter/kodi-binary-addons/audiodecoder.2sf/package.mk
```diff
@@ -0,0 +1,48 @@
...
+################################################################################
+
+PKG_NAME="audiodecoder.2sf"
+PKG_VERSION="014b430"
+PKG_REV="1"
+PKG_ARCH="any"
+PKG_LICENSE="GPL"
+PKG_SITE="https://github.com/notspiff/audiodecoder.2sf"
+PKG_URL="$DISTRO_SRC/$PKG_NAME-$PKG_VERSION.tar.xz"
+PKG_DEPENDS_TARGET="toolchain"
```
stefansaraev: add kodi-platform ?

notspiff: not needed. it's only needed on win32 for (most) audio decoders.

stefansaraev: then kodi (zlib would be pulled as dependency while building kodi). all addons should be build-able by scripts/create_addon. without doing "make release" first :)

notspiff: right. thought addon env was part of toolchain. then we should do kodi-platform in case it is ported to win32 at some point..

stefansaraev:
built fine on all supported projects. you may pick https://github.com/stefansaraev/OpenELEC.tv/commit/d70f078439e25bf6a078c63fa0bc0c781ce0bfa9

@sraue objections?

notspiff:
added dependency, picked commit.

PR #4078, opened 2015-04-07
Replace b43 with bcm_sta

lrusak:
This is a rebase of #3031 as @chewitt is busy :)

I do not have a Broadcom chipset so I cannot test. It would be nice if someone could test before this is merged.

I have included the patch source url in the linux 4.0 patch.

~~The bcm_sta package needs to be updated in $DISTRO_SRC as we no longer need the 32bit version of the driver. This will also require some reworking of the patches. I can do this but I don't have FTP access.~~

EDIT: This is the actual source. https://github.com/longsleep/bcmwl-ubuntu/tree/master/debian/patches
I would like to use the patches here, ~~but will wait till the package is update in $DISTRO_SRC so I don't have to edit them~~

EDIT 2: Should probably wait to remove the 32bit support from the package as people may still be using it. So this is good to go for now.


sraue:
would like it if you can add the seperate patches from the patch source, not as a single big one - then i will merge, cant test, but we will so who complains

lrusak:
Update with upstream patches. Note that not all patches were needed.

It would be nice to repackage the bcm_sta package in $DISTRO_SRC to only use the x86_64 sources as this package is listed as PKG_ARCH="x86_64" anyway. This would make it so I don't have to edit the patches to comply with our packaging.

chewitt:
The current bcm_sta driver package we host is used by a few people self-building 32-bit OE 5.0.x custom builds so we either change the package name used in master to bcm_wl and make sources for the new package 64-bit only, or we continue as-is for a while longer and drop 32-bit support when the next wl version comes along. It's not just the patches that need 32/64 tweaking; historically I have to take the separate vendor provided 32/64 sources and combine them.

lrusak:
Then I would rather just package the x86_64 version into a new package (bmc_wl if you want) and use that here going forward. That way we have un edited patches :+1:

sraue:
please tell me if its done, we should provide the driver for i386 too (maybe for OE-5.2)

lrusak:
@sraue not sure if it is a good idea to put this in 5.2, it has the possibility to have some breakage and could use testing. I recommend this only goes into master.

chewitt:
I'm not so worried about breakage; we switched from b43 on AppleTV about two years ago without any drama and there's not been wl related wireless complaints there or from any of the users who have built 5.0.x with it to get support more recently.

stefansaraev:
just FYI, when this breaks (and blocks me from upgrading kernel), I'll remove it immediatey, then it's your job to fix or re-enable in-kernel drivers ;)

lrusak:
@chewitt alright well I still don't think it should go in 5.2

@stefansaraev understood. :+1:

lrusak:
@sraue let me know what you want. If you want this for 5.2 then I will have to update the patches to support i386 builds.

sraue:
@lrusak can you rebase for master

lrusak:
rebased

sraue:
@lrusak: sorry could you add the change for the just merged nvidia-legacy project too, so they are in sync? thanks much

lrusak:
rebased with latest PR's

PR #4082, opened 2015-04-07
libva: doesn't fail with subdir

mrdominuzq:
compiles just fine :)


stefansaraev:
nope. it may fail in future. not building in subdir doesnt hurt, as we dont build libva:host, anyway. my 2 cents.

mrdominuzq:
that's correct :) but what are the odds for that ? 

all packages could fail with subdir when bumping :)

stefansaraev:
yea. that,s good point too. thanks

PR #4085, opened 2015-04-08
Fix compile flags

vpeter4:
Issues detected with new project I'm working on which is using some different compiler flags.
Tested on all current projects.

### DIFF
--- a/packages/compress/unzip/package.mk
+++ b/packages/compress/unzip/package.mk
@@ -38,8 +38,7 @@ make_target() {
       RANLIB=$TARGET_RANLIB \
       AR=$TARGET_AR \
       STRIP=$TARGET_STRIP \
-      CFLAGS="$CFLAGS" \
-      -f unix/Makefile generic
+      -f unix/Makefile generic LOCAL_UNZIP="$CFLAGS"
 }
 
 makeinstall_target() {
--- a/packages/sysutils/keyutils/patches/keyutils-02-cflags.patch
+++ b/packages/sysutils/keyutils/patches/keyutils-02-cflags.patch
@@ -1,0 +1,23 @@
+From d8f15e55b4d357ac0a2b384888080b120600d09d Mon Sep 17 00:00:00 2001
+From: vpeter4 <peter.vicman@gmail.com>
+Date: Wed, 8 Apr 2015 18:27:06 +0200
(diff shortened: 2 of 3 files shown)


stefansaraev:
ok for others.

stefansaraev:
however, the plymouth change doesnt hurt.

@sraue any objections ?

stefansaraev:
humh. the plymouth commit completely breaks splash loading for me, on wetek play. I'll revert.

MilhouseVH:
Also on Pi, currently in the middle of confirming this but you may have beaten me to it. :)

stefansaraev:
revert fixes it for me. would be nice if you can confirm @MilhouseVH - revert, clean linux busybox and plymouth-lite (just in case, to be 100% sure) then rebuild. thanks ;)

MilhouseVH:
Did you revert the whole PR or just the plymouth-lite commit? I tried reverting only the latter but still had a broken splash, then reverted the whole PR and splash is working. I'll run some more builds and confirm shortly.

stefansaraev:
> Did you revert the whole PR or just the plymouth-lite commit?

only plymouth-lite. scripts/clean linux; scripts/clean plymouth-lite is a must.

MilhouseVH:
Yep, did that but maybe something else went tits up - will confirm in a bit.

MilhouseVH:
Yep confirmed - revert bdde001075b23808cee9e12c7e8f7c31500108ec and splash works again. Many thanks.

stefansaraev:
thanks. lets revert for now then

vpeter4:
I checked on Generic build. But could be possible that something went wrong with my test and didn't actually tested this change.
But ply-image is not real static binary - it still needs libc.so. With my change it needs libm.so too. That's why it fails.
Revert and I will reinvestigate.

stefansaraev:
> But ply-image is not real static binary - it still needs libc.so

yes. it was never intendded to be a "real" static binary. but we dont want libm.so in initrd

extending -22-link-static.patch with "-Wl,-Bstatic -lm -Wl,-Bdynamic" will probably help on your use case, that's why I initialy pointed out that patch ;)

vpeter4:
So it is semi-static then :-)
Anyway, will check again.

PR #4095, opened 2015-04-13
wetekplay "fixes"

stefansaraev:
@Memphiz saradc is meh. wrong values and nobody working on it. should go

@codesnake @sraue 4916d8c ended up in master by accident ? it introduces a regression - massive livetv stuttering. this revert was in my builds for long


codesnake:
@stefansaraev I agree that temperature reported is wrong. But if you drop support for saradc, you can drop also the following patch: https://github.com/OpenELEC/OpenELEC.tv/blob/master/projects/WeTek_Play/patches/kodi/0004-Add-support-for-retrieval-of-CPU-temperature-on-Amlo.patch

Memphiz:
Someone from wetek said they wanted to ask about the right adc values at amlogic. But thats long ago. Agree - if nobody fixes it it could go - i don't care about temperatures at all ;)

stefansaraev:
@codesnake thanks. I missed that patch.

stefansaraev:
@codesnake I am going to merge this tonight. any comment on wetekdvb from your side? iirc, this particular version caused some issues on nand installs, but we dont support nand installs in master (it's separate "wetek" branch). do you agree with reverting it ?

codesnake:
@stefansaraev Yes I agree to revert the module. I tested it and it works good.

Issue #4099, opened 2015-04-18
Cannot set static IP without dynamic IP connection


miigotu:
I have DHCP disabled in my router, so initial connection from OE on rPi cannot succeed.

The only workaround is to first enable DHCP in my router, connect to the AP with OE settings addon, edit the connection with my static IP, disable DHCP in my router settings, and reconnect network in OE.

I should be able to specify IP, Gateway, and Netmask in order to connect for a new connection, not just edit an existing connection.

Also, I noticed this when swapping out my router for a new one which I preconfigured with the same setup as the old one. Of my 12 devices connecting over wireless, OE was the only device that failed to automatically connect to the new router (Same SSID and PW, different mac).

chewitt:
I do not support the idea of creating static connections from within the initial setup wizard; too many n00b users will take up the idea then complain about things not working that result from their own configuration mistakes. I do support the idea that you should be able to create a connection from scratch within the settings addon instead of editing a current one. Perhaps when you highlight/click on Ethernet or the SSID it should ask you whether to use DHCP (default) or configure static details? NB: Although I support this I am not the person to be fiddling with python code in the addon (if you want it to work, etc.).

stefansaraev:
duplicate of #4081
closing.

stefansaraev:
@chewitt

> I do not support the idea of creating static connections from within the initial setup wizard

to make it clear: this is already supported in wizard. but is broken in recent 5.0.x (connman crashing). now that's fixed in "next" stable.

miigotu:
@stefansaraev this is a different issue than #4081 
I cant even get to a menu to set the static IP because the connection hasn't been made yet. Currently you can only set a static IP if the connection is already existing and saved in connman. If DHCP is disabled in the router one cannot connect and save the connection (in order to go back and edit it with static ip) because it fails on the configuring stage and never gets saved as a valid connection.

stefansaraev:
> I cant even get to a menu to set the static IP because the connection hasn't been made yet

then you are talking about wifi ? because "edit" is possible for wired connections.

> If DHCP is disabled in the router one cannot connect and save the connection

we dont consider supporting your wifi edge-case.

miigotu:
Also, I just was doing some tests for an issue I was having and decided to use `create_sdcard` with Milhouse build `#418`

[...]

PR #4101, opened 2015-04-19
connman: revert debb37a

stefansaraev:
debb37a broke wifi in wizard / first run. as it doesnt seem to
be intentional, reverting.

ref https://github.com/OpenELEC/OpenELEC.tv/issues/4099#issuecomment-94216561 
ref http://openelec.tv/forum/92-openelec-os-settings/75617-turn-wifi-on-during-wizard#138096

this should NOT go to stables, the damage is already done. but I dont want to see the "wizard hangs" reports again, it's too late for 5.x

EDIT: personaly, I dont care for wifi, but we had this in 4.x, and it's supported in our settings addon / wizard, anyway. if we dont auto-enable wifi, the "network" wizard page does not make sense for people who dont have (or don want to use) wired.


stefansaraev:
@sraue

sraue:
@stefansaraev i am ok if it fixes this issue

stefansaraev:
it does. tested, but I dont want it in stables, really. there is some shit either in connman, or pydbus, or our settings addon idk, causing "deadlocks" in the past, it could be better after downgrading connman but better be safe than sorry, i we get reports after 6.0 betas, I'll see what I can do.

PR #4103, opened 2015-04-21
kodi: disable pvrplayback.signalquality on new installs

stefansaraev:
this was reported to cause stuttering/artifacts

note. this only disables tte pvr-specific "signal quality" part of codecinfo screen (g). user can re-enable it if he wants, we just set it default off

@sraue objections ?


stefansaraev:
this is also safe for our stable branch

PR #4109, opened 2015-04-26
kodi/udevprovider: filter out OE specific mounts

stefansaraev:
this hides /flash and /storage from videos/music -> files -> add -> browse.

@ksooo FYI


ksooo:
Thx. I'm fine with this.

PR #4116, opened 2015-04-30
Include libvcsm.so in RPi/RPi2 builds - needed by upcoming firmware

MilhouseVH:
Safe to include now, before next bcm2835-* bump.


popcornmix:
Don't forget this if you bump firmware on 5.0 branch.
(there are some useful firmware fixes not in 5.0 tree like reduced jaggies on deinterlace, and fix for hdmi dropouts with 1080i deinterlace).

stefansaraev:
yup. I dont know if @sraue will update it. cherry picked this to be safe. thanks.

PR #4117, opened 2015-04-30
Python: remove distutilcross not needed

mrdominuzq:
i havent tested the other python packages, because they are not included and maybe should be remove.

and i couldn't find anything about pycrypto, so maybe somebody could tell me what depends on it :)


stefansaraev:
I dont think this is good idea.

mrdominuzq:
it shouldnt cause any harm :)

stefansaraev:
are you sure, even compiling for arm, on x86_64 host ? how does setuptools:host handle that ?

mrdominuzq:
the only 2 who failed was pillow and pycrypto, that's why i added the cppflags. but i dont have the chance to test it since my rpi is not a home right now :)

but here's the test build
https://drive.google.com/file/d/0B-2ANc5gpRgBQWp6ZGZDbEJpdGM/view?usp=sharing

stefansaraev:
it's not only cppflags. it's a new world of mess. I'd keep distutilcross as is.

mrdominuzq:
after having a look at it again, found some missing file with setuptools.

it's missing libtiff and freetype. https://www.diffchecker.com/ejve2ddw
so your right, stay with distutilcross :)

sorry for the waste of time ;)

stefansaraev:
np. it's not a waste of time.

mrdominuzq:
it's a good reminder to absolutely certain  when making changes :)

PR #4120, opened 2015-05-02
Update kernel and DVB module for WeTek Play

codesnake:
* Update kernel to 3.10.75 (thanks @stefansaraev)
* Update WeTek DVB module


stefansaraev:
thanks @codesnake I've been using this for quite long with no visible regressions
will do one last build test on top of updated master

PR #4123, opened 2015-05-04
projects/WeTek_Play/patches/linux: Remove all NAND partitions in device tree

codesnake:
Having pre-defined NAND partition layout in device tree causes NAND driver to
update partition layout in NAND environment. If we are booting OE from SD-card
but have Android installed in NAND, this will overwrite Android NAND partition layout
with OE partition layout, and bootloader will no be able to boot Android anymore.

This update removes NAND partition layout from OE device tree. This causes NAND
driver to always use partition layout from NAND environment never overwriting it.

For OE NAND installations partition layout will be written to NAND environment
by recovery during installation, and then just reused by OpenELEC.


stefansaraev:
thanks much. so after this we can re-add 06e5bdcf7890d0b7de709226a80403619eb4069d (assuming an android OTA with updated bootloader comes soon) ?

I'll re-test all tonight.

codesnake:
Yes, I tested both with OE in NAND and Android + OE from SD-card. It seems works without problems. And also allows to access the Android partitions from OE.

stefansaraev:
humh, any chance you can rebase if I update wetek branch ?

codesnake:
ok

stefansaraev:
thanks. rebased and cherry picked

PR #4125, opened 2015-05-04
WeTek_Play: Add support for installing OE into NAND memory

stefansaraev:
one more attempt to push this to master. hopefuly the last one. 

after this is merged, for WeTek.Play to boot a sdcard build, it will need an updated bootloader (boot android once, do ota update).

I was told an android OTA update will be available later today. but as of time of writing this, it is still not available. 

@codesnake @sraue whats the plan, should we wait for the OTA or merge now. I tested this with the OTA zip you linked in irc yesterday and it's all good.


stefansaraev:
OTA on wetek servers now. all good.

PR #4128, opened 2015-05-05
projects/Nvidia-Legacy: strip unneded GPU drivers from config and kernel...

sraue:
... config

cc @stefansaraev @lrusak i see no reason it should not work, could maybe test later, but would be nice if you could test too


stefansaraev:
+1 if mesa works/can be built.

MilhouseVH:
I've tested this PR with new Nvidia-Legacy project on a Revo 3700/ION2 - working fine.

stefansaraev:
mesa is triggered by everything that depends on $OPENGL - kodi, glew, xorg-server etc. it's built successfuly here but not yet runtime tested (full build not yet finished). will test on my ION and report later.

PR #4130, opened 2015-05-05
linux: add patch to support Hauppauge HVR-955Q ATSC USB receiver

trsqr:
This patch adds support for the Hauppauge HVR-955Q ATSC USB receiver, which is a newer version of the HVR-950Q and one of the few ATSC receivers currently being sold on the market and having Linux support.

The patch looks rather large, but bulk of the patch is the new demodulator driver for the LGDT3306a demodulator.

Obsolete after kernel 4.1 is taken into use.


stefansaraev:
Nvidia-Legacy needs kernel config sync too (same as Generic), no big deal as CONFIG_DVB_LGDT3306A is autoselected anyway

PR #4132, opened 2015-05-07
bump audiodecoder add-ons

notspiff:
due to split of kodiplatform.

don't merge before you have added the platform lib.


stefansaraev:
thanks. I'll add libplatform when I bump libcec (it's in my git but still needs minor work)

PR #4134, opened 2015-05-08
libva-intel-driver: pick field_rate workaround for mcdi and madi on Gen6

fritsch:
Let's test this in OE 6.0 cycle. It was submitted upstream.

Fixes MCDI /MADI not running in field rate, cures slightly jumping content when deinterlacing


stefansaraev:
oh, this should be against master branch, for now everything that goes to master, goes to 6.0 too. and we avoid breaking changes.

fritsch:
will reopen.

PR #4144, opened 2015-05-12
bump audiodecoder.qsf

notspiff:
sorry for the quick bump but i found a bug that rendered the add-on unusable..


stefansaraev:
FYI, if version is not bumped also in addon.xml, nothing will be re-built and uploaded to repos ;)

notspiff:
meh. i guess i have to start paying attention to that from now on. haven't bothered since there was no downstreams

notspiff:
done.

PR #4145, opened 2015-05-14
IMX: Backport latest fixes

fritsch:
- Fixup for EDID
- Fixup for Artifacts
- Enumerate D modes for DVI monitors
- Fix RenderCapture

All those patches (besides the D modes) can be dropped when we reach v15


fritsch:
Tested on my i4. Working nicely.

PR #4147, opened 2015-05-14
u-boot: support for multiple targets, patches for udoo system

vpeter4:


packages/tools/u-boot/scripts/update.sh
```diff
@@ -32,6 +32,13 @@ if [ -z "$BOOT_DISK" ]; then
   esac
 fi
 
+# we don't have mounted /etc in initramfs so just load system type
+if [ -f $SYSTEM_ROOT/etc/profile.d/01-system_type.conf ]; then
+  . $SYSTEM_ROOT/etc/profile.d/01-system_type.conf
```
lrusak: Is this part of another PR? I see here that it is only applicable to the TBS-matrix board for now.

vpeter4: Here only for matrix yes. Because of bootloader. It could be used directly by reading /sys/bus/... but using conf file it's more generic for later.

PR #4148, opened 2015-05-14
buildsys: add $OS_VERSION to builddir for dev builds

stefansaraev:
switching branches and re-building without make clean, where target branch has different OS_VERSION is a call for a disaster

@fritsch FYI :)


stefansaraev:
added: OS_VERSION is added to target image name.

PR #4149, opened 2015-05-14
IMX: Power off LED and display after poweroff

fritsch:
This turns off the LED and the display after all other shutdown stuff is done.


stefansaraev:
this has my +1

PR #4159, opened 2015-05-22
Udoo poweroff

vpeter4:


stefansaraev:
ok.

//cc @fritsch vpu352 looks sane to me.

fritsch:
The info is wrong there it always prints. Should only when if has not
failed :-)
Am 22.05.2015 2:26 nachm. schrieb "Stefan Saraev" <notifications@github.com

PR #4161, opened 2015-05-23
add serial console support

haggaie:
Make debugging a raspberry pi simpler by adding agetty and a serial console service.

### DIFF
--- a/config/options
+++ b/config/options
@@ -79,9 +79,10 @@ LINUX_DEPENDS="$PROJECT_DIR/$PROJECT/linux/linux.$TARGET_ARCH.conf $ROOT/package
 # forcing install of all packages (yes/no)
   FORCE_INSTALL="no"
 
-# install devtools on development builds
+# install devtools, serial console on development builds
   if [ "$OPENELEC_VERSION" = "devel" ]; then
     DEVTOOLS=yes
+    SERIAL_CONSOLE=yes
   fi
 
 # read options from $HOME if available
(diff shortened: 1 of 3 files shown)


stefansaraev:
I dont think so. this should be:

1. project specific.
2. avoid agetty if possible.

here is an example what I use on my wetek_play. http://sprunge.us/cBdT

vpeter4:
Yes, this systemd thing stefansaraev posted works perfectly.

haggaie:
I wasn't aware systemd could do that without agetty. Regarding being project specific, if you need it for WeTek, and I need it for Raspberry Pi, don't you think it would be helpful to have it somewhere in the common code?

stefansaraev:
nope. because x86 wont need it. and I, personaly, keep it in my build tree, not upstream (regular joe does never debug anything)

haggaie:
I'm new here so I might be talking nonsense, but how many regular joe's build there own image from source?

In any case, if I submit a pull request for (an optional) service file for Raspberry Pi only, would you consider it? I just want the next person that tries to debug something to find it more easily.

stefansaraev:
> In any case, if I submit a pull request for (an optional) service file for Raspberry Pi only ...

of course I would (as long others agree)

haggaie:
Okay, great :)

haggaie:
A question: from what I understand, putting the service file in the project's filesystem dir would always install it. If I want to conditionally install it (when OPENELEC_VERSION=devel for instance), it has to be in a package, right? Or is there another way?
What would be the best way to do that?

vpeter4:
I think this file could be always installed but started conditionally. Like using in service file
```javascript
ConditionKernelCommandLine="console=ttymxc1,115200"
```
I didn't have time to test this if really works. According to  stefansaraev it should. So only if console is set in boot params then service is started.

haggaie:
Okay, I'll try that. Thanks.

PR #4163, opened 2015-05-25
[RPi/RPi2] fix kodi patch

InuSasha:


popcornmix:
The correct solution is to pick up the latest patch:
https://github.com/xbmc/xbmc/compare/master...popcornmix:isenguard_rbp_backports.patch
rather than patching up an old version.

InuSasha:
ah it's really yours :)
want kind of ping you want?

popcornmix:
Any mention (i.e. @popcornmix) on github will notify me.
I'd suggest commenting on the commit that no longer compiles on my isenguard_rbp_backports branch if you just want to inform me.
If you want OE devs to see it too, then comment on their commit (i.e. projects/RPi2/patches/kodi/kodi-001-isengard-rpb-backports.patch)

PR #4164, opened 2015-05-27
[systemd] update to systemd-220

lrusak:
This could use some more testing as this systemd release was over a ~3 month period.

There is an issue with the current tarball as it includes generated files. This is noted in the package.mk.
The location of mount and umount must be hard coded now.

Any other suggestions are welcome.


stefansaraev:
I am against merging this for OE6.

lrusak:
I am too, I would only consider for master ;)

stefansaraev:
master IS in fact 6.0. for next few months, we should be careful with master.

sraue:
still testing 220 here (i have done now the last days), there is one issue to fix (still not sure if its in kodi or udev), other then this issue it works not bad...

sraue:
... this means this PR like it is cant go to master or 6.0 branch...

stefansaraev:
dont waste time on pkg bumps (especialy this one, if you push - I will revert, no matter how well it works for you)

focus on bugfixing instead. we are beta. thanks

lrusak:
Will 6.0 not be branched soon?

stefansaraev:
branched or not. systemd bump is not going in, even into master, in next six months.

lrusak:
ok :+1: then just leave this here for testing.

PR #4166, opened 2015-05-27
[linux] use version patches to avoid downloading entire source

lrusak:
This may not seem relevant to most, but here where we have a 150GB bandwidth cap every little bit helps.

This method uses the kernel version patches to change the kernel version. This is the way fedora does it in their rpm spec files.

Some thoughts:
* Doesn't effect amlogic and imx6 kernel versions
* The process is transparent
* Uses a ~5MB patch compared to a 80MB tarball when changing kernel version

I'd like to hear any suggestions

### DIFF
--- a/packages/linux/package.mk
+++ b/packages/linux/package.mk
@@ -41,7 +41,9 @@ case "$LINUX" in
     ;;
   *)
     PKG_VERSION="4.0.4"
-    PKG_URL="http://www.kernel.org/pub/linux/kernel/v4.x/$PKG_NAME-$PKG_VERSION.tar.xz"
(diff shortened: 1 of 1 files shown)


lrusak:
No interest?

vpeter4:
I'm building at work with almost unlimited network resources so don't really think of this. But at home it's a different story. That's why I'm using xdelta3 program to send only a diff of images to home.

lrusak:
@sraue @stefansaraev thoughts?

stefansaraev:
I avoid using unpack() at any cost ;)

joke aside, I dont think we need this.

PR #4168, opened 2015-05-27
revert some imh meh

stefansaraev:
@sraue @fritsch

### DIFF
--- a/packages/multimedia/imx-vpu/package.mk
+++ b/packages/multimedia/imx-vpu/package.mk
@@ -17,7 +17,7 @@
 ################################################################################
 
 PKG_NAME="imx-vpu"
-PKG_VERSION="5.4.27"
+PKG_VERSION="3.10.17-1.0.0"
 PKG_REV="1"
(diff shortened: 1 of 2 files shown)


fritsch:
I have no idea what the firmware does. Perhaps it's good, but I know for sure vivante v5 looses display / window ptrs and this is known since longer. 

I was warned by @mk01 months ago. Tested that over the long weekend ... it's a no go. Every second refreshrate switch we end up with black screen + menu sounds.

stefansaraev:
k. dropped the firmware revert. ready to go. merging in 48 hrs if no objections.

stefansaraev:
@fritsch ca4344d is reverted. do you still want this in (imx-vpu as before) ?

fritsch:
I will rebuild when I find time. I want that the Adjust Refreshrate stuff does not leave us with blackscreen anymore. One of those driver bumps has fault in the blob, which forgets egl ptrs.

No idea which one exactly. But bumping of drivers was not needed at all - I did not even see a changelog.

PR #4170, opened 2015-05-29
[linux config] Build btrfs into kernel

lrusak:
I'd like to see this included, It allows us to boot from btrfs partitions.

This will not be the official method, only for users who want to manually make their system.

After I'd like to add support for partition mount options in the init.


fritsch:
Can you summarize for me with some dashpoints, why one would should
consider btrfs?

I googled and found:
http://www.phoronix.com/scan.php?page=article&item=linux_317_4ssd&num=1

To my knowledge there is no fsck support. We should be very careful,
especially when users use it on their shitty usb sticks or on sdcards.

2015-05-29 8:33 GMT+02:00 Lukas Rusak <notifications@github.com>:

> I'd like to see this included, It allows us to boot from btrfs partitions.
>
> This will not be the official method, only for users who want to manually
> make their system.
>
> After I'd like to add support for partition mount options in the init.
> ------------------------------
> You can view, comment on, or merge this pull request online at:
>
>   https://github.com/OpenELEC/OpenELEC.tv/pull/4170
> Commit Summary
>
>    - [linux config] Build btrfs into kernel
>
> File Changes
>
>    - *M* projects/Generic/linux/linux.x86_64.conf
>    <https://github.com/OpenELEC/OpenELEC.tv/pull/4170/files#diff-0> (2)
>
> Patch Links:
>
>    - https://github.com/OpenELEC/OpenELEC.tv/pull/4170.patch
>    - https://github.com/OpenELEC/OpenELEC.tv/pull/4170.diff
>
> —
> <https://github.com/OpenELEC/OpenELEC.tv/pull/4170>.
>



-- 
                   Key-ID:     0x1A995A9B
                   keyserver: pgp.mit.edu

vpeter4:
Would adding INITRAMFS_MODULES="btrfs" in projects/Generic/options help?

stefansaraev:
-1 from me.

lrusak:
@vpeter4 possibly, I missed that, will have to test.

btrfs has it's own fsck utility, btrfsck.

btrfs has many advantages, COW, subvolumes, compression. These aren't relevant to the OpenELEC boot volume but would still be nice to have the option to boot from an entirely btrfs formatted disk

fritsch:
btrfsck is deprecated, see: https://btrfs.wiki.kernel.org/index.php/Btrfsck

lrusak:
In which case you would need btrfs-progs, which I have here,
https://github.com/lrusak/OpenELEC.tv/commit/36939803f43c3b4da0d662d5b7030882eb7022ee

Anyways, this is just a can of worms and I will look into it again after OE 6.0 is released ;)

fritsch:
Don't close such PRs right away.

I want to understand why btrfs should be considered, what are the
advantages? Why should we have it?

2015-05-29 9:50 GMT+02:00 Lukas Rusak <notifications@github.com>:

> Closed #4170 <https://github.com/OpenELEC/OpenELEC.tv/pull/4170>.
>
> —
> <https://github.com/OpenELEC/OpenELEC.tv/pull/4170#event-316968378>.
>



-- 
                   Key-ID:     0x1A995A9B
                   keyserver: pgp.mit.edu

ekalin:
I was also affected by the lack of btrfs support when I upgraded to 6.0.0.

I can tell why btrfs is useful. It allows me to revert failed upgrades. It allows me to go back to the previous version when something fails. As a matter of fact, it was the fact that I used btrfs that allowed me to get back a working openelec system (exactly as it was before) in minutes when the 6.0.0 version failed to boot.

Before an upgrade, I create a snapshot of the current state of the /flash and /storage filesystems. I then upgrade, but if I ever want to go back, I can use the saved snapshots to restore the system to exactly as they were before.

There are other ways to make backups and restore, sure. I'm just saying that this is a very convenient one.

More importantly, having that filesystem built-in won't affect normal installations in any way, except for using a little more RAM for the kernel. It'll be unused by the vast majority of users, but it does no harm, nor it requires any support commitment from the developers - just enable the switch and forget about it. If a user decides to do a manual installation and use btrfs, he'll be aware of the consequences and how to handle any problem caused by that choice.

So I'm asking that this issue be reconsidered and that btrfs support be included in the kernel back again.

PR #4171, opened 2015-05-29
Fix typo "user-friendly" properly

tobbi:
According to dict.cc, "user-friendly" is more common. See http://www.dict.cc/?s=user-friendly

;-)


stefansaraev:
oh boy, so much pain for just NOOBS os.json ....

PR #4176, opened 2015-06-02
[RBP] A collection of pending RPi fixes

MilhouseVH:
The mailbox and overlays commits will be required by the latest RPi kernel (whenever you merge this).


popcornmix:
Thanks @MilhouseVH 
Changes look fine (but wait for next kernel bump before merging).

sraue:
@popcornmix can you inform me per email if you committed the last kernel update patches please? thanks much

popcornmix:
Rpi kernel drivers are gradually being accepted upstream. The upstream versions are generally rewritten, but based on downstream commits. They often have 2835 in the name rather than 2708.

We are slowly switching to upstream  drivers where they have equivalent functionality. This often needs a change to .config options (e.g. enabling the 2835 version of the driver).

We are also moving some of the non-currently-upstreamed drivers out of the mach-bcm2708 directory into the correct driver place in the kernel tree and adding the ability to load them through device tree. This allows the downstream drivers to be used with the upstream mach-bcm2835 kernel.

The eventual goal is to make downstream mach-bcm2708 and upstream mach-bcm2835 kernels become as close as possible. We can then switch to using the mach-bcm2835 kernel with a much smaller number of patches.

We are doing this in a number of small steps to reduce the risk. The general plan is I ping @Milhouse about any build changes. He adds it to the next nightly build, and assuming no complaints he'll submit a PR a couple of days later.

I can include @sraue in the ping of any changes if it's useful for you to know what's going on. Or if you ping me/@Milhouse before bumping the kernel, we can let you know if there's any outstanding build changes required.

sraue:
added in
https://github.com/OpenELEC/OpenELEC.tv/commit/277700ac32a7a1de2ae9011441d9989593aea950
https://github.com/OpenELEC/OpenELEC.tv/commit/0565d4a3dad6bab641271528e81dcdd7bd9000fa
https://github.com/OpenELEC/OpenELEC.tv/commit/2195fbb5900a73d1aca4c671f37b32c87946a79a
https://github.com/OpenELEC/OpenELEC.tv/commit/ac555915d926775e9bb714d2ff216a6422b81a26
many thanks!

PR #4177, opened 2015-06-03
project imx6: add sgtl5000 analog output alsa config file

vpeter4:
needed by TBS matrix and probably Solidrun's HummingBoard-i2eX


fritsch:
Fine with me. Thanks much.

PR #4183, opened 2015-06-05
sundtek-mediatv: rebuild for new devices support

vpeter4:


stefansaraev:
pls also bump changelog

PR #4186, opened 2015-06-06
libcec: build python wrapper

Forage:
Enables building the libcec Python wrapper included in libcec 3.0, in addition to the other wrappers.

It would handy to be able to control libcec from Python.
In my case I've got two cec adapters in my RPi, one to control Kodi, one to control a set-top box through a Python script running on the RPi.

### DIFF
--- a/packages/devel/libcec/package.mk
+++ b/packages/devel/libcec/package.mk
@@ -54,7 +54,6 @@ configure_target() {
 
   cmake -DCMAKE_TOOLCHAIN_FILE=$CMAKE_CONF \
         -DBUILD_SHARED_LIBS=1 \
-        -DSKIP_PYTHON_WRAPPER:STRING=1 \
         -DCMAKE_INSTALL_PREFIX=/usr \
         -DCMAKE_INSTALL_LIBDIR=/usr/lib \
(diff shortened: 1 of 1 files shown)


fritsch:
That one is considered useful. If it does not break anything +1 from me.

MilhouseVH:
Presumably using the Python wrapper will "break" Kodi CEC communication in the same way cec-client already does, if Kodi is active whenever the wrapper/cec-client is used (as only one CEC client can be active at a time).

That's not an objection, just trying to understand/confirm any potential limitations.

Forage:
@MilhouseVH: It will break communication if you touch the CEC adapter used by Kido yes. It can/will happen less often then cec-client though.
If you communicate directly to a different adapter then nothing will happen to the one used by Kodi.
Use a just a scan function like "DetectAdapters()", which tries all available adapters, and your connection will be broken.
cec-client does this straight away, using the wrapper it's up to you.

The wrapper can be tested by running "/usr/bin/pyCecClient.py"

stefansaraev:
I think this has been discussed before, but I cant find the issue.

@MilhouseVH is right. but I dont mind including this, IMO it's up to the user to break their own systems, or not. so, ACK.

up to @sraue

PR #4196, opened 2015-06-15
mkpkg: fix VERSION information for kodi

HiassofT:
I just noticed that the reported git revision in kodi.log and system information looked oddly unfamiliar. Instead of the expected daedd5f on a current master build I saw 42fb669 - the gitrev of the pvr.wmc addon.

I think GIT_REV in mkpkg_kodi should actually be GIT_HASH so that copy_sources creates the VERSION file with the correct info.


stefansaraev:
good catch. thanks ;)

PR #4198, opened 2015-06-15
WeTek_Play: Make ugly IEC958 workaround less ulgy

kszaq:
This patch makes my stereo-only USB sound card *not* look like a S/PDIF device.

After this patch DD and DTS passthrough still work fine.

### DIFF
--- a/projects/WeTek_Play/patches/kodi/0017-enable-IEC958.patch
+++ b/projects/WeTek_Play/patches/kodi/0017-enable-IEC958.patch
@@ -2,17 +2,16 @@ diff --git a/xbmc/cores/AudioEngine/Sinks/AESinkALSA.cpp b/xbmc/cores/AudioEngin
 index e22db7a..dcdaf2e 100644
 --- a/xbmc/cores/AudioEngine/Sinks/AESinkALSA.cpp
 +++ b/xbmc/cores/AudioEngine/Sinks/AESinkALSA.cpp
-@@ -1291,6 +1291,12 @@ void CAESinkALSA::EnumerateDevicesEx(AEDeviceInfoList &list, bool force)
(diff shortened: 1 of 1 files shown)


fritsch:
Fine with me :-) Though, to be honest, it's as ugly as before and also think of other AML devices that might use this code path (not officially, of course).

It needs fixing on the kernel / alsa side. Just rename the device to something IEC958 ... and we don't need any hack anymore.

fritsch:
@kszaq https://github.com/OpenELEC/OpenELEC.tv/blob/master/projects/imx6/filesystem/usr/share/alsa/cards/imx-spdif.conf <- if you come up with something like that, we can just rename the device and don't have to care in kodi at all.

kszaq:
@fritsch I've been using this for some time with Meson8 (https://github.com/kszaq/OpenELEC.tv/blob/50a8c94422bd7e6dc5ce8dc23d20e928aed2dbff/projects/Amlogic/devices/MK808B.Plus/filesystem/usr/share/alsa/cards/SOC-Audio.conf) but for some reason this doesn't work for Meson6, i.e. WeTek Play. Audio type is detected properly by kernel driver - you can see CT_AC_3 and CT_DTS popping in dmesg - but all you get is silence on both SPDIF and HDMI.

We need a fix on kernel side first.

fritsch:
Jep - I basically don't like "hacks" in kodi, that workaround BS in other layers. I only wrote this ugly fix, as the "other workaround" we had before _broke_ something. It announced 8 channels, but the hw can only open 2 pcm channels byself ... (AC3, DTS only need those two anyways).

fritsch:
And as we are at it: I asked 4 or 5 times already, when I reviewed the audio patches - please someone of you PR this stuff to Kodi upstream. There we can decide which can go directly in and what needs adjustments ... you should find my comments somewhere in git history.

PR #4200, opened 2015-06-16
Configure RPi/RPi2 platform with optimal CPU and IO settings

MilhouseVH:
The following settings are (or are about to become) the default on other distributions (Raspbian, OSMC etc.) and it is recommended they should also be the default for OpenELEC too:
```
echo 1 > /sys/devices/system/cpu/cpufreq/ondemand/io_is_busy
echo ondemand > /sys/devices/system/cpu/cpu0/cpufreq/scaling_governor
echo 50 > /sys/devices/system/cpu/cpufreq/ondemand/up_threshold
echo 100000 > /sys/devices/system/cpu/cpufreq/ondemand/sampling_rate
echo 50 > /sys/devices/system/cpu/cpufreq/ondemand/sampling_down_factor
```
The existing values are:
```
rpi22:~ # cat /sys/devices/system/cpu/cpufreq/ondemand/io_is_busy
0
rpi22:~ # cat /sys/devices/system/cpu/cpu0/cpufreq/scaling_governor
ondemand
rpi22:~ # cat /sys/devices/system/cpu/cpufreq/ondemand/up_threshold
50
rpi22:~ # cat /sys/devices/system/cpu/cpufreq/ondemand/sampling_rate
355000
rpi22:~ # cat /sys/devices/system/cpu/cpufreq/ondemand/sampling_down_factor
1
```
`up_threshold` is also being set (to 50) by the [cpufreq-threshold](https://github.com/OpenELEC/OpenELEC.tv/blob/master/packages/linux/system.d/cpufreq-threshold.service) service - if this is only for the benefit of the RPi/RPi2, then this service could be eliminated (I can squash removal into this PR, if desired).

This should be OK for master.

### DIFF
--- a/projects/RPi/initramfs/platform_init
+++ b/projects/RPi/initramfs/platform_init
@@ -1,0 +1,28 @@
+#!/bin/sh
+
+################################################################################
+#      This file is part of OpenELEC - http://www.openelec.tv
+#      Copyright (C) 2009-2015 Stephan Raue (stephan@openelec.tv)
+#
+#  OpenELEC is free software: you can redistribute it and/or modify
+#  it under the terms of the GNU General Public License as published by
+#  the Free Software Foundation, either version 2 of the License, or
+#  (at your option) any later version.
+#
+#  OpenELEC is distributed in the hope that it will be useful,
+#  but WITHOUT ANY WARRANTY; without even the implied warranty of
+#  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+#  GNU General Public License for more details.
+#
+#  You should have received a copy of the GNU General Public License
+#  along with OpenELEC.  If not, see <http://www.gnu.org/licenses/>.
+################################################################################
+
+# Enable io_is_busy for improved sdhost performance - essentially, equivalent of force_turbo=1 but for mmc
+echo 1 > /sys/devices/system/cpu/cpufreq/ondemand/io_is_busy
+
+# Configure frequency scaling properties - should improve performance a little (turbo, in most cases)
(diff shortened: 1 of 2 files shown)


popcornmix:
I can confirm these are the planned default settings for raspbian.

It should improve general performance a little by engaging turbo mode more often.

The io_is_busy setting is important for the new "sdhost" driver which is currently optionally available (by adding ``dtoverlay=sdhost`` to config.txt), but we'd like to switch to it being the default after further testing.

Without that setting "sdhost" performance can be reduced when not in turbo mode.

stefansaraev:
ack from me

> up_threshold is also being set (to 50) by the cpufreq-threshold service

not sure, it's probably good for imx6/wetek, too. I'd keep it as is for now.

MilhouseVH:
>  I'd keep it as is for now.

Fortunately it's using the same value that the RPi/RPi2 needs, but it's easy to see how it could become a problem if a different value is required. In future perhaps it should be moved to platform_init for imx6/wetek.

stefansaraev:
> Fortunately it's using the same value that the RPi/RPi2 needs

thats right, I'll take care when this PR is merged. just ping me ;)

MilhouseVH:
Thanks, will do.

BTW, to see the impact of io_is_busy (aka turbo) at higher core overclocks when using the new sdhost driver, I posted the results of a very simple benchmark [here](http://forum.kodi.tv/showthread.php?tid=224025&pid=2005396&highlight=io_is_busy#pid2005396), which documents the performance at different clock settings, with and without turbo.

sraue:
@MilhouseVH can you change the copyright line in both files :-) other than that if it works it has my ok

MilhouseVH:
Done, good catch. :)

Yes, tested on both RPi and RPi2 - both working as expected.

PR #4215, opened 2015-06-26
Add unRAID Project

eschultz:
The unRAID project contains a special kernel config with additional virtio and 9p drivers to support OpenELEC as a VM on unRAID (http://lime-technology.com/).


chewitt:
If we add a project to the official project buildsystem it implies we support the project. I appreciate the relative simplicity of these changes but I doubt any of the project staff have unRAID boxes; definitely none with OE installed, and thus we cannot provide support and should not accept this. I believe it should be managed via your own project fork as for various other OE derivatives. Should we see active users stats that indicate running OE on unRAID is wildly popular (the project name will be distinct and visible) we can always reconsider.

stefansaraev:
-1. I afree with @chewitt

jonpanozzo:
First and foremost, thank you guys for so quickly reviewing this and getting back to us. Accept or not, just getting feedback so quickly is greatly appreciated. 

We have no problem maintaining the fork at this point and will do just that. Do you guys have any issue with us referring to it as OE on unRAID by name and leaving the OE logo inside?  If so, I suppose we could rebrand, but we'd much rather not.

stefansaraev:
^ I, personaly, dont mind if it is not rebranded, as long users dont report bugs to me ;)

chewitt:
I'll close the PR. I see no harm as long as the support route is clear to users. There are currently four of them so it should all be first-names and no bother ;)

jonpanozzo:
If you guys desired, we could maintain a subforum off yours for virtualization support so that you guys could redirect users without asking them to leave your forum.  We can save that for another day if you'd like to see adoption first.  We haven't publicly handed out our build to very many folks yet, as we wanted to be respectful of your trademarks on the name and didn't want to create support confusion.

We have a decent amount of users with pent up demand for us to release, so I expect you will see much more in the future.

jonpanozzo:
And thanks again for your speed in communicating with us.

anoma:
This ticket is confusing worded. As I understand it, it is requesting that a 3rd party resurrect the old "never supported" virtual build for a small subset of virtulisation types.

Virtualisation is the request that will never go away so if it can be developed and supported by a 3rd party it is win win.

My advice, prove this by pull requests before even considering a ticket of this magnitude.

jonpanozzo:
?  This was a pull request, not a ticket.  But maybe I'm just misunderstanding what you're saying.

anoma:
My apologies I was being blind.

In that case I am doubly confused. Other than the misleading unRAID tag (i.e. this isnt unRAID support its partial virtulisation support which unRAID can consequently use) this seems like a win win for us.

Can this not be marked as unsupported and included, rather than there being a complete fork out there in the wild? It can always be removed later if it doesnt work out like the last virtual.

stefansaraev:
> Can this not be marked as unsupported and included

no.

PR #4219, opened 2015-06-28
kdbus support

lrusak:
Purely for testing purposes for those that are interested! Of course this will depend if kdbus is eventually merged into the mainline kernel.

Seems to work fine for me.

kdbus is now always built into systemd-221 and later but the --enable-kdbus switch is needed to avoid using kdbus=1 as a kernel command line argument.

systemd-bus-proxyd is needed for connman.

~~For those who want to test this put this PR on top of #4266 as I developed this with latest systemd~~
The latest systemd version is now included in this PR as removing it would cause merge conflicts.


chewitt:
closing at request of @lrusak - thanks

PR #4223, opened 2015-07-02
Fix 16bpp plymouth-lite patch which doesn't use stride consistently

MilhouseVH:
See https://github.com/OpenELEC/OpenELEC.tv/pull/4221#issuecomment-117840719

Thanks @popcornmix, @samnazarko


stefansaraev:
thanks @MilhouseVH I'll do runtime test tonight

@fritsch the patch is fine ;)

fritsch:
Okay, then in with it.

stefansaraev:
all good. should we merge now or you wait for some more feedback @MilhouseVH ?

thanks again to everyone involved ;)

fritsch:
Revert is always an option

PR #4226, opened 2015-07-06
Bump libshairplay in anticipation of Kodi PR7093

MilhouseVH:
See https://github.com/xbmc/xbmc/pull/7093

Not sure how the package for distribution is built - does this happen automatically once this PR is merged?


stefansaraev:
is this safe for master / isengard ?

> Not sure how the package for distribution is built - does this happen automatically once this PR is merged ?

no. we use mkpkg_shairplay (manualy) to generate a tar and then upload it to sources.openelec.tv. already tone ;)

MilhouseVH:
> is this safe for (nijna edit: OE master) master / isengard ?

Certainly should be OK for master (and needed by the Kodi commit), both have been tested in my builds for a couple of months with positive results, but probably of less importance for Isengard without PR7093. Although I'd expect the updated package to be OK in Isengard, perhaps @Mempiz can confirm (I don't use any of this myself - no Apple gear).

> already done ;)

Thanks!

stefansaraev:
sorry. I meant, that OE master will ship kodi isengard for next few months, not kodi master ;)

if @Memphiz says bumped shairplay is ok with Isengard - push the button.

Memphiz:
But this PR is against master - if OE stays on isengard - this wouldn't matter at all no?

Memphiz:
ops  - sry - you are quick ... this is the OE repo *now realises* haha

Memphiz:
bumped shairplay should be 100% backward compatible [tm] and ok for isengard.

PR #4227, opened 2015-07-11
RTL8812AU: Update driver with improved 5GHz support

MilhouseVH:
I've added this updated driver (47 new commits - the old repo seems to have stagnated) and [initial feedback](http://forum.kodi.tv/showthread.php?tid=230966&pid=2048753#pid2048753) seems positive, but needs more testing.

This is the [commit comparison](https://github.com/wuzzeb/rtl8812AU_8821AU_linux/compare/wuzzeb:b22cbdf3b9df225...abperiasamy:6280806f6ba5bef) between the old and new repo.

Might be worth adding to RC1 if there's time, or a later release if not? Certainly the current driver has problems, so this new driver can't be much worse...


sraue:
added in https://github.com/OpenELEC/OpenELEC.tv/commit/dec8e1fe2399c6066f56797dd5b93c8956a14e6b

PR #4230, opened 2015-07-14
Update vdr-plugin-xvdr-01_makefile.patch

frankr612:
Fixes the 2GB file limit for playing recordings on RPi(2)


stefansaraev:
xvdr is no more. and we dont update addons in 5.0 repos.

PR #4231, opened 2015-07-14
wetek_play: update bootloader

stefansaraev:
@nelim reported that, after installing fresh OE 6 beta1 on nand, it's not possible to boot from sdcard anymore (he had latest u-boot manualy flashed before) due to u-boot downgraded to 2015-03-04

this PR updates u-boot to latest (2015-05-04) - extracted from DB-J569-wetek-ota-2015-05-04

//cc @codesnake 

btw, why is a bootloader included in update zip, at all ?


codesnake:
u-boot from Android will not work correctly with UBI filesystem. OE uses modified u-boot. As far as I remember it didn't have any problems botting from sdcard.

stefansaraev:
> u-boot from Android will not work correctly with UBI filesystem.

o_O. ok. you just lost me. I give up on this.

stefansaraev:
indeed. this fails to boot oe from nand. ~!@#$%^&%$#@!~

however, with "openelec bootloader" - sd card is completely skipped. users will complain. here some very valid use cases

1. test drive new openelec version before installing it onto nand
2. test OpenATV.
3. test whatever else.

I am done with all this nand stuff. feel free to ping whoever is in charge at wetek. "distribution specific" bootloaders are meh meh.

codesnake:
We need this "distribution specific" bootloader because we use nand layout that is different from Android. If you remember we did that to increase the capacity of data partition.

stefansaraev:
> We need this "distribution specific" bootloader because we use nand layout that is different from Android. If you remember we did that to increase the capacity of data partition

what I remember, was a big meh:
1. you boot openelec from sdcard.
2. it changed the layout!!
3. you can NOT boot android anymore.
4. o_O

well, that was fixed. now, what happen is.
1. you install OE to nand
2. you can NOT boot from sd card anymore!
3. o_O.

that's all wrong if you ask me. but well, this time I wont make too much drama out of it. I'll just tell my friends "hey, you should never, ever install OE on nand!". I already prepared android zips to recover original os and bootloader for them.

codesnake:
@stefansaraev What is your propositions about the issues?

stefansaraev:
> @stefansaraev What is your propositions about the issues?

right now, I have none.

months ago, I expressed my concerns, some drama happened because we (nand driver) were doing so much nonsense, pissing of users. then I said "if you wand nand builds, I want them to NOT break sdcard builds". 

this should have been handled internaly at wetek. now it's a bit late (OE branded box is released, we have to support it). 

if it is not possible to have an unified bootloader, at least it should be documented that once you install openelec on NAND, you should NOT expect sd boots to be working)

codesnake:
My proposition is to abandon the UBIFS "system" partition and load kernel as image from it (the same way as Android does). This will allow us to still have larger data partition, use the same bootloader as Android. NAND layout update will no longer be dangerous, because while the nand layout is different, but partition names and kernel load mechanism are the same.
However this will not allow to implement uEnv.txt loading in bootloader, because there will not be any filesystem on system partition, just kernel image.

stefansaraev:
"raw" kernel image partition is bad if you ask me. and well, I never understood why ubifs was required for OE. android's /system is ext4.. u-boot can load / boot stuff from virtualy everywhere.

codesnake:
If you want to use the same bootloader for both Android and OE then we should use image for kernel. You can't read any other filesystem on raw NAND from within bootloader without NFTL layer (remember, it is not eMMC).

stefansaraev:
seriously ???
so you say that it is not possile, from "android" u-boot, to "fatload mmc 0 0xboo kernel.img; bootm" ?!

codesnake:
On sdcard - you can. In NAND - no.

stefansaraev:
> On sdcard - you can. In NAND - no.

whatever that means ^^^...

damn. its a bootloader, thats supposed to boot android, and whatnot. it's not some black box done by some "android guys" specialy for android. some dude at wetek has the damn source.

okay, I leave this. I dont give a single f** from now on. you just voided all my knowledge in how bootloaders work.

codesnake:
The problem is not in bootloader, but in NAND memory and how it works.

codesnake:
I just mean you can't read FAT or EXT filesystem on NAND partition from within bootloader.

vivianwang1999:
@codesnake ,I have just download openelec 4.2.0 and it used amlogic linux kernel 3.0.x. But when compile this kernel cannot be found. Could you please help to send this kernel to me? Thanks very much.

stefansaraev:
@vivianwang1999 this is not a help desk right there. thanks

vivianwang1999:
@stefansaraev I'm sorry. I'm a newer for github.

Issue #4232, opened 2015-07-14
Openelec partition too small


thelightning87:
Hello,

I manage several OpenElec setups at friends. I can do this quite easily because i just create a Kodi setup, zip the profile and put the zip on the Openelec installer USB disk.
After install, i remove /storage/.kodi and unzip the file from the USB stick to /storage.

With latest OpenElec versions, the FAT partition has been limited to only 255MB. Not leaving enough space for my profile causing me to have to carry other storage device for the profile and not bieng able to script anything handy.

Is there a specific reason the partition has been limited?

chewitt:
Closing because user support is provided via the forums. This is a development team bug tracker.

vpeter4:
You can put backup.tar.bz2 or backup.zip on installer disk and it will be automatically unpacked when installed to target disk.

https://github.com/OpenELEC/OpenELEC.tv/blob/05d1c76b9f6ed9ec24d3ea872d7f931e1e17b5fd/packages/tools/installer/scripts/installer#L359

thelightning87:
Thank you for the information vpeter4.
Unfortunately, the FAT partition is still to small (only 150MB or so available) and my profiles are almost 400MB.

PR #4235, opened 2015-07-16
installer clean up

stefansaraev:
'ssh' in syslinux.conf is not a good idea (anymore). must go.
backups.. this needs ack from @vpeter4 but "bug reports" like #4232 are meh if you ask me.

### DIFF
--- a/packages/tools/installer/scripts/installer
+++ b/packages/tools/installer/scripts/installer
@@ -206,13 +206,7 @@ do_install_quick() {
       esac
 
       prompt_gpt
-      prompt_ssh
       prompt_backup_unpack
-
-      EXTLINUX_SSH=""
-      if [ "$SSH" = "1" ]; then
-        EXTLINUX_SSH="ssh"
-      fi
 
       # check for confirmation (twice!)
       MSG_TITLE="\Z1[ Confirmation before installing ]\Zn"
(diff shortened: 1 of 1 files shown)


vpeter4:
Why would you remove backups option? Does it hurt someone? If not leave it in.

ssh: Are you going to remove it also from sshd.service?

stefansaraev:
> Why would you remove backups option? Does it hurt someone? If not leave it in.

k, I'll drop the commit.

> ssh: Are you going to remove it also from sshd.service?

nope. I keep it there.

stefansaraev:
objections ?

vpeter4:
Not from me. Even if I don't see a reason why ssh should be removed from here.

stefansaraev:
I dont like to give clueless people who do DMZ an easy way to have sshd always runing. right from beginning.

for OE7, the way sshd start / runs should be re-workd (default keys only, big fat warning when you enable password auth). I'll do that and post a PR when done.

chewitt:
I have no objections to SSH support being removed from the installer but I do object to any removal of ssh support from the init script. Being able to force it on is a useful debug capability for some situations and leaving that option there does no harm.

chewitt:
NB: Our SSH default is OFF so user boxes are not accessible from local LAN, let along the internet, and no matter what scheme we have, idiots will still be idiots and share their box to the world. Any move to key-only auth as the default creates the situation of needing access to the box to install a private key to make it secure, but access not being available as SSH access requires a key, and a pre-shared known key is worse than a known password as n00b users will presume "Secure SHell" access to be secure. At least a default password is understood to be weak by most users. I would prefer to see a solution to non-changable passwords, and if that is truly impossible I believe we already have the "n00b vs security" balance correct.

vpeter4:
I think this ssh issue is something should be leaved alone as is. Security issues will always be present if user doesn't understand things. I would say it is the same thing if you connect some 12 V electronic to 230 V. Instructions clearly say for what voltage is. And if user use higher voltage the best thing can be only blown fuse. The worst whole device in flame. So it is just like enabling ssh. In installer and in Openelec settings in Kodi there is a warning explaining thing. If user doesn't understand what this is it should educate first.

CvH:
What if you move the ssh on/off switch to a more hidden place instead directly at the front page.
This would catch at least the "better activate everything the installer says" kind of people.
Also sshd conf default "AllowUsers user@192.168.0.0/16" could be helpful (don't know if this interacts with activated dmz well).

stefansaraev:
right now, I moved it to more hidden place -> syslinux.cfg ;)

anoma:
Restricting SSH access to only IANA private addresses removes most of the "fat fingers/n00b user" risks.

PR #4240, opened 2015-07-18
[libcec] Add repeating key support from popcornmix repo

MilhouseVH:
This has been in RPi test builds for several months now, and works well.

This PR should be fine for master, but would also be worth having in openelec-6.0 (if you need a backport, let me know).

There's an upstream PR https://github.com/Pulse-Eight/libcec/pull/111 with more details, but this won't be merged until after Isengard, hence this PR which can be dropped once upstream is merged.


popcornmix:
@Milhouse can you also add "[cec] Don't discard buttons when repeat mode is enabled" (from newclock4/isenguard_rbp_backports).

sraue:
will this go upstream? cc @opdenkamp

popcornmix:
@opdenkamp has added the "merge next version" tag to the PR: https://github.com/Pulse-Eight/libcec/pull/111

PR #4239, opened 2015-07-18
[RPi]: Enable config options required for next RPi kernel bump

MilhouseVH:
Good for master on next linux-01-RPi_support.patch bump.

Issue #4238, opened 2015-07-17
6.0 beta 3 broke snd-hifiberry-digi dtoverlay


foresto:
I upgraded from 6.0 beta 2 to 6.0 beta 3 on my Raspberry Pi 2, and found that my optical audio output no longer works. The device is no longer listed in Kodi's audio output options, and dmesg reports the following:

[    4.563762] snd-hifiberry-digi sound: ASoC: CPU DAI (null) not registered
[    4.563788] snd-hifiberry-digi sound: snd_soc_register_card() failed: -517
[    4.845912] snd-hifiberry-digi sound: ASoC: CODEC DAI wm8804-spdif not registered
[    4.845946] snd-hifiberry-digi sound: snd_soc_register_card() failed: -517

Here's the relevant line from /flash/config.txt:

dtoverlay=hifiberry-digi

HiassofT:
CONFIG_SND_SOC_WM8804_I2C=m is missing in the kernel config.

But there's also another issue, the RPi linux support patches needs updating. The devicetree settings are wrong and I2S cards won't work at all on B+ and RPi2.
https://github.com/raspberrypi/linux/issues/1046

Could you try if this build works (I don't have any Hifiberry cards)
http://www.horus.com/~hias/tmp/OpenELEC-RPi2.arm-6.0-devel-20150717133012-r21123-g4f65955.tar

Here's a tree with updated kernel patches and config (based on master, so 2 commits ahead of 6.0b3).
https://github.com/HiassofT/OpenELEC.tv/commits/rpi-i2s

I've taken the kernel config changes from the Milhouse builds. ping @MilhouseVH could you have a look at them if they are fine?

If the build works fine I can make a PR

MilhouseVH:
> ping @MilhouseVH could you have a look at them if they are fine?

Yep, looks fine, I've been including the Wolfson I2C driver in my builds for a [while now](http://forum.kodi.tv/showthread.php?tid=224025&pid=2032956#pid2032956) and I don't think it's had any positive effect where Hifiberry Digi is concerned.

These are the kernel config patches I'm using on top of OpenELEC master HEAD and rpi-4.1.y:

[Mailbox options](http://nmacleod.com/public/oebuild/patches/01_kernel_4_2.txt) - See [PR1037](https://github.com/raspberrypi/linux/pull/1037)
[...]

PR #4244, opened 2015-07-21
[RPi] add missing module for Hifiberry Digi card

HiassofT:
see #4238 

Note: RPi B+ and 2B also require a bump of linux-01-RPi_support.patch together with #4239


popcornmix:
Thanks. Would be good to get this into Isengard stable release.

PR #4246, opened 2015-07-24
remove irserver support

stefansaraev:
after switching to systemd, irserver was never fixed/reworked. now broken and nobody cares

do we have any users who still want/need this @sraue ?

EDIT: to be safe, dont merge now, wait for OE7 ?


sraue:
yes we have users who want this, and its still on my todo... because of the issues we actually have, i plan to rework support for some remotes so we maybe can deprecate eventlircd (not sure how well this will work) keep irserver in in for now, i will look at this

stefansaraev:
well, it seems irserver was broken in last 2 years. but ok :)

PR #4247, opened 2015-07-25
mediatek_mt7610u_sta_driver_linux-64bit (Trendnet TEW-806UBH)

cch2375:
Can support for the Trendnet TEW-806UBH usb wireless adapter be added to OpenELEC?  It has a mediatek chipset, I believe the mt7610u, according to what I have compiled on Linux Mint.  This device works great, has fantastic range (better signal quality), and throughput (streams HD better).  It really has allowed me to do more (more quickly) in Mint than the before with the on board wireless.  It would be a great add on device for hardware without decent wireless adapters, or for people wanting to connect to 5GHz if their current wireless doesn't support it.
Installation steps used with some edits (different adapter):
http://amazeballsysadmin.blogspot.com/2015/07/getting-mt7610u-edimax-7711macsabrent.html


cch2375:
I am sure there is a specific reason this request will not be
entertained, but not knowing why doesn't help prevent me from asking for
things again in the future that I shouldn't even bother to request, or
doing it incorrectly.

Did I request something that I shouldn't have?
Did I request it in the wrong place?
Did I do something I shouldn't have, or not do something I should have
in my request?
Was my request for V5 when V6 is coming the wrong thing to do?

Anything that you can help me out with as far as requests would be
greatly appreciated.  I don't want to be incorrect, or waste anyone's
time.  My only goal is to try to help out in any way I can to further
the development efforts the OpenELEC team puts forth.  Even if my Linux
capabilities and knowledge are limited, I would like to contribute in
some way.

Best Regards,
Chris

stefansaraev:
dafuq ? ^^^^

fritsch:
You send a PR with 1000 merge commits absolutely not related to your issue. You need to rebase your changes on latest master and then PR again.

fritsch:
And - you did that three times (!) today.

PR #4248, opened 2015-07-26
linux: add missing I2C functions to the Hauppauge HVR-2205 patch

trsqr:
I2C function modifications were missing from the original HVR-2205 patch.

Reported by user Hoffy in the forums:
http://openelec.tv/forum/82-dvb-t-t2-support/77386-hauppauge-hvr-2205-support#144375


stefansaraev:
\o/

PR #4254, opened 2015-07-28
linux/drivers/input/joystick/xpad: fix operation for xbox 360/one con…

MattDevo:
…trollers

merged from https://github.com/paroj/xpad

- fixed blinking LED on Xbox 360 Wireless Controllers
- only expose actually connected Xbox 360 Wireless Controllers
- fixed kernel warnings due to submitting active URB requests
- updated Xbox One controller force feedback
- controller still works after suspend/ resume
- Xbox 360 Wireless button mappings are now compatible with Xbox 360 (non-wireless) mappings

Signed-off-by: Matt DeVillier <matt.devillier@gmail.com>


fritsch:
@MattDevo was this patch done by you or is there another author? If another author exists, that author needs to be added. If it was done by you, please use a git patch.

MattDevo:
Hi Peter,

The fixes were posted as a full source file on github, which I listed in
the commit msg. I created the patch, will redo in the proper format

Regards,
Matt

PR #4262, opened 2015-08-03
buildsys improvements

stefansaraev:
instead of comparing "timestamps" (package.mk newer than $STAMP and so on), this aims to detect changes by comparing md5 sums of the whole package dir (via md5deep)

this is good for people who need to do huge rebases often (or switching branches back and forth without actualy building).

### DIFF
--- a/scripts/build
+++ b/scripts/build
@@ -53,8 +53,14 @@ STAMP=$STAMPS/$PACKAGE_NAME/build_$TARGET
 
 $SCRIPTS/unpack $PACKAGE_NAME
 
-if [ -f $STAMP -a $PKG_DIR/package.mk -nt $STAMP ]; then
-  rm -f $STAMP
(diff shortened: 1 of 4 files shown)


stefansaraev:
@sraue 
// cc @MilhouseVH this could be useful for you.

MilhouseVH:
Interesting, thanks. This could reduce the number of packages that are rm'd unnecessarily under various circumstances. I'll test this, but for the time being will still rely on timestamps when cleaning packages before a build (need to give more thought how best to use this as part of my process).

lrusak:
If this saves me from automatically rebuilding a bunch of packages when I simply switch branches then yes. I will give it a test.

stefansaraev:
@lrusak it does.

MilhouseVH:
Seems to be working fine, is the plan to backport this to openelec-6.0, as that's where there should be a  big benefit (maybe this will happen automatically once master is merged to openelec-6.0?)

stefansaraev:
depends on @sraue but yea, the plan is to merge master onto 6.0.

PR #4268, opened 2015-08-08
wetek/dvb: reduce debug log spam

stefansaraev:
@codesnake FYI

~~I got new wetekdvb.ko from Sasa and repackaged it.~~


stefansaraev:
please hold on. dont merge yet.

stefansaraev:
yep. the new blob was bad, causing stuttering. removed a commit. now safe.

note to self: blobs cant be trusted.

PR #4276, opened 2015-08-14
Use a temporary filename when creating new crashlog

MilhouseVH:
Creating a new crashlog sometimes takes a while, and requesting users to "upload the latest crashlog" has resulted in partial crashlogs being presented (ie. the crashlog is being created as they're uploading it).

Using a temporary filename until the crashlog is complete would help alleviate this problem.


MilhouseVH:
I've also experienced this myself, for example when kodi.bin is in a crash loop, if you stop/kill kodi the last crashlog may often be incomplete, so in this case the crashlog of value is the penultimate `kodi_crashlog*.log`. This PR will also fix this issue.

stefansaraev:
+1

PR #4283, opened 2015-08-19
[rbp]Add MediaTek WiFi driver option (unset)

MilhouseVH:
The MediaTek option is new in 4.1.6 and required for a successful build.

Also drop CONFIG_BCM2708_MBOX as this is no longer required.

### DIFF
--- a/projects/RPi/linux/linux.arm.conf
+++ b/projects/RPi/linux/linux.arm.conf
@@ -1,6 +1,6 @@
 #
 # Automatically generated file; DO NOT EDIT.
-# Linux/arm 4.1.2 Kernel Configuration
+# Linux/arm 4.1.6 Kernel Configuration
 #
 CONFIG_ARM=y
 CONFIG_SYS_SUPPORTS_APM_EMULATION=y
@@ -1225,6 +1225,7 @@ CONFIG_RT2X00_LIB_CRYPTO=y
 CONFIG_RT2X00_LIB_LEDS=y
 # CONFIG_RT2X00_DEBUG is not set
 CONFIG_RTL_CARDS=m
+# CONFIG_WL_MEDIATEK is not set
 # CONFIG_RTL8192CU is not set
 # CONFIG_WL_TI is not set
 CONFIG_ZD1211RW=m
@@ -3018,7 +3019,6 @@ CONFIG_COMMON_CLK=y
 # CONFIG_SH_TIMER_TMU is not set
 # CONFIG_EM_TIMER_STI is not set
 CONFIG_MAILBOX=y
-CONFIG_BCM2708_MBOX=y
 # CONFIG_ARM_MHU is not set
 # CONFIG_PL320_MBOX is not set
 # CONFIG_ALTERA_MBOX is not set
--- a/projects/RPi2/linux/linux.arm.conf
+++ b/projects/RPi2/linux/linux.arm.conf
@@ -1,6 +1,6 @@
 #
 # Automatically generated file; DO NOT EDIT.
-# Linux/arm 4.1.2 Kernel Configuration
+# Linux/arm 4.1.6 Kernel Configuration
 #
 CONFIG_ARM=y
 CONFIG_SYS_SUPPORTS_APM_EMULATION=y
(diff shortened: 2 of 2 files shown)


chewitt:
This need extending to other projects (Generic, etc.) and the current binary driver in packages/linux-drivers/ can be removed :)

MilhouseVH:
> This needs extending to other projects (Generic, etc.) and the current binary driver in packages/linux-drivers/ can be removed :)

Right, but can that be done yet, as it looks like the mt7601u driver isn't included in the kernel until 4.2, however the Pi 4.1.6 kernel now includes a backport of this driver which is necessitating these config options (see: "[config: Add Mediatek mt7601u wifi driver](https://github.com/raspberrypi/linux/commit/65fe96220f731c79830f8b7085a97551e6321d5e)" and "[backport: mediatek: mt7601u: from v4.2-rc3](https://github.com/raspberrypi/linux/commit/5d2b422d7f7d495c0710c17b195351f9b7be5f63)").

As I hadn't realised we were already building the mt7601u driver as a package, I'm left to wonder what is actually being built for the Pi - I guess we might be building the mt7601u driver "twice"?

Perhaps the [backport](https://github.com/raspberrypi/linux/commit/5d2b422d7f7d495c0710c17b195351f9b7be5f63) commit could be used to replace the existing mt7601u package for all projects, but with 4.2 just around the corner it's probably not worth it.

Of course the other option, and probably what should happen, is that we drop the MediaTek WiFi commits from the Pi kernel and close this PR, until we move to 4.2.

chewitt:
It's not my decision but if the next release is a v5.95.4 beta (for testing remote fixes) then I would include the backport in all projects and get it tested. If not I would leave this until master bumps to a 4.2 kernel. Do you know if there are kernel plans for other MT chips? ..the MT7610 is also starting to appear in user devices and forum requests.

MilhouseVH:
This is just my opinion, but to minimise risk for OE6 I'd continue with what has (presumably) been working up until now - ie. the existing package - and simply drop the extra mt7601u driver and config commits from the rbp kernel patch.

With the mt7601u driver in 4.2, we can drop the existing package and align all the project kernel configs when we switch to 4.2 (which presumably will be targeting OE7?)

It looks like we need someone to decide what to do, as having two mt7601u drivers in the rbp builds seems a little excessive...

And I thought this PR was going to be a trivial little change! :-)

> Do you know if there are kernel plans for other MT chips? ..the MT7610 is also starting to appear in user devices and forum requests.

Sorry, I've no idea about that. Maybe @popcornmix or @rcn-ee?

popcornmix:
We pulled in the mt7601u from the 4.2 kernel (into 4.1) after a request here: https://github.com/raspberrypi/linux/issues/1090.

Perhaps it's best to unset CONFIG_WL_MEDIATEK in the Pi config while we are on 4.1 (so we'll continue using whatever out-of-stream driver that was previously used by OE).
When moving to 4.2, hopefully the out-of-stream driver can be dropped on all platforms and the upstream driver enabled instead.

Obviously without detailed testing it's hard to say which driver will perform better, but if both are usable then obviously using the upstream driver would be preferable.

clivem:
Which driver are you using for your out of tree package OE package? 
FYI, if you are already using the kuba-moo driver, https://github.com/kuba-moo/mt7601u, that is what is now the upstream driver in 4.2, and the content of the back-ported to 4.1 patch, now merged into the rpi-4.1.y tree.

MilhouseVH:
> Which driver are you using for your out of tree package OE package? 

https://github.com/OpenELEC/OpenELEC.tv/blob/master/packages/linux-drivers/mt7601u/package.mk

It is kuba-moo - the last commit in the package is 42f4084 (23 April), while the in-kernel driver seems to include commits up to e67366a (8 Jun).

If I were to close this PR and create another that unsets CONFIG_WL_MEDIATEK, does this need to happen for all platforms or just RBP?

popcornmix:
I think CONFIG_WL_MEDIATEK is pi only at the moment (as we have pulled the driver from 4.2 into our 4.1 tree - other platforms won't have that yet).

When OE moves to 4.2, then all platforms will need some change (probably dropping the kuba-moo driver and adding CONFIG_WL_MEDIATEK=y)

MilhouseVH:
I've updated the commit.

The CONFIG_WL_MEDIATEK option is required by the RBP build, but left unset while the packaged driver is used.

The CONFIG_BCM2708_MBOX option is no longer required.

sraue:
added in https://github.com/OpenELEC/OpenELEC.tv/commit/ce9105f7157fb52a70a518ca3df59e33ef3e3fac and https://github.com/OpenELEC/OpenELEC.tv/commit/41b0ef9285a6022332f95f6fb93f13146c76f173

PR #4287, opened 2015-08-22
zlib: Add adler32_vec/slhash optimization patch

arco:
https://github.com/kaffeemonster/zlib


sraue:
for what? is there a plan this will be included upstream?

arco:
No, probably not. It adds some arch specific optimizations, like for arm neon. We used it in CyanogenMod some time ago.

lrusak:
Thanks for the PR but not really interested in maintaining out of tree patches :)

PR #4288, opened 2015-08-22
new package: lzop

arco:
Needed for those that wants to compress the kernel with LZO

Add the following to your kernel entry in package.mk:

PKG_DEPENDS_TARGET="$PKG_DEPENDS_TARGET lzop:host"

### DIFF
--- a/packages/compress/lzop/package.mk
+++ b/packages/compress/lzop/package.mk
@@ -1,0 +1,35 @@
+################################################################################
+#      This file is part of OpenELEC - http://www.openelec.tv
+#      Copyright (C) 2009-2015 Stephan Raue (stephan@openelec.tv)
+#
+#  OpenELEC is free software: you can redistribute it and/or modify
+#  it under the terms of the GNU General Public License as published by
+#  the Free Software Foundation, either version 2 of the License, or
+#  (at your option) any later version.
+#
+#  OpenELEC is distributed in the hope that it will be useful,
+#  but WITHOUT ANY WARRANTY; without even the implied warranty of
+#  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+#  GNU General Public License for more details.
+#
+#  You should have received a copy of the GNU General Public License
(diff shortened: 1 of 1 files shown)


packages/compress/lzop/package.mk
```diff
@@ -0,0 +1,36 @@
...
+
+PKG_NAME="lzop"
+PKG_VERSION="1.03"
+PKG_REV="1"
+PKG_ARCH="any"
+PKG_LICENSE="GPL"
+PKG_SITE="http://www.lzop.org/"
+PKG_URL="http://www.lzop.org/download/$PKG_NAME-$PKG_VERSION.tar.gz"
+PKG_DEPENDS_HOST="lzo:host"
+PKG_DEPENDS_TARGET="toolchain"
```
stefansaraev: PKG_DEPENDS_HOST="" and remove PKG_DEPENDS_TARGET

stefansaraev:
imo, packages that are referenced anywhere in the buildsystem do not belong in OE tree. I just sent a huge python clean up pr, and I plan more ;)

packages/compress/lzop/package.mk
stefansaraev: sorry. ignore the PKG_DEPENDS_HOST part. I am blind.

arco:
Yup, that's true. But building the kernel will fail if they have selected LZO as the compression, and they don't have lzop installed in their system. Maybe the package should be a build dependency instead?

stefansaraev:
up to @sraue if he wants this.. 
honestly, I dont care much. for me. gzip is enough

sraue:
how much benefit you have from lzo support? and which system?

arco:
Decompression speed is faster, at the cost of slightly larger file. LZ4 is even better though, but LZO should be supported in all kernels since 2.6.33 I think.

I definitely notices an improvement with both on my Amlogic S805 based stick, but I settled with using LZ4 though, since it seems to perform the best overall. So if you don't want this in, it's ok for me. :)

chewitt:
I'm closing this PR because if LZ4 is all-round better than LZO we wouldn't bother with LZO, thanks.

PR #4296, opened 2015-08-30
security fixes: update links to https, libressl to 2.2.3, libgpg-error to 1.20

qods:
Tested, working fine


qods:
Oh, sorry, i missed the note on the page from libressl:

The latest OpenBSD-stable release is 2.1.7. The latest OpenBSD-current release is 2.2.3. 

Of course you should not switch to unstable releases. I will update the patches

PR #4297, opened 2015-09-01
Reworked: Update libgpg-error, links to secure ones and GPL-versions

qods:
I cleaned up things


lrusak:
You're commits need to be more descriptive. Please use better names to describe what package the commit is effecting.
```
[ffmpeg] update url to use https
```
```
[libgpg-error] update url to use https and licence to GPLV2
```
```
[libgpg-error] remove patch that has been merged upstream
```
etc.

PR #4298, opened 2015-09-01
Include RPi 2 overlock mode preset

Forage:
From the official RPi config tool: https://github.com/asb/raspi-config/commit/4ee1fde44ee544a7eade9ecf94141eb40aabab60

A small fix for turbo to match the config tool and an additional RPi2 preset.
Several tutorials only talk about 1000-500-500, but it actually needs the over_voltage in this config to be stable. So I think it's good to have it mentioned in the default config comments.


popcornmix:
Okay by me. It's just a comment so is safe to add.

PR #4309, opened 2015-09-09
Update source location of jsoncpp

islipfd19:
jsoncpp is no longer maintained on sourceforge but can be found on github. Updated version to 1.6.5 as per Christian's input.

### DIFF
--- a/packages/textproc/jsoncpp/package.mk
+++ b/packages/textproc/jsoncpp/package.mk
@@ -17,12 +17,12 @@
 ################################################################################
 
 PKG_NAME="jsoncpp"
-PKG_VERSION="src-0.5.0"
+PKG_VERSION="1.6.5"
 PKG_REV="1"
 PKG_ARCH="any"
 PKG_LICENSE="GPL"
 PKG_SITE="http://www.kodi.tv"
-PKG_URL="http://garr.dl.sourceforge.net/project/jsoncpp/jsoncpp/0.5.0/$PKG_NAME-$PKG_VERSION.tar.gz"
+PKG_URL="http://sources.openelec.tv/devel/jsoncpp-$PKG_NAME-$PKG_VERSION.tar.yz"
 PKG_DEPENDS_TARGET="toolchain"
 PKG_PRIORITY="optional"
 PKG_SECTION="multimedia"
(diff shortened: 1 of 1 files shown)


MilhouseVH:
The version on `sources.openelec.tv` is `1.6.5`, not `0.6.5`.

`1.6.5` seems like a rather large bump compared with the old version `0.5.0`. Not saying it shouldn't happen, but risk of more breakage?

islipfd19:
i did have it right the first time, 1.6.5. I agree that it's a large version jump but this is the version I updated the package.mk file to as per Christian's remarks to your open issue.

stefansaraev:
this is no no no.

you'd better pick https://github.com/stefansaraev/OpenELEC.tv/commit/ca800009ba1423b3229bbc95e2022b11ebc702a8

packages/textproc/jsoncpp/package.mk
```diff
@@ -17,12 +17,12 @@
...
 
 PKG_NAME="jsoncpp"
-PKG_VERSION="src-0.5.0"
+PKG_VERSION="1.6.5"
 PKG_REV="1"
 PKG_ARCH="any"
 PKG_LICENSE="GPL"
 PKG_SITE="http://www.kodi.tv"
-PKG_URL="http://garr.dl.sourceforge.net/project/jsoncpp/jsoncpp/0.5.0/$PKG_NAME-$PKG_VERSION.tar.gz"
+PKG_URL="http://sources.openelec.tv/devel/jsoncpp-$PKG_NAME-$PKG_VERSION.tar.yz"
```
MilhouseVH: `yz` should be `xz`.

Although it sounds like this isn't going to happen anyway - thanks for trying! :)

islipfd19:
Updated commit as per @stefansaraev comments/recommendation. Thanks! @stefansaraev I see what I did there.

stefansaraev:
now. squash it down to one commit. :)

islipfd19:
Closed this and created a new one #4310 so that it's one commit.

chewitt:
@stefansaraev .. I thought the same about the large bump, but I figured "what the heck, lets see if it works" when building official for ATV the other week, and it worked. I suggest others confirm their success with it first before we accept the change.

PR #4311, opened 2015-09-09
linux: add Elgato EyeTV Sat v3 (maybe v2.x)

CvH:
This adds support for Elgato EyeTV Sat v3 or maybe 2.x (revision is a bit unclear).
It is the same hardware (as already supported v1, v2) with a different hw id.

ping @trsqr if this is a valid fix [(fix is working of course)](http://openelec.tv/forum/83-dvb-s-s2-support/78070-elgato-eyetv-sat#147269)

### DIFF
--- a/packages/linux/patches/4.1.6/linux-230-elgato-eyetv-sat-v3.patch
+++ b/packages/linux/patches/4.1.6/linux-230-elgato-eyetv-sat-v3.patch
@@ -1,0 +1,50 @@
+diff --git a/drivers/media/usb/dvb-usb/az6027.c b/drivers/media/usb/dvb-usb/az6027.c
+index 92e47d6..2e71136 100644
+--- a/drivers/media/usb/dvb-usb/az6027.c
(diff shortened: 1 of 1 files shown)


trsqr:
+1. Looks good to me. It seems to just add another device ID to an existing driver, which is something that rarely causes any non-desired side effects.

CvH:
I didn't submit it due to absolutely zero knowledge how this is being done.
Post patch at https://patchwork.linuxtv.org/project/linux-media/list/ ?

trsqr:
Basically you just email (only ASCII email please) your patch to the mailing list at linux-media@vger.kernel.org with [PATCH] in the header and it will end up to the patchwork tool.

Here's an example:
https://www.mail-archive.com/linux-media@vger.kernel.org/msg91710.html

You may want to read https://www.kernel.org/doc/Documentation/SubmittingPatches as well. This is getting a bit off topic here though, so if you want to talk more about this send me an email or contact via PM on some forums.

PR #4316, opened 2015-09-11
(fix) fix util-linux:host after 490ce24

stefansaraev:
- parted:host needs libuuid
- mkimage needs uuidgen. --disable-all-programs is meh

EDIT: and well. f**k wall. there is chgrp wtf in Makefile. no-go without fakeroot.

@sraue


sraue:
fixed in https://github.com/OpenELEC/OpenELEC.tv/commit/62835db32238fce7667b63d325e374aa77e51f7e

stefansaraev:
> fixed in 62835db

really? did you test it?

I am done with this.

PR #4319, opened 2015-09-12
libcec&libopenmpt, update version, update links, fix #4294

### LINKED ISSUE #4294: Openelec 5.95.3 - CEC loose connection with LG TV - (libcec 3.0.1 seems to solve the issue)
Hello,
I'm using Openelec 5.95.3 on my Raspberry Pi 2 and I have some trouble with CEC. I have a LG TV and it's working fine with a fresh boot. But after two days of inactivity, the Raspberry can't be seen anymore by the TV, so I have to reboot the Raspberry every time I don't use it too often.
After investigating, I found that libcec 3.0.1 solve this bug. So I installed an Openelec testbuild with libcec 3.0.1 and all was working ok. (http://forum.kodi.tv/showthread.php?tid=231092&pid=2044074#pid2044074)
I just installed Openelec 5.95.4 but I expect the same behavior, so could you update libcec version in the next beta?
Thanks in advance
Tillstaff

qods:


lrusak:
libcec was previously updated to version 3.0.1 then later reverted, not sure the reason why. Maybe @sraue knows.

qods:
I know why he removed that. It was because the download link broke and the most quick and simple solution was to just revert to older version then searching for a new download location.
Thats why i searched long time for a propper download location and find the github link i changed the package to. This should solve the problem for a long time :)

sraue:
will check this, not more sure why it was reverted

sraue:
openmpt update added in https://github.com/OpenELEC/OpenELEC.tv/commit/e3e071547d327a961ae6cbe2b9b6ff9c93e9a2bb

RealJohnGalt:
libcec patching fails with the libcec commit due to "build.*/libcec-3.0.1" directory expected, when the actual directory is "build.*/libcec-libcec-3.0.1"

qods:
I have no idea how to implement into openelec download command following:

wget -o $PKG_NAME-$PKG_VERSION.tar.gz https://github.com/Pulse-Eight/libcec/archive/$PKG_NAME-$PKG_VERSION.tar.gz

Could someone do that and reuse the patch i did?

lrusak:
@qods please pick lrusak/OpenELEC.tv@b2859b6b594adfef99d46b4ce44b5f057b02fc8f

stefansaraev:
nope. https://github.com/Pulse-Eight/libcec/issues/132

Issue #917, opened 2012-07-08
documentation - md5 now needed for manual update


muzzamo:
The page at 
http://openelec.tv/installation/updating is incorrect when it says "(ignore the MD5 files)"

When updating from 2.0 Beta 4 to 2.0 Beta 5 (not sure about the others), these files are needed for the upgrade to proceed.

PR #4324, opened 2015-09-16
Allow building with xf86-input-libinput when specified

lrusak:
I'd like to still use libinput for testing on my builds (others may too) so adding LIBINPUT="yes" to you local options file will build xf86-input-libinput instead of xf86-input-evdev.

Also bumping libinput to version 1.0.1

### DIFF
--- a/packages/virtual/x11/package.mk
+++ b/packages/virtual/x11/package.mk
@@ -47,8 +47,14 @@ if [ -n "$WINDOWMANAGER" -a "$WINDOWMANAGER" != "none" ]; then
 fi
 
 get_graphicdrivers
-# Drivers
+
+# Drivers 
+if [ -n $LIBINPUT ]; then
+  PKG_DEPENDS_TARGET="$PKG_DEPENDS_TARGET xf86-input-libinput"
+else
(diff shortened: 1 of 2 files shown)


stefansaraev:
this is wrong. now xf86-input-libinput is always installed.

stefansaraev:
and we probably have the good old remote bug again. 

fyi 'if [ -n $boo ]' is NOT same as 'if [ -n "$boo" ]'

stefansaraev:
also, when you (again) switch to libinput, please, doubleecheck that keyboard layouts (and switching work. thanks.

MilhouseVH:
Switching between libinput and evdev requires a clean build - you're right that with this PR we're now building libinput by default, sigh (hadn't noticed any remote issues as I'm yet to do a clean build since this merged).

stefansaraev:
my suggestion is to completely remove packages/wayland but well. nothing more to say.

you wont see any remote issues on non-x86(xorg)

MilhouseVH:
> you wont see any remote issues on non-x86(xorg)

I know, I build x86 too - currently watching on a Revo 3700 with my own #0917 build that includes this PR and no remote issues. Suspect if I build clean then I will have issues.

PR #4338, opened 2015-09-23
installer: get rid of fs labels. use UUIDs instead.

stefansaraev:
here a quite common use case: joe has two fixed disks. joe has openelec 4.x on
disk1, joe wants to install ~~OE6~~ OE7 on disk2 for testing.
so does joe, but next boot, shit happens (blkid / boot order hazzard in init)
disk1 is /flash, disk2 is /storage - new install fucked up, or disk2 is /flash, disk1 is
/storage - old install fucked up  or even worse.. joe mad at ~~us~~ you (you, because I dont care too much)

this fix is imo long overdue, ~~but please do NOT merge for OE6. hold on until master is open for OE7/jarvis~~

this PR also removes the (partialy, on uefi only) broken "custom install / restore" feature. I have no intention to fix it. honestly, I think it is not needed at all and should go. so, please no bikesheding, if you want that feature, feel free to fix it, I'll be more than happy to include your fix in this PR.

### DIFF
--- a/packages/tools/installer/scripts/installer
+++ b/packages/tools/installer/scripts/installer
@@ -377,124 +377,6 @@ do_install_quick() {
   esac
 }
 
-do_install_custom() {
-  # show menu
-  MSG_TITLE="\Z4[ REPAIR/UPGRADE MENU ]\Zn"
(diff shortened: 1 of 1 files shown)


stefansaraev:
well. this is now tested by at least 3 people (4, including me). up to you to decide if it's worth for 6.0

PR #4343, opened 2015-09-27
add nvme block device support to kernel/installer

stefansaraev:
ref http://openelec.tv/forum/64-installation/78345-openelec-installation-no-devices-available

needs testing.


stefansaraev:
I compiled a tstbuild for Fux123 (at irc) and he reported all good

this is safe for master/OE6

I did not touch Nvidia_Legacy kernel because it's very unlikely that someone will need it.

fritsch:
Verified working on IRC. Thanks very much

PR #4348, opened 2015-10-01
Update Amlogic kernel and WeTek proprietary DVB module

codesnake:
* Update Amlogic kernel to 3.10-753dc30 (3.10.89)
* Update WeTek proprietary DVB module to 20150930


sraue:
@stefansaraev ok for you for master/6.0?

stefansaraev:
please hold on. I will test this tonight.

stefansaraev:
so far so good. no visible regressions. everything works as it should

btw, this seems to fix a sdcard corruption issue for me (I dont remember when it started, and did not bisect, because I thought my box is dying.. or just 3 bad sdcards... )

assuming 1. it may fix the sdcard corruption for others and 2. at wetekforums, wetek people will share builds with this PR in, anyway - better merge this soon than later ;)

@codesnake FYI https://github.com/stefansaraev/linux/commits/amlogic-3.10.89 some monkey rebasing work. 1:1 same as amlogic-3.10.y branch + http://sprunge.us/CZKN <- this is upstream and was reverted in your tree at some point. it would be nice if you if you can use this as base, would save me even more monkey work after few months ;)

codesnake:
@stefansaraev from all sdcards I have only one works fairly well, so now I always install OE into NAND.

codesnake:
@stefansaraev I always prefer to keep the commit history rather to rebase every time new kernel version comes out. This way I can see how code change in time and this really helps me sometimes. At all I don't understand why you are re-doing the work already done in my repository and do as you said "monkey work"?

stefansaraev:
cherry picking upstream commits looks messy to me ;)
however.

codesnake:
@stefansaraev But as you can see here https://github.com/codesnake/linux I'm not cherry-picking. linux-stable/linux-3.10.y was taken as a base. In fact you did this, and I already took your work as a base. You can even see your commit in my linux repo: https://github.com/codesnake/linux/commits/amlogic-3.10.y?author=stefansaraev.
Now I'm just merging from upstream, not cherry-picking.

PR #4349, opened 2015-10-01
add new package: packages/sysutils/serdisplib 1.97.9

vitalogy:
Hi,

this request adds support for displays driven by serdisplib through the lcdproc driver glcd.
Header files from serdisplib must be available when glcd is build, else glcd is build without support for serdisplib. 

extend packages/sysutils/lcdproc/package.mk 
If in the file "distributions/OpenELEC/options" SERDISPLIB_SUPPORT is set to "yes", then serdisplib will be a dependency of lcdproc. Furthermore it checks, if the driver glcd is set in LCD_DRIVER, else it will add them automatic to LCD_DRIVER. So an installed serdisplib is nothing without glcd.

add packages/sysutils/serdisplib/package.mk
adds support for serdisplib-1.97.9
Even if SDL is disabled, you must apply the patch "serdisplib-disable_SDL_crosscompile.patch", else configure will fail.

### DIFF
--- a/distributions/OpenELEC/options
+++ b/distributions/OpenELEC/options
@@ -146,7 +146,7 @@
 
 # LCD driver to Use - Possible drivers are ( Comma seperated:
 # bayrad,CFontz,CFontz633,CFontzPacket,curses,CwLnx,dm140,
-# ea65,EyeboxOne,g15,glcdlib,glk,hd44780,i2500vfd,
+# ea65,EyeboxOne,g15,glcd,glcdlib,glk,hd44780,i2500vfd,
 # icp_a106,imon,imonlcd,IOWarrior,irman,irtrans,
 # joy,lb216,lcdm001,lcterm,lirc,lis,MD8800,mdm166a,
 # ms6931,mtc_s16209x,MtxOrb,mx5000,NoritakeVFD,
--- a/packages/sysutils/lcdproc/package.mk
+++ b/packages/sysutils/lcdproc/package.mk
@@ -37,6 +37,16 @@ if [ "$IRSERVER_SUPPORT" = yes ]; then
   PKG_DEPENDS_TARGET="$PKG_DEPENDS_TARGET irserver"
 fi
 
+IFS=$','
+for i in $LCD_DRIVER; do
+  case $i in
+    glcd) PKG_DEPENDS_TARGET="$PKG_DEPENDS_TARGET serdisplib"
+      ;;
+    *)
+  esac
+done
+unset IFS
+
 PKG_CONFIGURE_OPTS_TARGET="--enable-libusb --enable-drivers=$LCD_DRIVER,!curses,!svga --enable-seamless-hbars"
 
 pre_make_target() {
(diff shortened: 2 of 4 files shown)


vitalogy:
Openelec (master gc5f82be) in action with an Alphacool 240x128
img_20151001_121134

Openelec Image for imx6 is available her:
http://milaw.biz/files/OpenELEC-imx6.arm-6.0-devel-20150930235801-r21349-gc5f82be.img.gz

vpeter4:
And which display is this you are using?

vitalogy:
I use an Alphacool 240x128,. This was installed in my x86-HTPC, but I use only my little arm-box now.
So had to buy a stand. Display and the separate stand are around 80-90€ together.
http://www.alphacool.com/index.php/cat/c35_Display.html/
Atelco (Germany) has some in there store.

All supported displays through serdisplib are shown on the homepage of serdisplib.
http://serdisplib.sourceforge.net/#displays

vitalogy:
hmm ... 
I had trouble to build serdisplib, something was going wrong with the SDL part at configure. So I contacted the maintainer of serdisplib and he helped with this patch.
The hole thing thing was writen a few month ago for OpenElec-5, but applied also for the latest git version ... will check this again.

vpeter4:
Patch is fine - tested by me too.
git log is using a and b folder. But running manual diff on two different folders uses the folder name.

What I'm looking is some cheap, big and usb based display. And easily used under openelec.

scripts/image
```diff
@@ -153,6 +153,9 @@ fi
 # Remote support
   [ "$REMOTE_SUPPORT" = "yes" ] && $SCRIPTS/install remote
 
+# Serdisplib support
+  [ "$SERDISPLIB_SUPPORT" = "yes" ] && $SCRIPTS/install serdisplib
```
lrusak: This isn't needed if you have it listed as a dependency in the lcdproc package already.

vitalogy:
But this would also build serdisplib if you set glcdlib in LCD_DRIVER.
Or is there a way to check for the unique string "glcd" in LCD_DRIVER? I doesn't have found one.

btw: For glcdlib you need a separate package graphlcd to work, but have not found this in tree. So should this also be handled like curses and svga in lcdproc/package.mk?

    PKG_CONFIGURE_OPTS_TARGET="--enable-libusb --enable-drivers=$LCD_DRIVER,!curses,!svga --enable-seamless-hbars"

vitalogy:
I will be back in two days and hopefully finish this.
So should we set glcd default in LCD_DRIVER, or optional?
And then check for glcd in lcdproc.mk to build serdisplib, when glcd is set!

scripts/image
vitalogy: yes, will remove this

vitalogy:
reworked and rebased
please have a look

regards
MichaeL

lrusak:
I just build tested this and it works fine. Everything is where it needs to be :+1: 

@sraue it's up to you

PR #4352, opened 2015-10-02
[download serivce] fix critical security bug - broken https

qods:
--passive-ftp is default for years. It does not need to be set.
--no-check-certificate - WTF?!  This is been added by sraue in 2010. This breaks everything HTTPS stands for. It enables the chance of MITM attacks and so on. Just like without any HTTPS.
Note written from the wget developers: It is almost always a bad idea not to check the certificates when transmitting confidential or important data. 
-c the "-c" can cause broken files on some servers. Its much better to reload the file.

This change have been tested. I have redownload all packages. There is also no broken certificate that would require no-check-certificate. If there is anytime any, then the download server have to be changed. Dont add no-check-certificate.

### DIFF
--- a/scripts/get
+++ b/scripts/get
@@ -33,7 +33,7 @@ if [ -n "$PKG_URL" ]; then
     PACKAGE="$SOURCES/$1/$SOURCE_NAME"
     PACKAGE_MIRROR="$DISTRO_MIRROR/$PKG_NAME/$SOURCE_NAME"
     [ "$VERBOSE" != "yes" ] && WGET_OPT=-q
-    WGET_CMD="wget --timeout=30 --passive-ftp --no-check-certificate -c $WGET_OPT -P $SOURCES/$1"
+    WGET_CMD="wget --timeout=30 $WGET_OPT -P $SOURCES/$1"
 
     NBWGET="1"
 


stefansaraev:
why did you remove "-c" ?!

qods:
-c as described above
wget developer say "Wget has no way of verifying that the local file is really a valid prefix of the remote file." This is causing often broken files.

Something else wget developer say:
"Another instance where you’ll get a garbled file if you try to use ‘-c’ is if you have a lame HTTP proxy that inserts a “transfer interrupted” string into the local file."

It can also cause broken files when the first download server broke and it try to confinue to download the file from the own openelec servers.

-c is not really usefull any more these days. Its much better to just make a new clean download. Its also a special case of security issue because wget did not check if the file is valid.

stefansaraev:
if you have a lame http proxy, you should fix it. but well. okay.

stefansaraev:
I have no idea what you mean by "reaload the file" but did you actualy test your change? 
you better do NOT touch -c.

qods:
Yes, i have tested the change. I wrote in the first message that i have tested it.
With "reload the file" i mean reload from the beginning. Not messing around with any existing parts of files already existing on the disk drive because of broken connections.
Could you explain the reason why NOT to touch -c ? I explained why to remove that. If it were usefull in most cases, it would be default in wget. Like --passive-ftp is. But its not usefull in the usecase in openelec.

stefansaraev:
it is broken. your PR. do not touch -c. I am out of this conversation.

qods:
I have tested this PR now few times. I never had problems. Here is been always told that something were broken. But not what is been broken and also this cant be reproduced by me. I have spend some hours in testing.
Could you all please add this PR to your build machines and tell me if something cant be download any more because of -c (that wont make any sense)? I dont know what to test here any more to find out what stefansaraev mean with "it is broken".

qods:
@lrusak @MilhouseVH 
Could you please test this change. As i remember right, you both are building openelec often from source.
The change is working perfectly fine for weeks here.

stefansaraev:
here I tested it. it is broken. as I said - remove the '-c' change!
http://sprunge.us/QWhY

test your own PRs and stop wasting other people's time. thanks.

chewitt:
I'm closing this because we have no plans to change this function of the build system. Thanks.

PR #4355, opened 2015-10-03
Revert "update to 0.2.5598, update link, fix licence"

stefansaraev:
This reverts commit e3e071547d327a961ae6cbe2b9b6ff9c93e9a2bb.

openmpt is now broken. I require a revert. ref #4319

PR #4357, opened 2015-10-03
[libopenmpt] update to recent version, fix broken link

qods:


stefansaraev:
nope

qods:
I wrote as answer to stefansaraev in https://github.com/OpenELEC/OpenELEC.tv/pull/4355 :8ball: 

They changed the download server file in the last 1-2 days. I will update it once again in few minutes.
Here is the recent link that should work for a long time and that i would add in few minutes to a recent change.
http://lib.openmpt.org/files/libopenmpt/src/libopenmpt-0.2.5602-beta14.tar.gz

Why did you want to revert to a older beta version with more bugs and a wrong licence?


@vpeter4 told:


qods, don't update packages without good reason. We already have enough problems and this doesn't help.

Also you don't need to send new PR - just update old one.


The reason of updating the version and not just the PR is because the developers of libopenmpt have fully removed the file from their servers. I cant update to a working link because there is none. They changed the kind of their servers. They are making now "releases" they keep and name those releases xxxx-betaX . Thats why i update the openelec version to such a "release" that wont get removed from their servers so fast any more (and that fix some bugs).

stefansaraev:
still nope ;)

PR #4358, opened 2015-10-03
[RBP] Add new kernel config options

MilhouseVH:
Adds support for SDIO and RPi official touchscreen, as well as other touchscreen hardware.


stefansaraev:
-1 on touchscreen :)

fritsch:
Yeah - we never wanted to support touchscreens in OE ... but now - we got a maintainer :-)

popcornmix:
Thanks @MilhouseVH.
+1 from me

fritsch:
Yeah - as said yesterday. The biggest user group OE has by far far far is the Raspberry Pi - and if this is a new gadget only 10% of those users will use than this is still more than ATV, IMX and wetek and a large part of x86_64 together.

So - it's a nobrainer - OE as a PI distribution - needs support for it.

PR #4362, opened 2015-10-06
libcec: update version to 3.0.1, update to secure links, fix #4294

### LINKED ISSUE #4294: Openelec 5.95.3 - CEC loose connection with LG TV - (libcec 3.0.1 seems to solve the issue)

qods:
Thanks lrusak for showing your changeset https://github.com/lrusak/OpenELEC.tv/commit/b2859b6b594adfef99d46b4ce44b5f057b02fc8f

### DIFF
--- a/packages/devel/libcec/package.mk
+++ b/packages/devel/libcec/package.mk
@@ -17,12 +17,12 @@
 ################################################################################
 
 PKG_NAME="libcec"
-PKG_VERSION="3.0.0"
+PKG_VERSION="3.0.1"
 PKG_REV="1"
 PKG_ARCH="any"
-PKG_LICENSE="GPL"
+PKG_LICENSE="GPLv2"
 PKG_SITE="http://libcec.pulse-eight.com/"
-PKG_URL="http://mirrors.xbmc.org/build-deps/sources/$PKG_NAME-$PKG_VERSION-6.tar.gz"
+PKG_URL="https://github.com/Pulse-Eight/libcec/archive/$PKG_NAME-$PKG_VERSION.tar.gz"
 PKG_DEPENDS_TARGET="toolchain systemd lockdev platform"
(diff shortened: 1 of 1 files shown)


stefansaraev:
as you continue to ignore me, see here https://github.com/Pulse-Eight/libcec/issues/132

vpeter4:
qods, why do you push some other people commits? I think this person is capable to send PR himself.

lrusak:
@qods Thank you for your enthusiasm but unfortunately libcec 3.0.1 is broken, we need to wait for 3.0.2.

qods:
@vpeter4 
Because @lrusak told so here: https://github.com/OpenELEC/OpenELEC.tv/pull/4319

@lrusak @stefansaraev 
Okay. I made this PR because the segfault bug https://github.com/Pulse-Eight/libcec/issues/132 is about ubuntu not openelec. Here its been tested and working on openelec: https://github.com/OpenELEC/OpenELEC.tv/issues/4294
Its written: "So I installed an Openelec testbuild with libcec 3.0.1 and all was working ok"

But ok, i will make a new PR when something newer then 3.0.1 is been released.

PR #4365, opened 2015-10-06
[gcc] update to 5.2.0

lrusak:
This PR updates gcc to version 5.2.0 and fixes a few build issues.

The configure switch ```--with-default-libstdcxx-abi=gcc4-compatible``` was taken from arch linux's build
See: https://projects.archlinux.org/svntogit/packages.git/tree/gcc/trunk/PKGBUILD#n81

systemd fails to run with specific compiler optimizations, turning off these optimizations in CFLAGS fixes this.
See: https://bugzilla.yoctoproject.org/show_bug.cgi?id=8291

arm kernel fails to boot with specific compiler optimizations, turning off these optimizations fixes this. This fix is included as three seperate patches for three different kernel versions (imx6, amlogic, and mainline).
See: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=65932

I have only tested this with the RPi2 build (as that is all I have), I would appreciate if people could test for imx6 and amlogic.

### DIFF
--- a/packages/lang/gcc/package.mk
+++ b/packages/lang/gcc/package.mk
@@ -17,7 +17,7 @@
 ################################################################################
 
 PKG_NAME="gcc"
-PKG_VERSION="4.9.3"
+PKG_VERSION="5.2.0"
 PKG_REV="1"
 PKG_ARCH="any"
 PKG_LICENSE="GPL"
@@ -69,7 +69,8 @@ BOOTSTRAP_CONFIGURE_OPTS="--host=$HOST_NAME \
                           --disable-decimal-float \
                           $GCC_OPTS \
                           --disable-nls \
-                          --enable-checking=release"
+                          --enable-checking=release \
+                          --with-default-libstdcxx-abi=gcc4-compatible"
 
 PKG_CONFIGURE_OPTS_HOST="--target=$TARGET_NAME \
                          --with-sysroot=$SYSROOT_PREFIX \
@@ -106,7 +107,8 @@ PKG_CONFIGURE_OPTS_HOST="--target=$TARGET_NAME \
                          --enable-clocale=gnu \
                          $GCC_OPTS \
                          --disable-nls \
-                         --enable-checking=release"
+                         --enable-checking=release \
+                         --with-default-libstdcxx-abi=gcc4-compatible"
 
 pre_configure_bootstrap() {
   setup_toolchain host
--- a/packages/linux/patches/4.1.10/linux-004-fix-build-with-gcc-5.patch
+++ b/packages/linux/patches/4.1.10/linux-004-fix-build-with-gcc-5.patch
@@ -1,0 +1,16 @@
+# see https://github.com/wongsyrone/openwrt-1/commit/93c0a5173414cfa7684547de3c3a1f3dc4240383
+# and https://gcc.gnu.org/bugzilla/show_bug.cgi?id=65932
+
+diff -Naur linux-4.1.10.orig/arch/arm/Makefile linux-4.1.10/arch/arm/Makefile
+--- linux-4.1.10.orig/arch/arm/Makefile	2015-10-03 04:49:38.000000000 -0700
++++ linux-4.1.10/arch/arm/Makefile	2015-10-05 13:43:26.066966475 -0700
+@@ -118,7 +118,8 @@
+ endif
+ 
+ # Need -Uarm for gcc < 3.x
+-KBUILD_CFLAGS	+=$(CFLAGS_ABI) $(CFLAGS_ISA) $(arch-y) $(tune-y) $(call cc-option,-mshort-load-bytes,$(call cc-option,-malignment-traps,)) -msoft-float -Uarm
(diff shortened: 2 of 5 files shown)


fritsch:
The biggest "No" that I can possibly say. No GCC update before at least the first, much better the second point release is out.

lrusak:
@fritsch please note that this is to merge into openelec-next branch. By no means should this be merged into master.

Simply providing this PR points out a few gotchas with gcc 5.2 and it is good to be aware of them.

fritsch:
Yes - I know that. But consider: Everything that you build with a non
stable / flawed compiler is not debuggable anymore. The compiler is the
most crucial part we have. You cannot trust your whole testsystem anymore
if that one is not known as good working.

If you remember the "flac" nightmare we had some years ago ... I would be
very, very conservative with compiler bumps.

2015-10-06 21:32 GMT+02:00 Lukas Rusak <notifications@github.com>:

> @fritsch <https://github.com/fritsch> please note that this is to merge
> into openelec-next branch. By no means should this be merged into master.
>
> Simply providing this PR points out a few gotchas with gcc 5.2 and it is
> good to be aware of them.
>
> —
> <https://github.com/OpenELEC/OpenELEC.tv/pull/4365#issuecomment-145973439>
> .
>



-- 
                   Key-ID:     0x1A995A9B
                   keyserver: pgp.mit.edu

popcornmix:
See: http://forum.kodi.tv/showthread.php?tid=240614&pid=2120469
"gcc 5.2 is know to cause issues"

fritsch:
It does not help ... I tell that since years ... ever and ever again - but the stupid bumping game always continues :-(

lrusak:
Again, be aware this is in openelec-next branch, which is aimed at for pure testing purposes, it is in no way considered a stable branch.

When things are considered stable, there will be a team discussion when we look to merge into master (this won't be anytime in the foreseeable future, hopefully 6 months or more down the road).

stefansaraev:
was there a team discussion to bump glibc? just curious..

fritsch:
If a version is known broken as spointed out by @popcornmix - there is no reason at all to use it and to test it ... cause one already knows (!) it is broken and causes unpredictable issues. The -next branch will be totally useless until gcc 5.2.x fixes the named issues.

stefansaraev:
I wait for someone th compile some 3rdparty stuff (addons, official, unofficial, whatever) against -next. and 50k users start using the sh**. it will be fun.

MilhouseVH:
`pvr.argustv`, `pvr.mediaportal.tvserver` and `pvr.nextpvr` all fail to compile for RPi and RPi2 with gcc 5.2.0, yet build on master with gcc 4.9.3.

The error is the same for all of the failing addons, in `uri.cpp`, so evidently common/shared code:
```
error: narrowing conversion of '(uri::char_class_e)-2' from 'int' to 'char' inside { } [-Wnarrowing]
```

##### Build logs:
`pvr.argustv`: http://pastebin.com/raw.php?i=XJjKTdfz
`pvr.mediaportal.tvserver`: http://pastebin.com/raw.php?i=kxdB9SUU
`pvr.nextpvr`: http://pastebin.com/raw.php?i=TDduAcXw

All other PVR addons build OK, so maybe a tweak to the compiler flags of the failing addons is needed somewhere? Or better still, a code fix for `uri.cpp` so that it doesn't generate this error with gcc 5.2.0.

MilhouseVH:
Other than the three PVR failures in the previous comment, the RPi and RPi2 builds completed successfully. Initial testing of an [RPi2 build](http://milhouse.openelec.tv/builds/next/) is going well - I've scraped a new MySQL video library (movies and tvshows) over nfs:// without any problems, and videos are playing normally over nfs:// and smb://.

lrusak:
@MilhouseVH thanks for testing and pointing this out.

I have added some workarounds in my tree, this isn't a proper solution and should be changed upstream.
lrusak/OpenELEC.tv@3604042f20d1cecf2d569692bb49b7690e170c92
lrusak/OpenELEC.tv@00b06dd4343ef7e6c8a0504e4d4b2067da64c4f9
lrusak/OpenELEC.tv@75bd6647c35f9aae6058e0172caa4461e0a24883

I built the rest of the binary add-ons and they all built

fritsch:
http://fate.ffmpeg.org/report.cgi?time=20151010052205&slot=x86_64-archlinux-gcc-experimental
https://gcc.gnu.org/bugzilla/show_bug.cgi?id=67920

lrusak:
@fritsch not sure what you are referring to here?

fritsch:
Just saying that uptodate gcc produces wrong code with ffmpeg ...

lrusak:
@fritsch it seems to me like those problems pertain to gcc 6.0 and not the gcc 5 branch no?

fritsch:
Jep. I had the best results with sticking to the compiler Ubuntu uses for their current release - as this guarantees major tested compiler ...

Btw. I did not see the reason why we bumped GCC to 5.2 yet - do we need something of gcc 5.2?

PR #4366, opened 2015-10-06
[RULES.md] initial document stating the OpenELEC repository usage rules.

lrusak:
I created this in hopes to gather the OpenELEC Team and other developers about specific issues that have been plaguing OpenELEC in recent years.

Firstly, this is just a first draft, it needs to be discussed, reviewed, and amended.

Secondly, if there are other packages that needed to be added to the ```Major Packages``` section, please let me know.

My Goals:
- To have a simple discussion about this document and come to an agreement about it's contents.
- To bring us closer as a Team and increase communications about what goes on in the OpenELEC tree.
- To avoid unnecessary breakage in the OpenELEC tree.

To see the actual document with markdown please see [this](https://github.com/lrusak/OpenELEC.tv/blob/rules/RULES.md) (note that the links won't work)

### DIFF
--- a/RULES.md
+++ b/RULES.md
@@ -1,0 +1,66 @@
+#### Table of Contents
+
+- [Package Version Bumps](https://github.com/OpenELEC/OpenELEC.tv/blob/master/RULES.md#package-version-bumps)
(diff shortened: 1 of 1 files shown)


fritsch:
Thanks for bringing this in. How about: "Every Change to OpenELEC has to be PRd via github first"?

qods:
Looks good :+1:

PR #4371, opened 2015-10-09
[make] update to 4.1, update licence, update to secure links

qods:
I have tested this change. Works without any problems. Recent 4.1 is a minor bugfix release of version 4 we already use. Its released in 2014 and used in most of the recent linux OS.


lrusak:
seems sane, but we will wait till after OE 6.0 release. Thanks.

PR #4374, opened 2015-10-10
systemd-addon-wrapper: make binary addons executable ...

vpeter4:
... before starting service

### DIFF
--- a/packages/mediacenter/kodi/scripts/systemd-addon-wrapper
+++ b/packages/mediacenter/kodi/scripts/systemd-addon-wrapper
@@ -30,6 +30,7 @@ if [ -f "/storage/.kodi/addons/$1/system.d/$1.service" ] ; then
   else
     # disable = false: setup
     systemctl enable "/storage/.kodi/addons/$1/system.d/$1.service"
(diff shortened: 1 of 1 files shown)


stefansaraev:
https://github.com/OpenELEC/OpenELEC.tv/blob/master/packages/mediacenter/kodi/patches/kodi-100.05-make-binary-addons-executable.patch should be enough, but this PR does not hurt. OK from me.

vpeter4:
Actually patch you mention doesn't help in this case. Seems service is started before files get executable permissions.

lrusak:
:+1: thanks for this vpeter

stefansaraev:
oh, I thought you forgot about the kodi patch. well as said :+1: push it.

vpeter4:
I checked the file xbmc/addons/Addon.cpp and how we set executable permissions there. I think it should be enough to set permission at top of function OnPostInstall() before actions for service addons are executed. Then chmod inside our wrapper would not be needed.
But even with current state nothing is wrong.

lrusak:
@sraue please pick for openelec-6.0

PR #4376, opened 2015-10-12
projects/RPi*: use correct environment for serial console

vpeter4:


projects/RPi2/filesystem/usr/lib/systemd/system/serial-console.service
```diff
@@ -5,7 +5,8 @@ ConditionKernelCommandLine=console
 
 [Service]
 WorkingDirectory=/storage
-ExecStart=/bin/sh -c '. /etc/profile; exec /bin/sh'
+Environment="ENV=/etc/profile"
+ExecStart=/bin/sh -c 'exec /bin/sh'
```
stefansaraev: assuming that this is runtime tested and ENV= works in ash. "ExecStart=/bin/sh" should be enough. no need to re-exec

vpeter4: In this case only shell is started just like before - no sign of environment variables.

stefansaraev: oh. so it's the same "hack". I dont see how this PR improves something then

vpeter4: Improvement is that now serial console shell actually executes /etc/profile.

stefansaraev: exactly as before

vpeter4: Almost correct. /etc/profile is executed. Environment variables are exported and transferred to shell. But if you set some aliases in profile.d folder they are not set in this new shell. That's why I think this PR fixes this shortcoming and calls profile script from actual shell serial console uses.

@haggaie, because this is originally your commit can you comment?

haggaie: I don't have any aliases so I haven't noticed the issue you are describing. A quick check showed that aliases indeed aren't inherited. I also don't see why you need the exec in this case.

vpeter4: So I checked with this and also works:
```
Environment="ENV=/etc/profile"
ExecStart=/bin/sh
```
Can you confirm?

haggaie: Yes, it worked for me too.

vpeter4: PR updated.

PR #4378, opened 2015-10-13
RFC: Disable libaacs, causes GUI hangs on some BD ISOs

MilhouseVH:
When starting playback of some BD ISOs, the Kodi GUI will hang (one core at 100% load, busy dialog spinning, GUI unresponsive), requiring that Kodi be restarted.

This behaviour is confirmed on RPi, RPi2 and x86 (with Kodi 15 and also Kodi 16).

The problem has been traced to the libaacs library. When playing a problematic BD ISO, this is the last entry written to the log:
```
19:31:12 T:139902096132096   DEBUG: LIRC: Update - NEW at 36030:1c 0 KEY_ENTER devinput (KEY_ENTER)
19:31:12 T:139902096132096   DEBUG: OnKey: 11 (0x0b) pressed, action is Select
19:31:12 T:139902096132096   DEBUG: OnPlayMedia /storage/freenas/media/Demos/ISOs/MVC/Drachenzähmen leicht gemacht 3D.iso
19:31:12 T:139902096132096   DEBUG: GetMovieId (/storage/freenas/media/Demos/ISOs/MVC/Drachenzähmen leicht gemacht 3D.iso), query = select idMovie from movie where idFile=5856
19:31:12 T:139902096132096   DEBUG: GetEpisodeId (/storage/freenas/media/Demos/ISOs/MVC/Drachenzähmen leicht gemacht 3D.iso), query = select idEpisode from episode where idFile=5856
19:31:12 T:139902096132096   DEBUG: GetMusicVideoId (/storage/freenas/media/Demos/ISOs/MVC/Drachenzähmen leicht gemacht 3D.iso), query = select idMVideo from musicvideo where idFile=5856
19:31:12 T:139902096132096   DEBUG: CAnnouncementManager - Announcement: OnAdd from xbmc
19:31:12 T:139902096132096   DEBUG: GOT ANNOUNCEMENT, type: 2, from xbmc, message OnAdd
19:31:12 T:139901324654336   DEBUG: SECTION:LoadDLL(libbluray.so.1)
19:31:12 T:139901324654336   DEBUG: Loading: libbluray.so.1
19:31:12 T:139901324654336   DEBUG: CDVDInputStreamBluray::Logger - bluray.c:1380: libbluray version 0.8.1
19:31:12 T:139901324654336   DEBUG: CDVDInputStreamBluray::Logger - bluray.c:1401: BLURAY initialized!
19:31:12 T:139901324654336   DEBUG: CDVDInputStreamBluray - Opening dir udf://%2fstorage%2ffreenas%2fmedia%2fDemos%2fISOs%2fMVC%2fDrachenz%c3%a4hmen%20leicht%20gemacht%203D.iso
19:31:12 T:139901324654336   DEBUG: CDVDInputStreamBluray - Closed dir (0x7f3d3c096370)
19:31:13 T:139901324654336   DEBUG: CDVDInputStreamBluray::Logger - aacs.c:77: AACS/Unit_Key_RO.inf found. Disc seems to be AACS protected.
19:31:13 T:139901324654336   DEBUG: CDVDInputStreamBluray::Logger - aacs.c:98: Using libaacs for AACS
19:31:13 T:139901324654336   DEBUG: CDVDInputStreamBluray::Logger - aacs.c:118: Loading aacs library (0x7f3d3c0ece00)
19:31:13 T:139901324654336   DEBUG: CDVDInputStreamBluray::Logger - aacs.c:133: Loaded libaacs (0x7f3d3c0ece00)
19:31:13 T:139901324654336   DEBUG: CDVDInputStreamBluray::Logger - aacs.c:136: Registering libaacs filesystem handler 0x8e1918 (0x7f3d3c0ece00)
19:31:13 T:139902096132096   DEBUG: ------ Window Init (DialogBusy.xml) ------
19:31:13 T:139902096132096   DEBUG: LIRC: Update - NEW at 36712:1c 0 KEY_ENTER_UP devinput (KEY_ENTER_UP)
```
Full Kodi 16 log (Nvidia_Legacy, with libaacs): http://sprunge.us/fKKM
Full Kodi 15 log (5.95.5 Nvidia_Legacy with libaacs): http://sprunge.us/MiWU

Disabling libaacs will fix the problem and the BD ISOs will play normally, at least on RPi/RPi2. On x86 I haven't been able to play the ISO at all as the "Play with 3d mode" menu isn't working (this is the only menu option that will successfully play the BD ISO on RPi without libaacs).

Building libaacs with all [upstream commits](http://git.videolan.org/?p=libaacs.git) since 0.8.1 (31 at the time of writing, HEAD=c6dc792eba169e34b34f542c32066c04601cf428) does not resolve the problem, so it is either unknown to the libaacs developers or they don't yet have a fix.

This PR proposes that libaacs is disabled for all platforms. Is there any point to libaacs?

Edit: Updated to confirm same "hang" behaviour in 5.59.5/Kodi 15, and added complete debug logs.

### DIFF
--- a/distributions/OpenELEC/options
+++ b/distributions/OpenELEC/options
@@ -47,7 +47,7 @@
 
 # build and install with AACS support
 # (BD decryption support in KODI) (yes / no)
-  BLURAY_AACS_SUPPORT="yes"
+  BLURAY_AACS_SUPPORT="no"
 
 # build and install with DVDCSS support
 # (DVD decryption support in KODI) (yes / no)


lrusak:
Hmm, never really used aacs but isn't it used for decrypting bluray ISO's?

Not sure if this is something we should really disable, but if it is broken then there is no use supporting it.
I have no way to test this at the moment (I don't use ISO's). Hopefully someone else will chime in.

MilhouseVH:
> but isn't it used for decrypting bluray ISO's?

You'd think that, but disabling libaacs still allows BD ISOs to play on the RPi so I'm not entirely sure what the purpose of libaacs is, as it doesn't seem necessary for playback, and may be causing more problems than it solves.

I'd like to be able to say I've confirmed successful BD ISO playback on x86 with libaacs disabled, but my only test BD ISO is a 3D file and having disabled libaacs it then turns out the "Play with 3d mode"[1] menu is entirely missing on x86 (tested Kodi 15 and Kodi 16), so even without libaacs the BD ISO is unplayable on x86. At least OE x86 without libaacs is no longer hanging though, which is what happens when libaacs is enabled.

And this isn't an OpenELEC specific issue either, as Kodi 16 on Ubuntu will also hang due to libaacs when testing my 3D BD ISO test file (any suggestions on how to build Kodi without libaacs?)

So based on a personal sample of one BD ISO file (but also several reports from users in my test build thread) there seems to be more chance of successful BD ISO playback with libaacs disabled than with libaacs enabled.

It would be interesting if others with a wider selection of BD ISOs (both 3D and 2D) could chime in with their feedback.

libaacs has been disabled since 10 Sep in my RPi builds (~200-250 installs per day) and there have been no reports of BD ISO playback problems resulting from this libaacs change. On the contrary, there have been several positive reports of BD ISO playback now working when it didn't before with libaacs enabled.

1. This menu:
screenshot

lrusak:
Well that is compelling to know that you have already disabled this for a while.

I would say that is evidence enough and this should get merged in for openelec-6.0.

As for building kodi without libaacs, it actually depends on libbluray.
see, https://github.com/OpenELEC/OpenELEC.tv/blob/master/packages/multimedia/libbluray/package.mk#L35
and, https://github.com/OpenELEC/OpenELEC.tv/blob/master/packages/mediacenter/kodi/package.mk#L116

MilhouseVH:
Thanks, adding `--disable-libbluray` to `configure` then means that Kodi on Ubuntu can't play the BD ISO at all, it doesn't even try, but at least it no longer hangs Kodi! Presumably there is no easy way to disable libaacs yet retain libbluray support when building Kodi for Ubuntu.

And just to be clear, 3D BD ISO playback on OE x86 is broken for potentially two reasons - if Kodi doesn't hang because of libaacs, then the ISO is unplayable anyway as there is no 3D menu offered to the user (the 3 remaining menu options all fail because the "disk is encrypted" despite the RPi being able to play encrypted BD ISOs without libaacs... it's all a bit of a mess)

What I'm not clear about is whether 2D BD ISO playback on x86 will continue to work without libaacs, as 2D ISOs shouldn't be dependent on the "Play with 3d mode" menu, but does 2D playback then require libaacs (eg. dependent on one of the other 3 menu items which all fail with 3D ISOs on RPi)? Reports from the RPi test thread would suggest libaacs is not required for either 2D or 3D...

By the way, I've reproduced this behaviour with Kodi 15 (DVDPlayer) and also Kodi 16 with both DVDPlayer and VideoPlayer, so it's not a regression resulting from the VideoPlayer changes, or anything new in Kodi 16.

popcornmix:
Most ISO files users have ripped will have had encryption removed.
For playback of these files libaacs is not required.
We have at one example ISO file where libaacs stops the file from being played (and hangs kodi).

The question is, does libaacs allow encrypted ISO files, or playback from a physical BluRay disk work?
My searching hasn't found any convincing evidence that it does work.
If it does work, then fair enough, libaacs should stay. If it's broken and it stops playback of some unencrypted files, then it should be dropped.

sraue:
@popcornmix generally libaacs works, so we want to keep this. can you provide me a sample which crashes with libaacs installed per email please?

stefansaraev:
I guess libaacs "generaly" "worked" fine. before it was bumped to 0.8.

MilhouseVH:
I tested a new RPi2/Kodi 16 build with libaacs 0.7.1 (64a70c43c0f9c920a9f88bd8ccf6fab38a5efaab), which is the libaacs version we used until 18 Mar 2015 when it was bumped to 0.8.1, and libaacs 0.7.1 also has the same "hanging" problem when playing the test ISO.

I also tested an old RPi2 build from 10 March 2015 (libaacs 0.7.1, libbluray 0.6.2, Kodi 15.0-alpha2-ccff322), and that hangs in the exact same way.

So this doesn't appear to be a new problem with libaacs (or libbluray), but most likely a libaacs bug that now trips up over certain already decrypted ISOs. The test ISO is a 3D ISO, not sure if that's pertinent.

MilhouseVH:
Note that playback of the problem disc is fixed by the following three commits on libaacs git:

http://git.videolan.org/?p=libaacs.git;a=commit;h=1c0ba6b53759ea57a6146b4ce4d902cfe5c44d3c
http://git.videolan.org/?p=libaacs.git;a=commit;h=7ee3b64639358a35188339f4547f70f7d5c2c9c2
http://git.videolan.org/?p=libaacs.git;a=commit;h=972679a52f6a38f08c0bd0ea5c8a0d4750d6a8e8

Commited by Roland Fischer: http://git.videolan.org/?p=libaacs.git;a=search;s=Roland+Fischer;st=author

lrusak:
So with the latest fixes for libaacs it works ok? I can prepare a patch that adds these upstream commits.

MilhouseVH:
Yes, I've had the patches in my test builds for the last few days.

PR #4379, opened 2015-10-15
imx6: redo system type service to start early

vpeter4:


lrusak:
Looks alright to me.

Maybe @stefansaraev can comment about the systemd units ;)

stefansaraev:
I can only say that this imx specific fan/soc thing abuses /etc/profile (profile.d)

for whatever $SYSTEM_TYPE is needed - it has no place in profile.d. it belongs to the systemd unit (preexec, or separate "setup" service, whatever makes sense)

vpeter4:
Thanks for valuable feedback. I took them into consideration and changed problematic parts.

PR #4380, opened 2015-10-17
scripts/checkdeps: fix gentoo deps 

vitalogy:
+ Fix the dependencies of OpenElec on a gentoo system in `scripts/checkdeps`
On Gentoo the header-files are always installed with the builded libs/exe, there are no dev- or header-packages available, so fix the `files_pkg`.
Also, build gcc with support for c++, if g++ is not available and use virtual/jre for java (the user can install which jre/jdk he/she like).
    `g++  ->   gcc[cxx]`
    `jre  ->   virtual/jre`

+ Remove spaces in `scripts/image`.

### DIFF
--- a/scripts/checkdeps
+++ b/scripts/checkdeps
@@ -44,7 +44,8 @@ case "$DISTRO" in
       ;;
     gentoo)
       deps="$deps g++ mkfontscale mkfontdir bdftopcf xsltproc java"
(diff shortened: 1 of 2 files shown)


lrusak:
looks fine to me :+1:

stefansaraev:
push it.

PR #4382, opened 2015-10-18
xf86-video-nvidia-legacy: Bump driver, add kernel 4.3 support

MilhouseVH:
See [here](http://forum.kodi.tv/showthread.php?tid=238393&pid=2136483#pid2136483) for a test build.

The 4.3 support patch ensures the driver will build with the current 4.3-rc5 kernel.

However to actually use this updated driver with a 4.2+ kernel requires one extra patch (not included in this PR - see [patch](http://nmacleod.com/public/oebuild/patches/01_hack_kern43.txt)) to fix the `mtrr_add`/`mtrr_del` issue - my current solution is a quick & dirty hack re-exporting the symbols, but hopefully someone can suggest a better fix that can be added as a separate PR  (or added to this PR if not already merged) once the move towards a 4.2+ kernel picks up pace.

As it stands, this PR will work with the current 4.1.10 kernel. Nvidia have added macros in the 304.128 driver which render the current 3.19/4.0 support patches obsolete and unnecessary.


stefansaraev:
@fritsch fyi :)

:+1: push the button.

fritsch:
Will check the Hack tomorrow :-)

fritsch:
This commit is fine and not a hack. One could think if one would like to check with has_seq_overflowed and return ! of it. But it is fine. Thanks very much. I will pick both into my current jarvis-egl tree that uses fernetmenta's videoplayer rework, which I consider the future of kodi.

MilhouseVH:
The hack is really the patch required to get the driver working with 4.2+ kernel, and not part of this PR. Currently the hack involves re-exporting `mtrr_add` and `mtrr_del` - there's a link to the patch in the first comment, you'll need this in your jarvis-egl tree.

fritsch:
I don't like it :-) - see: http://lists.openwall.net/linux-kernel/2015/03/21/42

MilhouseVH:
I did say it was a quick & dirty (tm) hack! I'm hoping a better fix for the driver will surface at some point, as we're not alone in this, though until then the hack does at least work...

fritsch:
Use: arch_phys_wc_add and arch_phys_wc_del  instead - same parameters.

MilhouseVH:
The only problem is that the value returned by `arch_phys_wc_add` needs to be passed to `arch_phys_wc_del` (according to the kernel source code) and I don't see an easy way of achieving that (my kernel-foo is seriously limited!)

MilhouseVH:
What about this: https://btrlinux.inria.fr/video-fbdev-nvidia-use-arch_phys_wc_add-and-ioremap_wc/ ?

fritsch:
>Drivers must store the return value to pass to mtrr_del_wc_if_needed, but drivers should not try to interpret that return value.

Can you please link me to the nvidia-legacy function that fails to build (very, very limited internet access here, sorry).

fritsch:
From that code you see that that index is already saved, so can be reused later ... but I cannot see code (of nvidia legacy) that currently fails - so will wait until you link me to that code :-)

fritsch:
Btw. as the "new nvidia driver" for non legacy does not fail - why not have a look how they do it?

MilhouseVH:
The driver will build as it stands (without the hack), but the system won't load the driver due to the missing symbols - you get the following in `dmesg`:
```
[    7.996711] nvidia: Unknown symbol mtrr_del (err 0)
[    7.996885] nvidia: Unknown symbol mtrr_add (err 0)
```

The only places where `mtrr_add` and `mtrr_del` are referenced in the nvidia driver are:

xf86-video-nvidia-legacy-304.128/kernel/os-agp.c: [link](http://sprunge.us/eGgh)
xf86-video-nvidia-legacy-304.128/kernel/os-mtrr.c: [link](http://sprunge.us/YFiK)

MilhouseVH:
> Btw. as the "new nvidia driver" for non legacy does not fail - why not have a look how they do it?

I can't find any reference to `arch_phys_wc_add` or `arch_phys_wc_del` in the new 352.55 driver - so that's their solution! :)

fritsch:
Lol - so they do it completely different then. Yeah - I sadly cannot help (your sprunge links are blocked in this country), connection is too slow for vpn :-)

I'd look around the os-agp.c and os-mtrr.c in the newer driver and see what they actually do in there now.

fritsch:
Finally they loaded - yes the link you gave above is the solution, using the new add and del function and ioremap to get the write combined memory allocated.

fritsch:
Btw. can you post the two files from 355 driver, too?

MilhouseVH:
The new 355 driver doesn't have any os-agp.c or os-mtrr.c files (I guess AGP isn't supported, and obviously neither is mtrr...)

fritsch:
:-) good. 

So then let's start here: http://sprunge.us/YFiK

```
#if defined(CONFIG_MTRR)
   int nv_handle = -1;
    if (mode != NV_MEMORY_WRITECOMBINED)
        return RM_ERROR;

    handle = arch_phys_wc_add(start, size);
    if (nv_handle < 0)
    {
        nv_printf(NV_DBG_ERRORS,
            "NVRM: failed to write-combine %lluM @ 0x%0llx.\n",
            (size >> 20), start);
        return RM_ERROR;
    }

    return RM_OK;
#endif
```

```
#if defined(CONFIG_MTRR)
    arch_phys_wc_del(nv_handle );
    return RM_OK;
#endif
```

That's it for that file. Now you need to find where to replace: pci_iomap with pci_iomap_wc

MilhouseVH:
`pci_iomap` isn't referenced anywhere in 304.128

fritsch:
Okay - then they do something internally, which sucks - for us - so then - let's wait for nvidia release a new legacy-driver - in the meantime, go with the "hack" please ... it just reexports working code - but mark the patch with hack please and if possible.

PR #4383, opened 2015-10-20
Various fixes

lrusak:
Picked from @stefansaraev 

Safe for openelec-6.0

Don't merge yet, I still have some more fixes coming


config/options
```diff
@@ -52,6 +52,11 @@ LINUX_DEPENDS="$PROJECT_DIR/$PROJECT/linux/linux.$TARGET_ARCH.conf $ROOT/package
 # If you have ccache installed, take care that LOCAL_CC don't point to it
   LOCAL_CC=`which gcc`
 
+if [ -z "$LOCAL_CC" ] ; then
+  echo "***** y u no install gcc? *****"
```
MilhouseVH: At the risk of sounding like I don't have a sense of humour, I'm not sure this message is entirely appropriate...

lrusak:
@sraue the revert of libopenmtp is need as the new package version fails to compile. I would get this in openelec-6.0

all others are just minors

PR #4385, opened 2015-10-20
[HACK] Allow nvidia legacy driver to work with 4.2+ kernel

MilhouseVH:
Don't merge until Nvidia_Legacy is switched to the 4.2+ kernel.

Hopefully there will be a better solution for this problem before that happens.

See discussion in #4382 for details.


sraue:
is this still needed?

fritsch:
Yes - without this legacy nvidia won't run.

fritsch:
Erm no: https://github.com/OpenELEC/OpenELEC.tv/commit/57b7916473da751bdbbe4f5bf72cbe84ef2c951b <- it's here.

Issue #4305, opened 2015-09-08
[5.95.x] CEC on WetekPlay does not work. It did in 5.x versions


walterheisenberg:
The CEC on every 5.95.x version of OE is not working. And it is not only me, many ppl reported, that CEC on the WetekPlay does not work with 5.95.x. It worked with the 5.x versions.

examples:
http://wetekforums.com/vb5/forum/wetek-play/linux/openelec/15492-openelec-5-95-5-released-for-wetek-play?p=15624#post15624
http://wetekforums.com/vb5/forum/wetek-play/linux/openelec/15290-cec-function-stopped-working-after-upgrade-to-5-95-4-nand (with dmesg-log)
http://openelec.tv/forum/148-usage/76782-cec-toshiba-regza-remote-not-working-with-6-x
http://openelec.tv/forum/146-wetek-play/77018-cec-support#141674 and below (#11 and #12)
```
echo 0x0f > /sys/class/amhdmitx/amhdmitx0/cec_config
```
or
```
echo 1 > /sys/devices/virtual/amhdmitx/amhdmitx0/cec_config
```
didn't work. Enabling CEC on Android before did also not help.

It worked in 5.x, so please make it work in 6.x versions also
Thx

qods:
You can try out 3.0.1 with this patch here:
https://github.com/OpenELEC/OpenELEC.tv/pull/4362

OpenELEC would bump to anything newer then 3.0.1 when its released to hopefully fix this bug. Until then it would be helpefull if you report if the listed patch fix your problem.

stefansaraev:
@qods you have no idea what you are talking about. please stop now!. wetek is not using libcec.

@walterheisenberg an official OE 5.0.x was never released for wetek play. so, honestly, most of us have no clue "how" it worked before ;)

cc @codesnake

codesnake:
@stefansaraev is right WeTek doesn't use libcec, because libcec doesn't support Amlogic SoCs.
In unofficial 5.0.x builds we had Amlogic CEC implementation enabled by default, but currently we disabled because it is unstable.
However `echo 0x0f > /sys/class/amhdmitx/amhdmitx0/cec_config` should work in most cases. At least it works on my LG TV, but may suddenly stop working.

qods:
@stefansaraev 
Thanks for the information. Because noone have answered to this bugreport for a month i tried helping.

kdietrich:
[...]

PR #4387, opened 2015-10-20
[WeTek_Play] enable cec as default

lrusak:
This can be disabled by using the cmdline argument ```nocec```

There is a problem with the fact that we advertise cec support but it is not enabled by default.

#4305 

ping @stefansaraev @codesnake 

Disclaimer: I don't have a Wetek device so I am unable to test ;)

### DIFF
--- a/projects/WeTek_Play/filesystem/usr/lib/systemd/system/hdmi-cec.service
+++ b/projects/WeTek_Play/filesystem/usr/lib/systemd/system/hdmi-cec.service
@@ -1,0 +1,12 @@
+[Unit]
+Description=hdmi cec service
+DefaultDependencies=no
+ConditionPathExists=!/storage/.no-hdmi-cec
+
+[Service]
+Type=oneshot
+RemainAfterExit=yes
+ExecStart=/bin/sh -c "echo 0x0b > /sys/class/amhdmitx/amhdmitx0/cec_config"
+
+[Install]
+WantedBy=sysinit.target


stefansaraev:
it is not possible to pass boot args without messing up with uboot env (with a serial cable)

codesnake:
Amlogic HDMI driver already supports enabling from kernel command line, just pass `hdmitx=cecf`, no need for script.

lrusak:
Updated to use a systemd service instead.

sraue:
@lfiebach can we add this to our settings addon (if /sys/class/amhdmitx/amhdmitx0/cec_config exist write "0x0f" there, maybe we should do in the way we enable our services via the settings addon)

stefansaraev:
> @lfiebach can we add this to our settings addon (if /sys/class/amhdmitx/amhdmitx0/cec_config ..

no point if you ask me. you dont know how it will work in next amlogic (3.14 or newer) kernel.

kdietrich:
I don't know if it is a good idea to enable CEC that is not fully working. I'm using a Samsung TV, enabled CEC a few days ago and I will disable it again. The reason: Every time I turn off my TV now, the Wetek gets shut down too. So: No PVR recordings anymore. However when I turn my TV on, the Wetek does not boot up, I have to use the Wetek remote.

So the only solution at the moment would be to turn off CEC so that my Wetek remains active and does PVR recording when I turn off my TV. Or can we prevent this default behavior? Otherwise my suggestion: Don't advertise CEC when it is not fully working and - who knows - maybe sometimes libcec will do a proper android build..

stefansaraev:
^ this. was what I meant yesterday, @lrusak

lrusak:
@stefansaraev yep :) let's wait for the next amlogic kernel and go from there.

codesnake:
@kdietrich For MOST of the users HDMI CEC is only the ability to control the underlying device using TV's remote.
@lrusak You can do `echo 0x0b > /sys/class/amhdmitx/amhdmitx0/cec_config`. This wil enable CEC, but will not enable the "one touch standby" function.

lrusak:
So for documentation purposes I'll leave this here.

After looking at the code it looks like there are a few functions
```
/*CEC UI MASK*/
#define CEC_FUNC_MSAK                        0
#define ONE_TOUCH_PLAY_MASK                  1
#define ONE_TOUCH_STANDBY_MASK               2
#define AUTO_POWER_ON_MASK                   3
```
so this becomes a bitwise mask ```3210```

0x0f (1111) turns everything on
0x0b (1011) turns everything on except ONE_TOUCH_STANDBY_MASK
0x03 (0011) turns everything on except ONE_TOUCH_STANDBY_MASK and AUTO_POWER_ON_MASK
...
etc

lrusak:
updated to us ```0x0b```

kdietrich:
^ :+1:

lrusak:
I have packaged this into an add-on instead. see https://github.com/lrusak/wetek-cec

The add-on is available as a zip [here](https://dl.dropboxusercontent.com/u/21742150/service.system.wetek-cec-6.0.0.zip)

jester-xbmc:
if people are looking, its here:
https://github.com/lrusak/lrusak-openelec-addons/tree/master/addons/service/system/wetek-cec

Issue #4390, opened 2015-10-23
Mistake at compilation of OPENELEC in UNUNTU 15.10


ed68:
./scripts/image release

 =================================================================================
 Configuration for OpenELEC (unofficial)
 =================================================================================

 Buildoptions:
 ======================================================
 - CPU (ARCH):                           x86-64 (x86_64)
 - FLOAT:
 - FPU:
 - SIMD support:                         yes
 - Optimizations:                        size
 - LTO (Link Time Optimization) support: yes
 - GOLD (Google Linker) Support:         yes
 - LLVM support:

 Graphic configuration:
 ======================================================
 - Xorg Graphic Drivers:                 nvidia-legacy
 - XORG Composite support:
 - WindowManager:                        fluxbox
 - OpenGL (GLX) support (provider):      yes (mesa)
 - OpenGLES support (provider):          no (no)
 - uvesafb support:                      yes

 Hardware decoder configuration:
 ======================================================
 - Kodi Player driver:                   default
 - VAAPI Support:
 - VDPAU Support:                        yes
 - DCA decoder Support:                  yes

 Input device configuration:
 ======================================================
 - Remote support:                       yes
 - ATV Remote support:                   yes
 - CEC Adapter support:                  yes
 - IRTrans support:                      yes
 - Kodi Joystick support:                yes

 Misc. hardware configuration:
 ======================================================
 - ALSA support:                         yes
 - Bluetooth support:                    yes
 - LCD drivers:                          irtrans,imon,imonlcd,mdm166a,MtxOrb,lis,dm140,hd44780,CFontz,SureElec,vlsys_m428,serialVFD,shuttleVFD
 - Include driver:                       RTL8192CU
 - Include driver:                       RTL8192DU
 - Include driver:                       RTL8188EU
 - Include driver:                       RTL8812AU
 - Include driver:                       dvbhdhomerun
 - Include driver:                       bcm_sta
 - Include firmware:                     misc-firmware
 - Include firmware:                     wlan-firmware
 - Include firmware:                     dvb-firmware
 - Include firmware:                     iwlwifi-firmware

 Network service configuration:
 ======================================================
 - Avahi (Zeroconf) support:             yes
 - NFS mounting support:                 yes
 - SAMBA mounting support:               yes
 - SAMBA server support:                 yes
 - SFTP server support:                  yes

 OS configuration:
 ======================================================
 - OEM Support:                          no
 - Default ROOT Password:                openelec
[...]

PR #4393, opened 2015-10-23
buildsys: do poor man's md5deep

stefansaraev:
this is a fix for #4390

buntu 15.04+ is packaging md5deep differently, and the tool is not available anymore


lrusak:
:+1: for less dependencies. Looks alright to me.

Issue #4375, opened 2015-10-11
scripts/checkdeps does not check for dosfstools


vitalogy:
note: edited the hole thing, after my asininity is gone

I get this error:

    scripts/mkimage: line 109: mkfs.vfat: command not found

I ran into this months ago, so i had manual installed dosfstools and i had to add sudo command before `mkfs.vfat` in `scripts/mkimage`.

From `scripts/mkimage` runs the command `parted` without the addition of sudo, but why i had to add it before `mkfs.vfat`? Should `scripts/checkdeps` not also check if `mkfs.vfat` and `parted` are available?

Using here an up-to-date gentoo (x86_64).

lrusak:
Humm, I just updated the gentoo deps, but perhaps I missed some (I don't use gentoo fulltime) so if we need to add ```dosfstools``` and ```parted``` to the deps then that is ok.
see, https://github.com/OpenELEC/OpenELEC.tv/blob/master/scripts/checkdeps#L47


As for sudo, the entire mkimage script is run as sudo
see https://github.com/OpenELEC/OpenELEC.tv/blob/master/scripts/image#L324

vitalogy:
by running the normal build command `$ PROJECT=imx6 ARCH=arm make image`
i get this

    Parallel mksquashfs: Using 4 processors
    Creating 4.0 filesystem on /home/michael/workspace/OpenELEC.tv/target/OpenELEC-imx6.arm-6.0-devel-20151018001514-r21418-g3f21054.system, block size 131072.
    <snip>
    mkimage: boo. now root access (sudo) is required...
    mkimage: see scripts/image and scripts/mkimage if you dont trust us :)
    image: creating image: /home/michael/workspace/OpenELEC.tv/target/OpenELEC-imx6.arm-6.0-devel-20151018001514-r21418-g3f21054.img...
    292+0 records in
    292+0 records out
    306184192 bytes (306 MB) copied, 0.218713 s, 1.4 GB/s
    image: creating partition table on /home/michael/workspace/OpenELEC.tv/target/OpenELEC-imx6.arm-6.0-devel-20151018001514-r21418-g3f21054.img...
    image: creating part1 on /home/michael/workspace/OpenELEC.tv/target/OpenELEC-imx6.arm-6.0-devel-20151018001514-r21418-g3f21054.img...
[...]

PR #4394, opened 2015-10-23
Build mkfs.(v)fat from dosfstools for the host

vitalogy:
Build dosfstools for the host from `scripts/image` by default, like it is done for squashfs.
Tested the build `dosfstools:host` with a clean build* dir, except toolchain nothing else is needed.
The error i had descriped here #4375 is gone.

### DIFF
--- a/packages/sysutils/dosfstools/package.mk
+++ b/packages/sysutils/dosfstools/package.mk
@@ -40,9 +40,25 @@ make_init() {
   : # reuse make_target()
 }
 
+pre_build_host() {
+  mkdir -p $PKG_BUILD/.$HOST_NAME
+  cp -RP $PKG_BUILD/* $PKG_BUILD/.$HOST_NAME
+}
+
+make_host() {
+  cd $ROOT/$PKG_BUILD/.$HOST_NAME
+  make PREFIX=/usr
+}
+
 makeinstall_init() {
   mkdir -p $INSTALL/sbin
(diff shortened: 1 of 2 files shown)


stefansaraev:
this is not going to work. dosfstool is not built in subdirs for :host and :target so, you risk installing a native tool to rpi/imx6/wetek targets.

the package needs more rework

stefansaraev:
here. something like this. http://sprunge.us/RGbN

quick and dirty, not tested, you better do ;)

packages/sysutils/dosfstools/package.mk
```diff
@@ -23,6 +23,7 @@ PKG_ARCH="any"
 PKG_LICENSE="GPLv3"
 PKG_SITE="https://github.com/dosfstools/dosfstools"
 PKG_URL="https://github.com/dosfstools/dosfstools/releases/download/v$PKG_VERSION/$PKG_NAME-$PKG_VERSION.tar.xz"
+PKG_DEPENDS_HOST="toolchain"
```
stefansaraev: PKG_DEPENDS_HOST="". you can't have toolchain here.

vitalogy: leftover from testing

vitalogy:
yes, you are right ... rebased

stefansaraev:
+1.

Issue #4351, opened 2015-10-02
Fully broken VDR in openelec 5.95.5


odc2:
I cant use VDR in any case i tried. It worked perfectly in 5.0.8. In 5.95.5 it simply cant start. Images are always x86_64.

How i used it in 5.0.8: http://freaktab.com/forum/advertisers-and-vendor-support-area/wetek/24582-tutorial-setting-up-vdr-on-openelec

How it look in 5.95.5:
When i scan for "VDR" i find 3 things:
1. VDR Configuration
2. VDR PVR Backend
3. VDR VNSI Client

I press install on VDR PVR Backend. It installes then all the 3 things (seems to be dependency now). After that i go to TV part and enable it. It tells me that the client is been disabled and opens a page where i can see that VNSI is turned off. I press enable.
After that TV is been enabled in settings. I go now to http://s5.postimg.org/jvmwz4u93/screenshot014.png
I press enter like on 5.0.8, but it never opens http://s5.postimg.org/638i3i3hj/screenshot015.png

I reboot the system. At reboot it every time tells me that VDR is been enabled and i have to enable it (same message like described above). I try again to open the VDR settings. It also never open.
Yes, there is the adapter detected fine:
/dev/dvb/adapter0

kodi log with enabled debuging: https://pastebin.mozilla.org/8848149

This bug can be reproduced on 3 fully different devices with different hardware. It can also be reproduced on a pc without DVB card.

stefansaraev:
this is not a bug. you are supposed to 1. open pvr client's settings 2. save 3. enable it.

odc2:
Thanks for the quick answer.
I went to the "VDR VNSI Client" (version 1.10.10) where you can Uninstall, Disable and Configure it. I choose the "Configure" on the top. It opens the page with "VDR Hostname or IP - 127.0.0.1" and so on. I didnt change anything and press OK. When i try to restart (disable, enable) i get message that it cant disable it because its been used by the VDR PVR Backend. I then disable and enable the VDR PVR Backend.
It should work now, right? It doesnt. I go again to "Client specific" and try to view the VDR settings. They didnt open as described above.

vpeter4:
Seems your VDR PVR Backend doesn't run. That's why you can't get VDR OSD. Configure that one first. And also check in SSH if process vdr.bin is running.
I need to reboot the box once to get OSD visible. Don't know why.

fritsch:
[...]

PR #4396, opened 2015-10-23
Pick some commits for openelec-6.0

lrusak:
A final few fixes and tweaks that should make it into openelec 6.0

Update to kodi 15.2 final

@sraue please review


fritsch:
Push those, too ... we have not a single need to wait for 6.0 anymore with kodi 15.2 is out ... let's not fuck it up in the last minute by bumping some random patches and then press the release button ...

sraue:
please dont push such PRs we still sync master with openelec-6.0 branch, for new stuff we have openelec-next branch

fritsch:
Those are cherry-picks - every "sync" will get them easily ...

odc2:
Could you please help fixing the VDR problem? VDR is the only official way to watch live-TV and i tested on two different machines with different hardware and different DVB (one DVB-S and other DVB-T) and both was not propperly working.

This is the Bug i am talking about: https://github.com/OpenELEC/OpenELEC.tv/issues/4351

PR #4389, opened 2015-10-23
plymouth-lite: black background in distro based splash

vpeter4:

PR #4397, opened 2015-10-24
plymouth-lite: black background in distro based splash

vpeter4:
sending again as replacement for #4389 because of prematurely deleted branch


lrusak:
\o/ thanks much :+1: from me

sraue:
added in https://github.com/OpenELEC/OpenELEC.tv/commit/019589d077bf4a4662d189fd9b4fb520e156402f

PR #4399, opened 2015-10-25
[RBP] Drop mkknlimg, use version shipped with kernel

MilhouseVH:
No need to carry (and maintain) our own mkknlimg script when we can just call the version supplied by the kernel.


MilhouseVH:
Added second commit to avoid confusing `linux-firmware-1234` with `linux-*`.

lrusak:
looks good :+1:

PR #4406, opened 2015-11-02
kodi: add backported patch for xbmc/xbmc#8334

lrusak:
@stefansaraev anything else you can think of?


fritsch:
Pick the other one as well, but please wait a few days data dvds perhaps is getting some love, too :-)

stefansaraev:
hold on. I'll push a fix for data cd/dvds later. and then you can pick the whole PR as is. once if it is approved upstream.

it will still need testing with isengard. I test on jarvis..

stefansaraev:
why the h** you merged this @sraue ?

PR #4414, opened 2015-11-03
kodi: disable pvermanager.syncchannelgroups (fixup)

fritsch:
Please also pick to 6.0 as pointed out by @anaconda


iLLiac4:
I have a feeling that this commit has broke Group in Kodi. Now the only group listed is All channels. Reverting back to previous build brings back Groups. To test that you have to select System/TV/General/Clear data.

PR #4426, opened 2015-11-05
imx6 generate multiple images

xoddark:
This change permit to generate an OpenELEC image for each imx6 system than need it (Cubox and Udoo for now).
The goal is to easily have an image for Wandboard when the support will be integrated.

This change add the possibility to set the wanted system image with the shell var "IMAGES", not sure if it's should be documented and where. If nothing is set it generate all the images.

### DIFF
--- a/scripts/image
+++ b/scripts/image
@@ -310,32 +310,66 @@ fi
           KERNEL_NAME="KERNEL"
         fi
 
-        UBOOT_SYSTEM=""
-        if [ "$PROJECT" == "imx6" ]; then
-          if [ -n "$SYSTEM" ]; then
-            UBOOT_SYSTEM="$SYSTEM"
+        # default target for image
+        if [ "$PROJECT" = "imx6" ]; then
+          if [ -z "$IMAGES" ]; then
+            UBOOT_IMAGES=$IMAGES_SYSTEMS_NEEDED
           else
-            UBOOT_SYSTEM="cuboxi"
+	    for IMAGE_ASKED in $IMAGES; do
+	      SYSTEM_NEED_IMAGE=false
+	      for IMAGE_SYSTEM_NEEDED in $IMAGES_SYSTEMS_NEEDED; do
+		if [ $IMAGE_SYSTEM_NEEDED = $IMAGE_ASKED ]; then
+                  SYSTEM_NEED_IMAGE=true
+		fi
+              done
+              if [ $SYSTEM_NEED_IMAGE == true ]; then
+                UBOOT_IMAGES+=( $SYSTEM_ASKED )
+              else
+                echo "mkimage: '$SYSTEM_ASKED' is not a valid DEVICE name for build image."
+              fi
+            done
           fi
+        else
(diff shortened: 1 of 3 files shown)


scripts/image
```diff
@@ -310,32 +310,66 @@ fi
...
+              done
+              if [ $SYSTEM_NEED_IMAGE == true ]; then
+                UBOOT_IMAGES+=( $SYSTEM_ASKED )
+              else
+                echo "mkimage: '$SYSTEM_ASKED' is not a valid DEVICE name for build image."
+              fi
+            done
           fi
+        else
+          UBOOT_IMAGES=( "undef" )
```
MilhouseVH: This "undef" is rather ugly.

Also, using the variable name `UBOOT_IMAGES` for what may not be a uboot-based system (ie. RPi, x86 etc.) seems inappropriate. A more generic naming solution would be better, and the undef should be eliminated.

scripts/image
```diff
@@ -310,32 +310,66 @@ fi
...
+        do
+          echo "mkimage: boo. now root access (sudo) is required..."
+          echo "mkimage: see scripts/image and scripts/mkimage if you dont trust us :)"
+
+#          IMAGE_NAME="$DISTRONAME-$TARGET_VERSION"
+#IMAGE_NAME="$DISTRONAME-$PROJECT.$TARGET_ARCH-$OS_VERSION-$OPENELEC_VERSION"
+          if [ "$UBOOT_IMAGE" = "undef" ]; then
+            IMAGE_OUT_NAME="$IMAGE_NAME"
+          else
+            IMAGE_OUT_NAME="$DISTRONAME-$PROJECT-$UBOOT_IMAGE.$TARGET_ARCH-$OS_VERSION-$OPENELEC_VERSION"
```
MilhouseVH: If the purpose of the UBOOT_IMAGES array is to hold a list of images to be built, why not just populate the array with the final image output name which avoids the ugly "undef" in line #333?

Iterating over `UBOOT_IMAGES` for non-uboot systems seems all kind of wrong...

vpeter4:
For now you can build image like
```
PROJECT=imx6 ARCH=arm SYSTEM=wandboard make image
```
Just add everything for wandboard just like I did for udoo and matrix.

I don't think there will be official support for other systems soon.

xoddark:
Hello, sorry for the long time without new.
@MilhouseVH: Thank you for the review, I will try to improve my change.
@vpeter4 : Wandboard use upstream uboot, I think the best should to use upstream u-boot for cubox too.

vpeter4:
Upstream u-boot doesn't cover all solidrun's devices. Which means it is no-go.
That's why I'm using two of them on my private builds (u-boot-imx6-e817fa3 and u-boot-2015.10-rc5). Second one is used for Udoo and TBS Matrix and it is build after first one.

chewitt:
I'm closing this PR down as after some internal discussion it appears there are no active plans for adding Udoo, Matrix or Wandboard as official projects. Thanks.

vpeter4:
Maybe one day not in near future.

PR #4427, opened 2015-11-06
pulseaudio and kodi

lrusak:
EDIT: Updated 2015-11-07, see included update.

``` 
kodi-100.15-allow-pulse-and-alsa-together.patch
```
~~Allows enumerating both pulseaudio and alsa devices in Kodi.~~

~~I am aware that this could be problematic, but it is something we need otherwise pulseaudio won't be used. I have see xbmc/xbmc#5908 and reading over some of the negative effects I don't think they will effect us on OpenELEC.~~

``` "We will plain segfault when we open alsa devices while PA server is initializing." ```
systemd solves this for us and makes sure that pulseaudio is started before kodi.

``` "PulseAudio might hog your Audio device and you cannot open the ALSA device exclusively" ```
I haven't seen this yet but have only done limited testing

ping @fritsch 

----------

```
kodi-100.16-allow-pulseaudio-on-the-rpi.patch
```
Not sure why this isn't used by default but maybe there is something I'm missing.
Here is a log from my RPi2 with this patch http://sprunge.us/WKfJ

ping @popcornmix

----------
The other commit is to remove the pulseaudio module that creates a null sink (if no audio devices are detected)

----------
Update 2015-11-07:

I have made some changes to this PR since my first message and I will edit it to explain what this PR currently does.

This PR, with the changes to our OE settings add-on allows the user to use exclusively pulseaudio or alsa. So with pulseaudio active the following is possible:

 - On x86_64 this can be a bluetooth audio device or any other hardware audio device.
 - On the RPi this can only be a bluetooth device (we do not load snd_bcm2835 by default so no alsa devices are enumerated). This may change at some point.

### DIFF
--- a/packages/audio/pulseaudio/config/system.pa
+++ b/packages/audio/pulseaudio/config/system.pa
@@ -55,7 +55,6 @@ load-module module-suspend-on-idle
 ### Enable positioned event sounds
 load-module module-position-event-sounds
 
-
 ### Automatically load modules for dbus
 .ifexists module-dbus-protocol.so
  load-module module-dbus-protocol
@@ -72,10 +71,6 @@ load-module module-position-event-sounds
 
 .ifexists module-zeroconf-publish.so
   load-module module-zeroconf-publish
-.endif
-
-.ifexists module-zeroconf-discover.so
-  load-module module-zeroconf-discover
 .endif
 
 load-module module-native-protocol-tcp auth-anonymous=1
--- a/packages/audio/pulseaudio/package.mk
+++ b/packages/audio/pulseaudio/package.mk
@@ -110,7 +110,3 @@ post_makeinstall_target() {
   cp $PKG_DIR/config/system.pa $INSTALL/etc/pulse/
   cp $PKG_DIR/config/pulseaudio-system.conf $INSTALL/etc/dbus-1/system.d/
 }
-
-post_install() {
-  enable_service pulseaudio.service
-}
--- a/packages/audio/pulseaudio/system.d/pulseaudio.service
+++ b/packages/audio/pulseaudio/system.d/pulseaudio.service
@@ -1,10 +1,15 @@
 [Unit]
 Description=PulseAudio Sound System
 After=syslog.target local-fs.target
+ConditionPathExists=/storage/.cache/services/pulse.conf
 
 [Service]
+Restart=on-failure
 ExecStart=/usr/bin/pulseaudio --system
-Restart=always
+TimeoutStopSec=1s
+RestartSec=2
+StartLimitInterval=60
+StartLimitBurst=10
 
 [Install]
 WantedBy=multi-user.target
--- a/packages/mediacenter/kodi/patches/kodi-100.16-allow-pulseaudio-on-the-rpi.patch
+++ b/packages/mediacenter/kodi/patches/kodi-100.16-allow-pulseaudio-on-the-rpi.patch
@@ -1,0 +1,48 @@
+diff -Naur a/xbmc/cores/AudioEngine/AESinkFactory.cpp b/xbmc/cores/AudioEngine/AESinkFactory.cpp
+--- a/xbmc/cores/AudioEngine/AESinkFactory.cpp	2015-11-02 01:54:40.000000000 -0800
++++ b/xbmc/cores/AudioEngine/AESinkFactory.cpp	2015-11-05 15:00:36.874646180 -0800
+@@ -28,6 +28,9 @@
+ #elif defined(TARGET_RASPBERRY_PI)
+   #include "Sinks/AESinkPi.h"
+   #include "Sinks/AESinkALSA.h"
++  #if defined(HAS_PULSEAUDIO)
++    #include "Sinks/AESinkPULSE.h"
++  #endif
+ #elif defined(TARGET_DARWIN_IOS)
+   #include "Sinks/AESinkDARWINIOS.h"
+ #elif defined(TARGET_DARWIN_OSX)
+@@ -67,6 +70,9 @@
+ #elif defined(TARGET_RASPBERRY_PI)
+         driver == "PI"          ||
+         driver == "ALSA"        ||
++  #if defined(HAS_PULSEAUDIO)
++        driver == "PULSE"       ||
++  #endif
+ #elif defined(TARGET_DARWIN_IOS)
+         driver == "DARWINIOS"  ||
+ #elif defined(TARGET_DARWIN_OSX)
+@@ -112,6 +118,10 @@
+   if (driver == "ALSA")
+     sink = new CAESinkALSA();
+   #endif
++  #if defined(HAS_PULSEAUDIO)
++  if (driver == "PULSE")
++    sink = new CAESinkPULSE();
++  #endif
+ #elif defined(TARGET_DARWIN_IOS)
+     sink = new CAESinkDARWINIOS();
+ #elif defined(TARGET_DARWIN_OSX)
+@@ -202,6 +212,13 @@
+   CAESinkPi::EnumerateDevicesEx(info.m_deviceInfoList, force);
+   if(!info.m_deviceInfoList.empty())
+     list.push_back(info);
++  #if defined(HAS_PULSEAUDIO)
++  info.m_deviceInfoList.clear();
++  info.m_sinkName = "PULSE";
++  CAESinkPULSE::EnumerateDevicesEx(info.m_deviceInfoList, force);
++  if(!info.m_deviceInfoList.empty())
++    list.push_back(info);
++  #endif
+   #if defined(HAS_ALSA)
+   info.m_deviceInfoList.clear();
+   info.m_sinkName = "ALSA";

### CODE REVIEW

**Comment by fritsch** [comment]
That won't work as you expect it ... as the the default device will be routed through pulse automatically, see aplay -L. Try the following: Play something via the "Default" device (it has pulseaudio something in it's name but is AESinkALSA ... then switch over to AESinkALSA exclusive device - and see it fail.

Debuglog welcome.

**Comment by lrusak** [comment]
I see what you mean, http://sprunge.us/NPUF
```
09:24:03 T:140031271565056   DEBUG: CActiveAESink::OpenSink - trying to open device ALSA:@:CARD=PCH,DEV=0
09:24:03 T:140031271565056    INFO: CAESinkALSA::Initialize - Attempting to open device "@:CARD=PCH,DEV=0"
09:24:03 T:140031271565056    INFO: CAESinkALSA - Unable to open device "sysdefault:CARD=PCH" for playback
09:24:03 T:140031271565056    INFO: CAESinkALSA - Unable to open device "default:CARD=PCH" for playback
09:24:03 T:140031271565056    INFO: CAESinkALSA - Unable to open device "front:CARD=PCH,DEV=0" for playback
09:24:03 T:140031271565056   ERROR: CAESinkALSA::Initialize - failed to initialize device "@:CARD=PCH,DEV=0"
09:24:03 T:140031271565056   DEBUG: CActiveAESink::OpenSink - trying to open device PULSE:Default
```

So there is no easy solution it seems.
1. [easy] enable/disable pulseaudio from within OE and restart kodi (unless re-enumeration is possible)
2. [hard] create some sort of pavucontrol add-on in kodi to be able to adjust pulseaudio settings.

**Comment by fritsch** [comment]
What do you want to use Pulseaudio for? Is it all about Steam? Is ist for bluetooth output devices?

**Comment by lrusak** [comment]
For use with bluetooth audio mainly, but it would be nice to support options such as pulseaudio network sound server and audio streaming from bluetooth devices like phones. I have all this working but being able to select between audio outputs has been the issue.

I will look into what you are suggesting. Thanks.

**Comment by fritsch** [comment]
That's easy - just disable it completely for local devices, see my last post about this topic.

**Comment by lrusak** [comment]
We want the ability to use it for local devices so OE can be use to send and receive network sound.

**Comment by fritsch** [comment]
Nope - you don't need this. The same I told @samnazarko. Implement an AESource device, that use PA to receive the data and then put it into kodi via the AE interface.

**Comment by lrusak** [comment]
Updated, made it possible to enable/disable pulseaudio via the OE settings add-on
see, lrusak/service.openelec.settings@c486dbf4b81e512522d7eae6a940f66b0f892d6f

**Comment by MilhouseVH** [comment]
Great work! This is working very well with a RPi2, cheapo BlueNext BN01 Bluetooth dongle (based on CSR chipset) and SonyEricsson HBH-DS970 BT headphones (circa 2006). Listening to a DTS-HDMA audio track on this setup is a little crazy, but it works perfectly! (Well almost, see below - couple of minor issues) :)

I'll include this in my next test build.

I have noticed a couple of problems so far, neither terribly major:

1) If the Bluetooth headphones are powered off during playback, Kodi will crash.

`journalctl` from the point where the BT device is powered off:
```
Nov 07 06:21:11 rpi22 dbus[234]: [system] Rejected send message, 3 matched rules; type="method_return", sender=":1.3" (uid=0 pid=263 comm="/usr/bin/pulseaudio --system ") interface="(unset)" member="(unset)" error name="(unset)" requested_reply="0" destination=":1.6" (uid=0 pid=392 comm="/usr/lib/bluetooth/bluetoothd ")
Nov 07 06:21:11 rpi22 systemd-udevd[708]: Process '/usr/bin/ir-keytable --delay=1000 --device=/dev/input/event4' failed with exit code 255.
Nov 07 06:21:28 rpi22 kodi.sh[578]: Segmentation fault (core dumped)
Nov 07 06:22:29 rpi22 kodi.sh[578]: Crash report available at /storage/.kodi/temp/kodi_crashlog_20151107062128.log
Nov 07 06:22:29 rpi22 systemd[1]: kodi.service: Main process exited, code=exited, status=139/n/a
Nov 07 06:22:29 rpi22 kill[730]: kill: you need to specify whom to kill
Nov 07 06:22:29 rpi22 systemd[1]: kodi.service: Control process exited, code=exited status=1
Nov 07 06:22:29 rpi22 systemd[1]: kodi.service: Unit entered failed state.
Nov 07 06:22:29 rpi22 systemd[1]: kodi.service: Failed with result 'exit-code'.
Nov 07 06:22:32 rpi22 systemd[1]: kodi.service: Service hold-off time over, scheduling restart.
Nov 07 06:22:32 rpi22 systemd[1]: Stopped Kodi Media Center.
Nov 07 06:22:32 rpi22 systemd[1]: Started Kodi splash animation.
Nov 07 06:22:32 rpi22 systemd[1]: Starting Kodi Media Center...
Nov 07 06:22:32 rpi22 systemd[1]: Started Kodi Media Center.
```
Debug-enabled crashlog: http://sprunge.us/jFji

2) Audio is a little delayed over Bluetooth headphones, enough to be noticeable. This is probably inevitable, but can be fixed by advancing audio about +100ms and setting for all videos, however this then becomes a problem when not using Bluetooth headphones as audio is now unnecessarily advanced, and again out of sync. No idea if it's possible, but perhaps some sort of automatic audio compensation is possible, ie. always advance audio +100ms (or maybe a configurable amount) when using Bluetooth output?

**Comment by samnazarko** [comment]
@lrusak 

Why have you chosen to create a PulseAudio sink? 

>but it is something we need otherwise pulseaudio won't be used

The ALSA sink, when the PA sound server is running should have a module which bridges the two (i.e. you can play to a BT PA sink without creating a CAESink()). This is the case on OSMC with PulseAudio 5.0. This is allowing streaming from a phone to OSMC and from OSMC to a phone or Bluetooth speaker via Kodi.

I was always under the impression that Pulse badly breaks passthrough if we use a Pulse sink in Kodi. This may not be the case with Pulse and ALSA together, but then you may confuse the user / have to think about default options, especially on Pi where we will now have AESinkALSA, AESinkPULSE, AESinkPI...

**Comment by lrusak** [comment]
Solved the kodi crashing issue by keeping the module module-always-sink

@samnazarko Thanks for chiming in but I'm a bit confused about what your are saying.

I have made some changes to this PR since my first message and I will edit it to explain what this PR currently does.

This PR, with the changes to our OE settings add-on allows the user to use exclusively pulseaudio or alsa. So with pulseaudio active the following is possible:
 - On x86_64 this can be a bluetooth audio device or any other hardware audio device.
 - On the RPi this can only be a bluetooth device (we do not load snd_bcm2835 by default so no alsa devices are enumerated). This may change at some point.

I hope this clears things up.

**Comment by samnazarko** [comment]
Thanks for your clarification. I was not aware that you were not loading snd_bcm2835 with Pulse on Pi.

>any other hardware audio device.

This is what I am unsure of: which devices? After modprobing the appropriate module, then regardless of x64/ARM, we are able to get PulseAudio network streaming (from network clients) and BT working (both directions), without actually using AESinkPULSE. The current implementation in this PR means that on a Pi you can stream audio from Kodi to a Pulse device (such as a BT speaker), but not from a phone via A2DP to Pi (unless you modprobe snd_bcm2835). 

Another advantage is that we ship PulseAudio as an 'add-on', so if we don't need to link against libpulse with Kodi, we have a smaller footprint for those that don't use this feature.

**Comment by lrusak** [comment]
@samnazarko 

>any other hardware audio device.
by this I mean whatever the default alsa devices are.

Regarding the RPi we do not load snd_bcm2835 at all, when using pulseaudio or otherwise. 

So what you're saying is that you are only using AESinkALSA even when pulseaudio is running because the alsa sink just points to pulseaudio anyway? I can look into this, but we don't use alsa at all in kodi on the RPi (unless you use some dt overlay)

**Comment by samnazarko** [comment]
>are only using AESinkALSA even when pulseaudio is running because the alsa sink just points to pulseaudio anyway?

Yes -- correct. If you modprobe snd_bcm2835 or add to /etc/modules.d (maybe for OpenELEC, put it in ExecStartPre= in your PulseAudio systemd unit), then even without building against libpulse (check ldd) or building with --disable-pulse, you can get BT audio out from Kodi to a BT device and from a BT device to the Pi or x64 box. Obviously, because stream goes to Pulse, Kodi is not aware of playback, so we need to be careful of that; as starting a movie while streaming from a phone could be 'interesting'.

**Comment by lrusak** [comment]
@samnazarko thanks for the insight, I will have to experiment :+1:

**Comment by fritsch** [comment]
This method "works" - but it will route the Default device through pulseaudio. This means that users that play audio files with the default device are limited to the setup pulseaudio capabilities. This can be really problematic as in other cases, e.g. kodi opens the HDMI output device, which is also "blocked" by PA won't open. And at any time, e.g. PA plays a sound of its own - the device is taken back.

This is btw. the very same as if you start kodi with:

```
AE_SINK=ALSA kodi
```

that way kodi will also only enumerate ALSA devices and ignore PA completely. That way you can still build against PA and don't need kodi patches.

lrusak:
I have created another branch here, https://github.com/lrusak/OpenELEC.tv/tree/fritsch_pulseaudio
This works as @samnazarko and @fritsch describe, however, on the RPi2 I cannot seem to get passthrough working when selecting the ```Pulseaudio Sound Server``` device in kodi.

@MilhouseVH has seen this and it has been reported in the builds that he makes (as he included both these branches)

Maybe @samnazarko can chime in and say if passthrough is even possible when using pulseaudio and if so we can maybe see how pulseaudio lists the devices (particularly HDMI)

If this isn't possible then we will have to go with the PR here as it allows you to disable pulseaudio completely.

samnazarko:
Passthrough with PulseAudio as the actual sink seems problematic. When we use CAESinkALSA and just use the 'PulseAudio sound server', you will have problems with passthrough as well -- unless the user manually changes back to the right sink. 

BTW @lrusak -- you added a disable PulseAudio option to OE settings? For us -- even when I disable with sudo systemctl disable pulseaudio; it is running on reboot. I get the impression it is DBus socket activated somehow.. 

>This can be really problematic as in other cases, e.g. kodi opens the HDMI output device, which is also "blocked" by PA won't open. And at any time, e.g. PA plays a sound of its own - the device is taken back.

Yes -- we are working on a 'dummy' player. When Pulse starts, we have an application launched from ExecStartPost= and it monitosr DBus for BT traffic (incoming streams). If we get one, we use our adjusted kodi-send API to send track information and artist information. Our dummy player lets you get the visualisations and makes sure no one tries to play from Kodi as well (ApplicationMessenger is called to suspend AE). 

I think that there will be fewer issues with Kodi sending to Pulse, and more with the device (Pi, x64 etc) receiving a stream (from BT or Pulse over the network) and Kodi not being aware of it.

samnazarko:
>if so we can maybe see how pulseaudio lists the devices (particularly HDMI)

use ```pactl list```. It should be the same as what Kodi is 'aware' of.

lrusak:
replaced by #4482

PR #4434, opened 2015-11-09
glibc: fixed locale support by adding on-target locale archive genera…

hsteinhaus:
…tion

The glibc package seems to include support for localized programs. However, the current support is incomplete as there are are no means for generating a locale archive. This PR adds a systemd-based approach to generate this archive during the first boot on target, localization is available from the second boot of the target. 

Unfortunately there seems to be no way to generate the archive at the build time  - different archives are generated when trying so (which aren't working).


lrusak:
Why is this needed? To me this just looks like bloat that is not needed :)

hsteinhaus:
There are people in the world, that do not speak English (AFAIK  around 6.5 billion).

Without proper locale support, almost all applications (except kodi, that brings own loclisation via language addons) fall back to an english user interface, which is in fact impossible to use for these people. Affected applications in stock OpenELEC that are visible to the user are e.g. VDR and it's web interfaces.

lrusak:
Why don't we package this as an add-on then?

5MB is almost a 4% increase in size of the base image. This is not a small increase.

hsteinhaus:
I think the locale support is an integral part of glibc, just like malloc() or fopen(). It is still possible to build applications on a glibc not providing this functionality, but this application will be heavily limited in its own functionality.

On the other hand, setting GLIBC_LOCALE to "no" allows to remove any traces of locale data and related services.  What would be the advantage of moving them into an addon?

vpeter4:
I'm using extra locales in my lamp addon (build from eglibc package). Before starting httpd daemon I set LOCPATH environment variable with path where locales are saved. Seems to be working.

rdorsch:
I also hit this issue while searching for localization support of the VDR live plugin. See http://openelec.tv/forum/20-development-discussion/78261-vdr-pvr-backend-local-language-in-live-plugin for details.

chewitt:
@sraue - please review this and make a decision (merge or close). Thanks.

PR #4439, opened 2015-11-13
RPi/RPi2: use rc-core-based gpio_ir_recv driver and ir-keytable instead of lirc_rpi/lircd for GPIO IR rereivers.

hsteinhaus:
This PR enables the ir_gpio_recv driver by adding a Device Tree overlay. This driver, together with ir-keytable, is a perfect replacement for the combination of lirc_rpi and lircd.

Using the kernel-based rc_core drivers leads to a way better user experience, as the remote control  respone is faster, smoother and more consistent. Addtionally, it prevents the usual double-key trouble often caused by LIRC/gpio_lirc, without delay-prone application-side filters.

How it works:
The kernel patch introduces a new Device Tree overlay, which will be compiled together with the Linux kernel. When installing the boot loader, the DTOverlay will be copied to the boot partion and added to the config.txt. 

When using the ir_gpio_recv, the LIRC daemon is no longer needed on the Raspberries, but it doesn't hurt if it is present either. However, the config.txt should not contain a dtoverlay=lirc-rpi line, as both drivers (loaded automatically by the kernel) would probably race for pin resources.

How to use:
1. make sure that config.txt contains the line "dtoverlay=ir-gpio" and the receiver is connected to GPIO#18
2. enable appropriate protocols (example: RC5 and RC6): "ir-keytable -p RC5,RC6", load keytable as necessary
3. make sure that eventlircd is active
4. test it: "irw /run/lirc/lircd"

### DIFF
--- a/packages/tools/bcm2835-bootloader/files/3rdparty/bootloader/config.txt
+++ b/packages/tools/bcm2835-bootloader/files/3rdparty/bootloader/config.txt
@@ -103,6 +103,9 @@
 # decode_DTS=0x00000000
 # decode_DDP=0x00000000
 
+# enable rc-core driver for RPi-style GPIO IR receiver at GPIO pin 18
+dtoverlay=ir-gpio
+
 ################################################################################
 # End of default configuration
 # all values below this line were inserted from config.txt.bk (your old config)
--- a/packages/tools/bcm2835-bootloader/scripts/update.sh
+++ b/packages/tools/bcm2835-bootloader/scripts/update.sh
@@ -32,6 +32,7 @@
 
   cp -p $SYSTEM_ROOT/usr/share/bootloader/*.dtb $BOOT_ROOT
   cp -pR $SYSTEM_ROOT/usr/share/bootloader/overlays $BOOT_ROOT
+  cp -p $(get_build_dir linux)/arch/arm/boot/dts/overlays/ir-gpio-overlay.dtb $BOOT_ROOT/overlays/
 
 # cleanup not more needed files
   rm -rf $BOOT_ROOT/loader.bin
--- a/projects/RPi2/patches/linux/linux-02-gpio_ir_rpi.patch
+++ b/projects/RPi2/patches/linux/linux-02-gpio_ir_rpi.patch
@@ -1,0 +1,51 @@
+diff -Naur linux-4.1.12/arch/arm/boot/dts/overlays/ir-gpio-overlay.dts linux_new/arch/arm/boot/dts/overlays/ir-gpio-overlay.dts
+--- linux-4.1.12/arch/arm/boot/dts/overlays/ir-gpio-overlay.dts	1970-01-01 01:00:00.000000000 +0100
++++ linux_new/arch/arm/boot/dts/overlays/ir-gpio-overlay.dts	2015-11-13 19:40:22.893626713 +0100
+@@ -0,0 +1,36 @@
++// Definitions for lirc-rpi module
++/dts-v1/;
++/plugin/;
++
++/ {
++	compatible = "brcm,bcm2708";
++
++	fragment@0 {
++		target-path = "/";
++		__overlay__ {
++			ir_gpio: ir-receiver {
++				compatible = "gpio-ir-receiver";
++				gpios = <&gpio 18 1>;
++
++				// parameter for keymap name
++				linux,rc-map-name = "rc-rc6-mce";
++			};
++		};
++	};
++
++	fragment@1 {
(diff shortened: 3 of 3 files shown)


lrusak:
Where does this patch come from?

My problem with this PR is that OE across all systems uses the same lirc implementation. Using this PR requires manual intervention to change how lirc works on the RPi, which isn't overly user friendly.

packages/tools/bcm2835-bootloader/scripts/update.sh
```diff
@@ -32,6 +32,7 @@
 
   cp -p $SYSTEM_ROOT/usr/share/bootloader/*.dtb $BOOT_ROOT
   cp -pR $SYSTEM_ROOT/usr/share/bootloader/overlays $BOOT_ROOT
+  cp -p $(get_build_dir linux)/arch/arm/boot/dts/overlays/ir-gpio-overlay.dtb $BOOT_ROOT/overlays/
```
lrusak: This function (and directory) do not exist on a running system. This line isn't even needed all together.

packages/tools/bcm2835-bootloader/files/3rdparty/bootloader/config.txt
```diff
@@ -103,6 +103,9 @@
 # decode_DTS=0x00000000
 # decode_DDP=0x00000000
 
+# enable rc-core driver for RPi-style GPIO IR receiver at GPIO pin 18
+dtoverlay=ir-gpio
```
lrusak: This will only be included in new systems, anyone who has any existing install would have to add this manually.

MilhouseVH:
This dts should be sent upstream - submit (just) your dts pull request here: https://github.com/raspberrypi/linux

Assuming it is merged upstream it will be picked up by OpenELEC and any RPi/RPi2 users that want to use it can then add it to their config.txt.

The way this PR is written:
* It will only build for RPi2 users (what about RPi1 users?)
* It will be installed automatically for new RPi2 users (even if they don't want/need it)
* Upgrades for all users will produce an error (due to non existent function/directory)

At the very least the config.txt and update.sh changes should be dropped which just leaves the dts change, in which case it doesn't need to be here at all - submit it to the upstream repo which automatically addresses the lack of RPi support. If you need this in your build you can pull it from the upstream repo until it is merged and finds its way into OpenELEC.

hsteinhaus:
> Where does this patch come from?
The patch comes from diff'ing the a scratch-written dts file against an empty file ;-)

> My problem with this PR is that OE across all systems uses the same lirc implementation. Using this PR requires manual intervention to change how lirc works on the RPi, which isn't overly user friendly.

With 99% of the remotes for media PCs on the market today the LIRC layer is not needed, because it is talking to /dev/input/event* directly and all decoding functionality is provided by the kernel. With may be two exceptions: the Pi GPIO receiver and the homebew serial stuff (still anybody out there using that?). Kodi on the other hand is talking to eventlircd, so it won't note a difference. When putting LIRC in between, this only seems to add delay and inconsistent timing (from button pressed to reaction on the screen), plus sometimes doubling pressed buttons. 

I tried to use lirc_pi for about a week now with about 5 different remotes, the response behavior  drove me nuts. I guess nobody who is not a dev will ever accept an entertainment device that reacts so mulish on RC commands.

Regarding Pi vs. RPi2: a Pi2 is what I have here for testing, but port to Pi1 should be staight forward.

@MilhouseVH 
Will consider that.

packages/tools/bcm2835-bootloader/files/3rdparty/bootloader/config.txt
hsteinhaus: Correct, but I don't have an idea how to solve that. BTW, lirc_rpi needs the same manual intervention.

packages/tools/bcm2835-bootloader/scripts/update.sh
hsteinhaus: Ok, good argument. Need to find something different here...

hsteinhaus:
Submitted an enhanced (parameterizd) and renamed  DTS file as https://github.com/raspberrypi/linux/pull/1199.

Closing this PR due to the changes to the DTS in the above PR.

aharonha:
This indeed broke my IR setup, my remote does not work after configuring it using the previous version

See my topic on the forum
http://openelec.tv/forum/133-installation/79014-lirc-not-working-after-upgrade-from-5-0-8-to-6

lrusak:
@aharonha this PR was never merged, I'm not sure why you are commenting here.

PR #4442, opened 2015-11-16
libressl: update ca-cert bundle

lrusak:
The old bundle was from 2011.

Thanks @stefansaraev 

I have also included a bump for libressl.

This should be picked for openelec-6.0 also


stefansaraev:
you may remove packages/security/libressl/cert/certdata.txt - thats not used at all.

lrusak:
updated :+1:

stefansaraev:
why bump ?

lrusak:
because newer is better?

but for real, libressl 2.2.2 removes support for sslv3

stefansaraev:
> because newer is better?

did soname change? api/abi changes? anyone? if you are going to push this to OE6. you will fail.

I wont play this game anymore. do whatever you want.

lrusak:
updated and removed libressl bump

MilhouseVH:
This change seems to be responsible for Pandora no longer working:

http://xbmclogs.com/pvo9hrqda (line 973)

[forum confirmation](http://forum.kodi.tv/showthread.php?tid=231092&pid=2174761#pid2174761)

InuSasha:
seems the same like here: https://bugs.launchpad.net/ubuntu/+source/openssl/+bug/1014640
post 13 may be helpful

MilhouseVH:
If there is no easy fix, should this at least be reverted for the time being as it's not been a positive change, and the old bundle didn't seem to be causing any problems (unlike the new bundle).

InuSasha:
in the old bundle are invalid RootCAs, so this is dangerous...
A quick fix (and dirty one) is to add the cert.
An other solution can be, to switch to an other bundle (from mozilla for example)

my theoretical 2cents,
Sascha

lrusak:
@MilhouseVH can you test this again? it seems that it was cause by a misconfiguration on pandoras side. See, https://github.com/mcrute/pydora/issues/27

MilhouseVH:
I'll ask the users that were reporting the problem for an update.

MilhouseVH:
#4498 fixes this.

PR #4443, opened 2015-11-16
Nvidia_Legacy/linux: sync kernel config with Generic

lrusak:
There is no reason to have different kernel configs for the two projects. Just keep them the same so it's easier to keep track of changes.


sraue:
please make PR for master branch too

lrusak:
Master should be in sync already from linux 4.3 merge.

Issue #4333, opened 2015-09-19
5.95.5: WiFi Set to "Inactive" on Reboot


michaelrcarroll:
## Setup Details
OpenELEC Version: 5.95.5
Build: Generic x86_64
Hardware: Intel NUC DC3217BY w/ Intel 7260 WiFI

## What Did I Do?
A working config on 5.0.8, I use WiFi to connect to MySQL database for library. Use *Wait for Network* to delay Kodi startup while WiFi connects. Upgraded from 5.0.8 to 5.95.5 for testing.

## What Happened?
* After update, no network is active
* Going into OpenELEC Configuration and setting WiFi to *active* turns on the WiFi
* Performing ```systemctl restart kodi``` reconnects to database
* Upon reboot, WiFi is set back to *inactive*
* Not usable in this state. Reverted to 5.0.8 and problem is gone

## What Should Happen?
* WiFi should remain *active* across reboots

## Logs
* Kodi.log: http://sprunge.us/NaGO
* System.log: http://sprunge.us/MHXg
* Hardware.log: http://sprunge.us/PiVV
* Audio.log: http://sprunge.us/EHiN
* Network.log: http://sprunge.us/Adeh
* Input.log: http://sprunge.us/jdAK
* Filesystem.log: http://sprunge.us/FKFM
* Journal_cur.log: http://sprunge.us/gadh
* Journal_prev.log: http://sprunge.us/TXij

For now I've reverted to 5.0.8, but I'd really like to be up to date if possible. If I can provide any additional information to help debug, I'd be happy to.

ggp759:
This also is happening to me all the time with the latest beta on wifi. Had to use wired. Can anyone please fix this? Also sometimes both wifi and wired are enabled at the same time which is odd

MrSparklle:
Same issue here with OpenElec 6.0.0
It backs to work if you reset the adpater by SSH:

connmanctl disable wifi
connmanctl enable wifi

michaelrcarroll:
fwiw: I've moved to kodibuntu *because* of this issue. Remote library (mysql) was unusable because of the lack of network connectivity at boot.

spixlaren:
[...]

PR #4455, opened 2015-11-19
connman: update to connman-1.30

chewitt:
This updates to 1.30 and removes four patches now included upstream. We can't stick on old releases forever so lets see what breaks :)


chewitt:
FWIW.. the 5.95.5 and 6.0.0 releases I made for ATV both included this without apparent issues. I'm not sure that counts too much, but fingers crossed..

chewitt:
in the spirit of not ignoring advice ..commitstrip is now added to my rss feeds :)

stefansaraev:
here is what you must be ABSOLUTELY sure is not broken before merging this PR:

1. go wired.
2. set dhcp
3. go static
4. go back to dhcp
5. reboot

MilhouseVH:
I'll can include this in my next test build. Although the builds are RPi/RPi2 only, they are at least being used a fair bit and may give a slightly higher level of confidence before merging.

lrusak:
@stefansaraev Just tested your concerns and it doesn't seem to be an issue. I can acquire a dhcp lease just fine after reboot.

Anything else that should be tested?

stefansaraev:
merge away and find out.

chewitt:
Are there other packages that have (or should have) mandatory testing steps? ..and what are those steps? - we need to record these things somewhere.

MilhouseVH:
> Are there other packages that have (or should have) mandatory testing steps?

All of them probably should have mandatory testing... bumping packages "just because" is Russian Roulette. Though the reality is that nobody is going to be interested in managing any test process.

Perhaps the more critical packages, and those with a bad track record, should be more carefully managed when being bumped... curl, connman, systemd etc.

chewitt:
https://github.com/OpenELEC/OpenELEC.tv/pull/4366 is a move towards putting some rules around how changes to core and non-core packages are handled. It's based on "lessons learned" but hasn't been merged yet.

fritsch:
This won't help. I tried a discussion lately on IRC before I closed my OpenELEC membership.

OpenELEC I was told is and always was and will be: Latest, greatest, stuff. Every core lib is bumped, two days after it is out. This stuff is only (if at all) compile tested (see latest imx, wetek regressions, amd oss regression and so on). Users shall be endtesters and upstream reports don't happen. Stable version don't need to be stable as .x can be released within no time.
So - it's nice that you want to bring that up - but it won't make it. The kodi team members of you can read the backlog of team kodi's IRC channel.

The policy you link here would produce a stable system - which is - as I was told - not wanted.

MilhouseVH:
Possibly connman-1.30 related: http://forum.kodi.tv/showthread.php?tid=238393&pid=2168307#pid2168307

Could the network hardware now be seen differently by this new connman which might cause (some) working systems to require reconfiguration? Shouldn't IMHO stop the bump, not sure if there's an easy solution but maybe just something to be aware of.

chewitt:
The connman service name format hasn't changed in aeons .. my personal suspicion is something might be different with dbus signalling - maybe a message that isn't being trapped and handled right.

MilhouseVH:
Oh well, the dbus bump just merged...

chewitt:
I mean something probably changed in connman and our settings addon needs to be tweaked to handle the messages being send via dbus - I doubt the dbus change itself has any effect.

sraue:
@lfiebach ^^^^ can you check our settings addon with the latest master and dbus-1.10?

MilhouseVH:
@sraue/@chewitt: The WiFi needing to be reactivated every boot could be related to #4333

MilhouseVH:
This might be causing a reboot/shutdown issue with RPi2 - haven't seen it myself, but it's been reported on the forum [here](http://forum.kodi.tv/showthread.php?tid=231092&pid=2171077#pid2171077).

Reverting the connman-1.30 bump appears to resolve the issue. In the forum post are shutdown logs with and without connman-1.30, in case they provide a clue - it looks like connmand-1.30 doesn't continue past `Stopped target Multi-User System.`.

lfiebach:
Can someone post a kodi.log with OE DEBUG enabled ?

MilhouseVH:
@lfiebach: see comment above yours, follow the forum link...

lfiebach:
@MilhouseVH
In this logs is no Addon Debug enabled

MilhouseVH:
You're right, and they're not even kodi.logs... :)

However the wifi automatic connection issue is resolved after #4470, so no further connman-1.30 issues from me.

sraue:
thanks much!

MilhouseVH:
Frack... frack! Forgot this is also causing a shutdown issue on RPi (one report) - hopefully this is also related to the systemd-228 crap (and is now fixed). Will see if the reboot issue remains after the next build.

MilhouseVH:
Reboot issue returns after the connman bump.

alex240188:
Here is a debug log if it is any use
http://pastebin.com/0kLfYzqs

PR #4457, opened 2015-11-20
dbus: update to version 1.10.4

lrusak:
@MilhouseVH can you include this in your builds?


MilhouseVH:
Will do, I've also added #4458.

sraue:
@MilhouseVH all ok so far?

MilhouseVH:
All OK so far (one minor issue with connman that I'll post in 4455).

PR #4465, opened 2015-11-24
glibc: add upstream patch

lrusak:
see, https://sourceware.org/bugzilla/show_bug.cgi?id=18781

Thanks to seo for pointing out ;)

This should be picked for openelec-6.0 also


stefansaraev:
yea. I could not care less for the "stable" releases...

sraue:
added in https://github.com/OpenELEC/OpenELEC.tv/commit/2490495df17f3b63ddda78abf1c2e7be77a5984c and https://github.com/OpenELEC/OpenELEC.tv/commit/0507ee3c5a1b13d6a0d44c3cbdf8d0ceac2fd250

PR #4469, opened 2015-11-27
Add pvr hdhomerun

MilhouseVH:
Both commits are needed to successfully build [`pvr.hdhomerun`](https://github.com/kodi-pvr/pvr.hdhomerun).

Creating the pkconfig file in `libhdhomerun` is a rather ugly hack, due to `libhdhomerun` not having a usable `make install` and not installing itself as a library, as it should.

Hopefully Silicon Dust can update their package in future to avoid hacks like this.

### DIFF
--- a/packages/multimedia/libhdhomerun/package.mk
+++ b/packages/multimedia/libhdhomerun/package.mk
@@ -41,4 +41,10 @@ makeinstall_target() {
 
   mkdir -p $INSTALL/usr/lib/
     cp -PR libhdhomerun.so $INSTALL/usr/lib/
+
+  mkdir -p $SYSROOT_PREFIX/usr/include/hdhomerun
+    cp *.h $SYSROOT_PREFIX/usr/include/hdhomerun
+
+  mkdir -p $SYSROOT_PREFIX/usr/lib
+    cp libhdhomerun.so $SYSROOT_PREFIX/usr/lib
 }
--- a/packages/mediacenter/kodi-binary-addons/pvr.hdhomerun/package.mk
+++ b/packages/mediacenter/kodi-binary-addons/pvr.hdhomerun/package.mk
@@ -1,0 +1,50 @@
+################################################################################
+#      This file is part of OpenELEC - http://www.openelec.tv
+#      Copyright (C) 2009-2014 Stephan Raue (stephan@openelec.tv)
+#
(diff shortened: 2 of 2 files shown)


lrusak:
I think it would be nicer to have the libhdhomerun.pc file included as a patch. Then simply one line can be used in the package.mk to copy this file to it's destination.

MilhouseVH:
Where should the patch go, as it's not a traditional patch, just have it in the root of the libhdhomerun package directory?

stefansaraev:
@lrusak this was quick and dirty fix proposed by me. a propeper fix is not only a patch that adds pkgconfig file to libhdhr sources

fyi. fedora is doing exactly the same, see their rpm spec file.

if you want to support silicondust's hardware, and if you want it properly fixed, you should work with upstream to get-it-done-right. if not - close this and enough drama for today ;)

lrusak:
How about not even needing the pkg-config file.
see, https://github.com/OpenELEC/OpenELEC.tv/compare/master...lrusak:add_pvr.hdhomerun

MilhouseVH:
Updated, thanks @lrusak.

packages/multimedia/libhdhomerun/package.mk
```diff
@@ -42,3 +42,10 @@ makeinstall_target() {
   mkdir -p $INSTALL/usr/lib/
     cp -PR libhdhomerun.so $INSTALL/usr/lib/
 }
+
+post_make_target() {
```
lrusak: not needed, just include it in the makeinstall_target()

lrusak:
Just the one minor thing then I will merge. I already uploaded the source to the ftp

packages/multimedia/libhdhomerun/package.mk
```diff
@@ -42,3 +42,10 @@ makeinstall_target() {
...
+  mkdir -p $SYSROOT_PREFIX/usr/include/hdhomerun
+    cp *.h $SYSROOT_PREFIX/usr/include/hdhomerun
+  mkdir -p $SYSROOT_PREFIX/usr/lib
```
lrusak: add space for readability

MilhouseVH:
Updated - hopefully now OK.

lrusak:
Thanks :)

PR #4474, opened 2015-11-30
scripts/create_addon: fixes

lrusak:
I'm not sure why this didn't happen already :)
It is useful when you are making an add-on repo

As for using the specific add-on repo, this helps to maintain separate repo's

### DIFF
--- a/scripts/create_addon
+++ b/scripts/create_addon
@@ -122,6 +122,9 @@ pack_addon() {
     if [ -f $ADDON_BUILD/$PKG_ADDON_ID/icon.png ]; then
       cp $ADDON_BUILD/$PKG_ADDON_ID/icon.png $ADDON_INSTALL_DIR/icon.png
     fi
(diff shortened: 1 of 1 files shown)


stefansaraev:
my 2 cents on this: 

you are not supposed to build both official an unofficial addons in same build environment. if you do - eventualy all sort of crap may happen to officials because optional dependencies built/installed / headers available that come from unofficial.

that's why this was never done before.

build in a separate build envs for unofficial and official repos, for 5.0 / 6.0 and so on. save yourself (and your users) a trouble.

lrusak:
Ah, yes I understand what you mean and it is a good point. I will drop it and find a different solution for myself.

stefansaraev:
I do not mean this PR is not acceptable. push it if you want. 

> The other commit still makes sense to me though

hum. why would you want addon.xml in zipdir ?

PR #4477, opened 2015-12-01
Jarvis egl

Lusephur:
Is it possible to add this patch found in imx6 to the intel egl build. 
https://github.com/fritsch/OpenELEC.tv/blob/jarvis-egl/projects/imx6/patches/linux/linux-010_tivo_slide_pro.patch


lrusak:
Please be more careful about what you are doing

PR #4486, opened 2015-12-14
Roll nvidia-legacy driver into Generic

lrusak:
The title says it all.

Some said it couldn't be done, some said it shouldn't be done.
This is something that should have been done instead of creating the Nvidia_Legacy project.

It sucks that Nvidia forces us to do this but we don't have any other choice.

I have created a script to easily get the pciid's of the supported cards for the latest nvidia driver.
This script will create a set of udev rules and should be run anytime the nvidia driver is updated.

All other libs and binaries are symlinked and setup depending on the driver version need for the detected card.

This increase the size of the Generic image by 24MB.

I have tested with my Nvidia GeForce GT 520
@MilhouseVH has tested with his ION2


lrusak:
@sraue need some input here. It would be nice to have this in before branching 7.0

CvH:
Wasn't there in the past (OE 2?) the same approach for the nVidia drivers and they failed horribly due to double assigned id's from nVidia ?

lrusak:
@CvH Not sure exactly, but with this method it basically checks if the device is supported by the newer driver and if not it falls back to the legacy driver.

stefansaraev:
yes. that was what happened. in long term - this is a no go and will cause more harm than good. legacy should go away after OE 7. my 2c.

lrusak:
> and that will fail miserably, soon or later.

why?

stefansaraev:
merge it and find out ;)

MilhouseVH:
Could you move `packages/linux/patches/4.4-rc5/linux-020-export-mtrr-symbol.patch` to a project patch (ie. `projects/Generic/patches/linux/linux-020-export-mtrr-symbol.patch`)?

This would accomplish two things:

1. This PR has no dependency on a specific version of Linux (ie. 4.4-rc5, or 4.4-rc6) as 4.4 Linux is a moving target right now
2. The `linux-020-export-mtrr-symbol` patch - which is a horrible hack  - should be limited in scope to the Generic projects, as it's not needed or wanted by RPi, Wetek etc. projects.

lrusak:
Will do

lrusak:
Done. now it doesn't depend on the linux version.

MilhouseVH:
Many thanks!

sraue:
please remove the Nvidia_Legacy drop from this PR, this can and should be done later after testing

lrusak:
Done

PR #4487, opened 2015-12-14
libaacs: add upstream patches

lrusak:
This fixes #4378 

patches are from, http://git.videolan.org/?p=libaacs.git;a=search;s=Roland+Fischer;st=author


MilhouseVH:
Message from Roland, something to bear in mind when bumping libaacs in future:

>Good news - although my final fix didn't make into the branch the main developer behind libaacs fixed the problem a lot better than my fix.
>
>The disc itself is completely valid - the aacs specification contains instructions how unencrypted content can be included and this part of the specification is now also supported (support for the cci files on the disc).
>
>So when you include the latest master of libaacs no additional patches are needed any longer

PR #4494, opened 2015-12-15
ffmpeg: Package kodi's ffmpeg releases directly

lrusak:
This makes sense as the kodi-ffmpeg tracks upstream.

This removes some kodi specific patches as well


stefansaraev:
:+1: 

offtopic but imo worth mentioning: https://github.com/stefansaraev/OpenELEC.tv/blob/K%2B%2B%2B/scripts/unpack#L73-L89
^^ with this you dont need 1. unpack() hacks for github sources 2. scripts/extract. meh.

lrusak:
Thanks for the tip :+1: I will look into it

PR #4496, opened 2015-12-15
libinput: update to 1.1.2

lrusak:


lrusak:
Merging because I am probably the only person that uses this :)

PR #4488, opened 2015-12-14
ffmpeg: enable mjpeg encoder

stefansaraev:
this may be required for kodi 17

EDIT: so enable it sooner than later.

PR #4497, opened 2015-12-15
ffmpeg: enable png encoder

MilhouseVH:
Following on from #4488, this will be required by https://github.com/xbmc/xbmc/pull/8583 to enable `png` encoding. No harm in enabling now.


lrusak:
Just saw this as well, thanks.

PR #4498, opened 2015-12-15
Update cacert.pem and provide a new method of creating it

lrusak:
Method was taken from fedora
see, http://pkgs.fedoraproject.org/cgit/ca-certificates.git/tree/ca-certificates.spec

script syntax isn't great, but it only needs to be run once in a while.


lrusak:
Seems that this fixes the issue that was previously described. [link](http://forum.kodi.tv/showthread.php?tid=250817&pid=2188057#pid2188057)

Will wait 24hours or so to see if any issues turn up otherwise I'll merge.

MilhouseVH:
Another positive report on x86 - looks good to merge.

PR #4505, opened 2015-12-17
openelec-6.0 backports

lrusak:
I'll include various backported commits here.

Let me know if I missed anything.


lrusak:
good to go :+1:

PR #4507, opened 2015-12-18
audiodecoder.usf: don't build on RPi

chewitt:
package build fails on RPi without this (and is currently missing from 6.0 repo) as first gen. pi doesn't support neon fpu. - this has been "build" tested, I've no idea how to test the actual decoder


stefansaraev:
I dont think this will work on rpi. quoting notspiff

```
lazyusf requires simd, neon or sse.
```

lrusak:
We'll just have to use
```
PKG_ADDON_PROJECTS="Generic Nvidia_Legacy RPi2 imx6 WeTek_Play"
```

chewitt:
I've updated and rebased the patch so this is not built on RPi

PR #4509, opened 2015-12-20
Alsa updates

sraue:


sraue:
@lrusak test please if we need PA support in SDL (disable there and test with enabled PA support)

PR #4512, opened 2015-12-21
Package updates

sraue:


lrusak:
Didn't last time we bumped iptables something broke?

I still think we should track kodi's version of ffmpeg.

sraue:
kodi is on 2.8.4 (rc)
we dont touched iptables for ages... the broke was because we built statically

MilhouseVH:
> I still think we should track kodi's version of ffmpeg.

+1. What's the advantage of maintaining a potentially different ffmpeg package when it's only really used by Kodi?

sraue:
we always use the upstream packages, like every distro does... they should see to include their patches upstream fast as possible... and in the end we use the same version with their patches

MilhouseVH:
But in the meantime, if Kodi adds useful patches in their version of the ffmpeg package, we need to carry those same patches too (until they're merged upstream). Or maybe we miss them out, and then experience unusual and difficult to explain behaviour that is already fixed by Kodi.

Sorry but I'm failing to see the advantage of slavishly tracking upstream given the extra effort and potential negatives involved when Kodi is the main (if not sole) consumer of the ffmpeg package, so IMHO we should track Kodi instead.

sraue:
every time they update the patches we must update the package... where is the advantage to update the package over updating the patches... there is no... and we provide the changes changes we have done over the original package and can follow the GPL better because we dont must provide the whole package byself. so we have no extra effort, we uses the same patches, we can use the upstream sources which we dont must provide and we provide only the changes we have done...

MilhouseVH:
> every time they update the patches we must update the package... where is the advantage to update the package over updating the patches... 

It's a lot easier and we're a lot less likely to miss something - instead of us updating against upstream then having to add whatever Kodi patches have been added we just need to bump to the new Kodi package - their patches are already in their package. It's a lot cleaner. We don't have to work out what has been added or removed, and replicate in our own repo, instead we just use their package.

Instead we're tracking a different package and then having to port over the Kodi patches as and when they're added. Seems like a lot of extra and unnecessary work that is already being done by Kodi, so we should take advantage of that.

chewitt:
in early OE days our XBMC peers were doing "occasional" maintenance on ffmpeg so tracking upstream directly had a positive effect on OE stability. Kodi now tracks ffmpeg as a critical app so that problem doesn't exist, and I think we gain more from reducing our maintenance burden than we lose maintaining our own package. @sraue: You're on public record complaining that package review and bumping work is boring. Let's grab the opportunity to do less of it.

sraue:
if we dont update the patch regulary we would miss the same then not updating the package regulary... there is really no benefit in using forks then using the upstream patches + the few patches which dont changes much.

@chewitt we dont maintain a package here byself... https://github.com/xbmc/FFmpeg/compare/644179e0d4155ae8f5ddd5c3f6bd003e2e13cf94...release/2.8-xbmc.patch is the whole patch we use.

this discussion can be moved to https://github.com/OpenELEC/OpenELEC.tv/pull/4494 it dont belongs here and takes more time then updating the patch the 3 years...

stefansaraev:
> we always use the upstream packages, like every distro does... they should see to include their patches upstream fast as possible... and in the end we use the same version with their patches

```
x220 OpenELEC.tv # find packages/ -iname *.patch | wc -l
327
```

cool story.

PR #4514, opened 2015-12-21
Linux 4.4 fixups

lrusak:
Might as well just leave this open until final is released. I will update it as we go.

RPi patches need updating but must wait for @popcornmix to rebase on rc6

let me know if I missed anything.


lrusak:
Updated and added new rpi kernel patches.

MilhouseVH:
This wasn't meant to be merged yet...

PR #4515, opened 2015-12-21
addons: update repo descriptions and lock unofficial to 6.0

chewitt:
These are the same cosmetic/description changes that I made to the 7.0 branch yesterday. We also need to lock the unofficial repo to 6.0 addons to avoid it picking up earlier 4.3 ones.


lrusak:
:+1: Thanks

PR #4519, opened 2015-12-22
samba: add restore share

chewitt:
Add a "Restore" share to make it easier for users to complete a backup > reimage > restore sequence once we grow Generic with multiple nvidia drivers.

### DIFF
--- a/packages/network/samba/config/smb.conf
+++ b/packages/network/samba/config/smb.conf
@@ -158,3 +158,11 @@
   public = yes
   writable = yes
   root preexec = mkdir -p /storage/backup
+
+[Restore]
+  path = /storage/restore
+  available = yes
+  browsable = yes
+  public = yes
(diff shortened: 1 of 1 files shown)


sraue:
every major release NEEDS a reinstall (which again can be reinstalled if anyone selcts the wrong build), this was always the case and will be in the future.... i see no need for this and this could be danger if anyone puts some random file there...

MilhouseVH:
This was discussed briefly on slack and I may have misunderstood the need for this, considering we already have the `Backup` share. What would this new `restore` share provide that the `Backup` share doesn't? The OE Backup is not a facility I've had much reason to use, but couldn't a user simply drop a suitable tar into the Backup share and then restore it?

chewitt:
I was misunderstanding the operation of the restore function - my bad - closing :)

PR #4520, opened 2015-12-22
addons: bump 6.0 official pkg_rev for rebuilds

chewitt:
please validate if anything else is needed? .. then I can get rebuilds done and pushed to the repo


sraue:
why it must be rebuild for 6.0.1? 6.0.1 is compatible to 6.0.0 and this will not be changed

PR #4522, opened 2015-12-23
distributions/OpenELEC/options: enable Pulseaudio

sraue:


sraue:
lrusak, please merge if there are no complains... does this work on RPi/WeTek/imx6 too? if not we should disable for this projects

lrusak:
Yes It should be fine all all projects.

only bluetooth audio and *sending* network audio is enabled by default. If people want to use there HDMI or analogue connections for pulseaudio they can do so manually by ```load-module module-udev-detect```

islipfd19:
Iv'e been testing the latest master builds for a while on RPi2.

How does one load this for pulse audio. 'load-module module-udev-detect'?
Running that at the CLI with and without 'systemctl' yields no results.
I've attempted to run them while having pulse audio select as the sound
server and not.

lrusak:
@islipfd19 we don't load the snd_bcm2835 module on the rpi so when loading the pulse module, module-udev-detect, it won't find anything. However this is unsupported currently and I'm not going to support it. Pulse is currently on for Bluetooth audio.

This isn't a place for discussion. Use the forums or IRC.

PR #4523, opened 2015-12-23
fuse: change download url

CvH:
- no nice dl url due to github

- fixes #4521

PR #4524, opened 2015-12-23
fuse : fix unavailable package due to bad url

aragua:
fuse packages are not available on sourceforge so url:
PKG_URL="$SOURCEFORGE_SRC/fuse/fuse-2.X/$PKG_VERSION/$PKG_NAME-$PKG_VERSION.tar.gz"
is deprecated and building the image with:
 PROJECT=imx6 ARCH=arm make image
fail because archive cannot be found.

Now the repo is available on github:
https://github.com/libfuse/libfuse.git

This patch modify the makefile informations for fuse and add a script to generate the package.

Before merging this patch, the archive genrated with ./tools/mkpkg/mkpkg_fuse must be copied on "$DISTRO_SRC/" mirror.
Be careful that the commit id has not changed since the creation of this patch.

### DIFF
--- a/packages/sysutils/fuse/package.mk
+++ b/packages/sysutils/fuse/package.mk
@@ -17,12 +17,12 @@
 ################################################################################
 
 PKG_NAME="fuse"
-PKG_VERSION="2.9.4"
+PKG_VERSION="0285462"
 PKG_REV="1"
 PKG_ARCH="any"
 PKG_LICENSE="GPL"
-PKG_SITE="http://sourceforge.net/projects/fuse/"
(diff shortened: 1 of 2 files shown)


sraue:
why the versioning change and whats wrong with downloading from https://github.com/libfuse/libfuse/releases/download/fuse_2_9_4/fuse-2.9.4.tar.gz instead?

sraue:
fixed with https://github.com/OpenELEC/OpenELEC.tv/pull/4523

aragua:
Hello

I have no problem with this url. I just didn't know that the package has
been available on github. =-D
Thank you for the fix.

Fabien

2015-12-23 23:11 GMT+01:00 Stephan Raue <notifications@github.com>:

PR #4529, opened 2015-12-25
pvr.mythtv: update to pvr.mythtv-5450056

CvH:
v2.8.4 @janbar 
this is the correct, up to date + stable, version for isengard right ?

### DIFF
--- a/packages/mediacenter/kodi-binary-addons/pvr.mythtv/package.mk
+++ b/packages/mediacenter/kodi-binary-addons/pvr.mythtv/package.mk
@@ -17,7 +17,7 @@
 ################################################################################
 
 PKG_NAME="pvr.mythtv"
-PKG_VERSION="8cb158c"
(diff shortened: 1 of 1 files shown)


chewitt:
@CvH - if bumping we also need to consider how/where we get the source package from. Please take a look at the mkpkg scripts for kodi and kodi-binary addons.

janbar:
Correct. The last version i released for Isengard in repository "kodi-pvr" is 2.8.4 with this commit.

janbar:
The official source of the package for the addon is "https://github.com/kodi-pvr/pvr.mythtv" and you have to choose de right branch named like kodi core, i.e Isengard , and next will be Jarvis.
Thanks

janbar:
Also, I think users will be happy when OpenELEC repo will up to date.

chewitt:
closing as superceded by #4531

PR #4531, opened 2015-12-26
addons: bump pvr.dvblink pvr.mythtv pvr.nextpvr - add kodi patch

chewitt:
Kodi sources contain outdated addon hashes (not updated since 15.2 final) so mkpkg scripts used to package kodi-binary-addon sources don't give latest packages. If applied to downloaded Kodi sources this patch updates hashes to current. It's not required to build Kodi - but we have to stash a reminder somewhere.


lrusak:
Looks good :+1:

chewitt:
good - will let you know when final tests are done

chewitt:
multi-lib patch added .. good to go

PR #4537, opened 2015-12-28
[scripts] Add PKG_SOURCE_NAME for non-standard package tarballs

MilhouseVH:
The vast majority of packages create source tarballs along the lines of `$PKG_NAME-$PKG_VERSION.tar.xz` (or `.gz` etc.).

However some packages do not, eg. https://github.com/OpenELEC/unofficial-addons/pull/96, which creates a package tarball with the file name `v4.0.8.tar.gz`, and requires an `unpack()` function to work around this situation.

With this PR, `PKG_SOURCE_NAME=$PKG_NAME-$PKG_VERSION.tar.xz` (or `.gz` etc.) can be added to `package.mk` to simplify the handling of these non-standard package filenames.

In the case of tvheadend, the entire `unpack()` function can be removed following the addition of `PKG_SOURCE_NAME=$PKG_NAME-$PKG_VERSION.tar.gz`. In addition, the tvheadend source tarball hosted on the OE mirror server can be given the "standard" file name `tvheadend-4.0.8.tar.gz` instead of the ambiguous `v4.0.8.tar.gz` that would otherwise have to be used.

With this change, existing `package.mk` files should work as before, including those with custom `unpack()` functions, or with multiple `PKG_URL`s (ie. [makemkv](https://github.com/OpenELEC/unofficial-addons/blob/master/addons/multimedia/makemkv/package.mk)).

Packages with a single `PKG_URL` could be switched to using `PKG_SOURCE_NAME`, if considered appropriate.

I would suggest switching `tvheadend/pacakge.mk` if this is ever merged.

### DIFF
--- a/config/path
+++ b/config/path
@@ -61,6 +61,8 @@ SED="sed -i"
   PKG_LICENSE="unknown"
   PKG_SITE=""
   PKG_URL=""
+  PKG_SOURCE_NAME=""
+  PKG_SOURCE_DIR=""
   PKG_DEPENDS_TARGET=""
   PKG_DEPENDS_HOST=""
   PKG_DEPENDS_INIT=""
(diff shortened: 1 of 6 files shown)


lrusak:
This generally looks ok to me, however, I wonder if it would be better to just force all packages to use the ```$PKG_NAME-$PKG_VERSION``` naming convention rather than adding ```PKG_SOURCE_NAME```. This would break makemkv but it could be worked around.

I would like @sraue to comment any ideas he has in order to solve this issue.

CvH:
Is there a case (besides makemkv) where it is necessary to name the `*.archive` different to
`$PKG_NAME-$PKG_VERSION` ?
This change would possibly break some out of repo addons but who cares (easy to fix)! 
Also all unpack() workarounds have to be fixed :) (if this is optionally nobody would fix them)

MilhouseVH:
>  I wonder if it would be better to just force all packages to use the $PKG_NAME-$PKG_VERSION naming convention rather than adding PKG_SOURCE_NAME.

I did think about automatically setting `PKG_SOURCE_NAME` in `config/path` once `package.mk` is sourced, but when you look at all the different file types being handled by `extract` (not just archives, but also patches, diffs etc.) then it's clear that setting it automatically may cause breakage, so I chickened out on that change.

In addition, any package that specifies multiple urls in `PKG_URL` will be a problem (ie. makemkv) if `PKG_SOURCE_NAME` is set automatically. That's not to say it can't be done, but `makemkv` is a problem.

If we were to set `PKG_SOURCE_NAME` automatically, then one option might be to set `PKG_SOURCE_NAME` in `config/path` if not already set and only if `basename $PKG_URL` is the same as `$PKG_NAME-$PKG_VERSION` - if there's a difference, then abort with an error as `package.mk` should be setting it manually.

Something like...
```
  if [ -z "${PKG_SOURCE_NAME}" ]; then
    if [[ $(basename "${PKG_URL}") =~ ^${PKG_NAME}-${PKG_VERSION}\. ]]; then
      PKG_SOURCE_NAME=$(basename "${PKG_URL}")
   else
      echo "Error - Non standard package source name, must set PKG_SOURCE_NAME in package.mk"
      exit 1
    fi
  fi
```

and `get` and `extract` can drop references to `basename $PKG_URL`, using `$PKG_SOURCE_NAME` exclusively.

Advantage: Any package that needs to set a custom `PKG_SOURCE_NAME` could do so (eg. tvheadend), the rest would automatically use the "standard" naming convention, and `PKG_SOURCE_NAME` would always be set with the name to be used.

Disadvantage: Assumes there is a solution for those packages specifying multiple urls in `PKG_URL` as they screw up this automatic solution (though not the solution as presented in this PR).

Ultimately I think the approach used by this PR is likely to cause fewer problems than any automatic solution.

> This change would possibly break some out of repo addons

It shouldn't break anything unless `PKG_SOURCE_NAME` is added to `package.mk` (or set automatically).

Existing packages will continue to work as they do now when `PKG_SOURCE_NAME` is not set, but could the packages be changed in due course if the addition of `PKG_SOURCE_NAME` is a better solution for them.

MilhouseVH:
Updated with my preferred solution - tested with clean builds on RPi/RPi2 and Generic. Not tested for imx6 or Wetek.

Updated `systemd`, `xf86-video-nvidia-legacy` and `xf86-video-nvidia` packages, eliminating or simplifying the `unpack()` function as required.

`makemkv` will require conversion to `mkpkg`. Other addon packages will require similar `unpack()` updates (eliminating or simplifying).

lrusak:
Although it adds complexity, I like it and it's what we need. I'll give it a test soonish. :+1:

sraue:
dont merge until i have looked in this, i think i dont like this, but not looked in this

MilhouseVH:
@sraue: thoughts?

sraue:
wait with this i need to check this

MilhouseVH:
How's the checking?

PR #4538, opened 2015-12-28
vdr-addon: fixes for missing dummydevice plugin and undefined symbol …

vpeter4:
…in restfulapi plugin


stefansaraev:
does strip_lto and / or strip_gold in cxxtools / restfulapi help ?

vpeter4:
Only strip_gold in restfulapi help.

vpeter4:
Oh, the other solution is just to pass LIBS with inversed libraries when calling make.

PR #4540, opened 2015-12-28
Added support for nspawn containers

hobbsAU:
I saw Sony2's Docker PR adding kernel support for namespaces. systemd has it's own mini container manager that ships with systemd itself utilising kernel namespaces for isolation. I've been using the systemd-nspawn for a little while now to quarantine my own containers preserving the rest of the OpenElec system.

nspawn allows you to run lightweight containers (including Docker containers) and can be enabled in systemd pretty easily...

> systemd-nspawn may be used to run a command or OS in a light-weight namespace container. In many ways it is similar to chroot(1), but more powerful since it fully virtualizes the file system hierarchy, as well as the process tree, the various IPC subsystems and the host and domain name.

Would it be possible to include in the main OpenElec project?

### DIFF
--- a/packages/sysutils/systemd/package.mk
+++ b/packages/sysutils/systemd/package.mk
@@ -78,7 +78,7 @@ PKG_CONFIGURE_OPTS_TARGET="ac_cv_func_malloc_0_nonnull=yes \
                            --disable-backlight \
                            --disable-rfkill \
                            --enable-logind \
-                           --disable-machined \
+                           --enable-machined \
                            --disable-importd \
                            --disable-hostnamed \
                            --disable-timedated \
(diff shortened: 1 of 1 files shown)


lrusak:
I don't mind adding this, it adds < 1MB, however, it's not something many people will use and it really doesn't suit the core purpose of OpenELEC.
```
-rwxr-xr-x. 1 lukas lukas 493K Dec 28 23:23 build.OpenELEC-Generic.x86_64-7.0-devel/systemd-228/.install_pkg/usr/bin/systemd-nspawn

-rwxr-xr-x. 1 lukas lukas 392K Dec 28 23:23 build.OpenELEC-Generic.x86_64-7.0-devel/systemd-228/.install_pkg/usr/bin/machinectl
```

I have used this as an add-on for a while now. see, https://github.com/lrusak/lrusak-openelec-addons/blob/master/addons/tools/systemd-tools/package.mk

This add-on doesn't include machinectl (as it isn't really needed) however, we could always enable building machinectl and just removing the binary then including it in the add-on.

chewitt:
@lrusak - please update on whether this is something we want in core builds? (merge or close)

lrusak:
I'd rather not have this in core OE.

I have built this as an add-on and it works fine.

PR #4541, opened 2015-12-29
RPi/RPi2: add docker support to kernel

lrusak:
Ok so this will update the RPi and RPi2 to be able to use docker. The Generic and Nvidia_Legacy projects already have the necessary configs to support docker. WeTek_Play and imx6 use too old of kernels to be able to use docker because we will be using overlayfs.

I used this script to check the configs, https://github.com/docker/docker/blob/master/contrib/check-config.sh

I have only enabled the configs that are listed as "Generally Necessary" and none of the configs listed as "Optional Features".
before http://sprunge.us/KIGH
after http://sprunge.us/STLV
Use ```curl``` in a terminal with colours enabled to see the output better.

I would appreciate @popcornmix and @MilhouseVH to comment on this if there are any config options that are dangerous.

I still have to runtime test this on the RPi to make sure docker will run with only these options, but the check-config.sh script provides a good starting point.


popcornmix:
I've had a look and I believe these options match what we have in raspbian, so I'm happy there is no significant performance issue with these options.
I don't know if this makes docker usable in OE, but similar options did work in raspbian.
But no objection here.

MilhouseVH:
If @popcornmix is happy, then I'm happy too.

PR #4545, opened 2015-12-30
jsoncpp: update package to 1.6.5

chewitt:
I used this when building official addons for ATV a couple of months ago without problems - package is uploaded to the sources server.

### DIFF
--- a/packages/textproc/jsoncpp/package.mk
+++ b/packages/textproc/jsoncpp/package.mk
@@ -17,7 +17,7 @@
 ################################################################################
 
 PKG_NAME="jsoncpp"
-PKG_VERSION="src-0.5.0"
+PKG_VERSION="1.6.5"
 PKG_REV="1"
 PKG_ARCH="any"
 PKG_LICENSE="GPL"
--- a/tools/mkpkg/mkpkg_jsoncpp
+++ b/tools/mkpkg/mkpkg_jsoncpp
@@ -1,0 +1,44 @@
+#!/bin/sh
+
+################################################################################
+#      This file is part of OpenELEC - http://www.openelec.tv
+#      Copyright (C) 2009-2012 Stephan Raue (stephan@openelec.tv)
+#
+#  This Program is free software; you can redistribute it and/or modify
+#  it under the terms of the GNU General Public License as published by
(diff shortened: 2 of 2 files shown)


stefansaraev:
PKG_URL="https://github.com/open-source-parsers/jsoncpp/archive/$PKG_VERSION.tar.gz"

you dont need a mkpkg_script

stefansaraev:
and you should probably be using 0.10.x. not 1.x

chewitt:
I saw, but then the sources don't download with $PKG_NAME in the filename - or there is some new renaming voodoo I am missing to make that happen?

stefansaraev:
see systemd.

MilhouseVH:
systemd has the same problem as tvheadend, that #4537 tries to address.

stefansaraev:
PKG_SOURCE_NAME is not the right way. you should aim to clean up. not add even more hacks.

scripts/extract should go (it is basicaly doing something like "brute-force" on pkg name / ext now, while all the info required to extract is already available). see my scripts/unpack. packages with more than one PKG_URL (only makemkv in unofficial) should also go away.

my 2 cents.

MilhouseVH:
Your 2 cents are always welcome although I'm not sure how your scripts/unpack solves the issue (I took a quick look in your K+++ branch) when the problem starts in scripts/get, as this is where we download files with potentially non-standard filenames that then can't easily be hosted on our mirror servers (assuming we want to identify them in future), and that also require custom (although most likely boiler-plate) `unpack()` functions. Normalizing the names of the files we download would solve both issues.

I'm certainly not saying #4537 is the best solution, or even the right solution, but if there is a better solution we should probably discuss it there. :)

stefansaraev:
in my K+++ branch I dont need unpack() functions. at all. even for systemd. unless a package has really stupid packaging. there is no such thing like "standard" filenames. but well. this is offtopic, do whatever is easier for you ;)

stefansaraev:
I runtime tested jsoncpp 1.6.5 with pvr.stalker (I use it daily) and everything works as before. so the bump  seems safe. can't speak of other pvr clients that use jsoncpp. just fyi.

PR #4548, opened 2015-12-30
readline: use http download

chewitt:
use https instead of ftp to avoid download restrictions in certain locations (e.g. work)


fritsch:
You know that https is blocked in ~ 50% of the world? or certificates highjacked? http://www.blockedinchina.net/?siteurl=https%3A%2F%2Fgoogle.de

chewitt:
we can use http then .. rebased

PR #4557, opened 2016-01-01
backport 6.0: Update cert bundle and new bundle method.

lrusak:
:+1: 

This new method has been successful so far with @MilhouseVH builds

Unless anyone objects this should go in 6.0.1


MilhouseVH:
:+1: 

The current certs in openelec-6.0 branch are just going to cause a world of hurt if we release them as they are.

PR #4564, opened 2016-01-02
make possible run Openelec as guest on KVM

piotrasd:
when we add this modules we will able to run Openelec with KVM driver as guest on virtual system
tested and working (changes based on http://www.linux-kvm.org/page/Tuning_Kernel)

Without this modules we must emulate LAN driver, SATA driver, etc. :( please consider add this to Generic image
More about testing etc. https://lime-technology.com/forum/index.php?topic=44683.0


CHBMB:
I'd like to see this too.

lrusak:
:+1: 

is anything else needed? graphics drivers, etc? or will the Generic image run via kvm with just these changes?

piotrasd:
no just this, because we mapping hardware to virtual (example like Nvidia GT730 or sound Realtek etc)
we dont need anything more ;)

lrusak:
Ah, so you're just using PCI passthrough? then?

I'm ok with this, will leave it open for a day or so and if there are no objections I will merge.

piotrasd:
yes sorry for VIA ... this you can disable

packages/sysutils/busybox/scripts/init
```diff
@@ -1,4 +1,4 @@
-#!/bin/sh
+!/bin/sh
```
stefansaraev: no.

MilhouseVH:
Presumably this causes no performance overhead for non-Virtual users, ie. existing users won't notice any difference?

piotrasd:
@MilhouseVH im used one source after this patches for my Virtu Openelec and for Zotac ID42 - and im dont notice any performance issue, after this changes on normal x86 machine (zotac id42)

MilhouseVH:
Ok just thought I'd ask - would be nice to have some metrics/numbers before/after this change although I'm not entirely sure of the best way to get those.

piotrasd:
@MilhouseVH no problem give me some idea how i can get this numbers :) some script or benchmark to test and i will do.

I think all added modules when OE is starting on x86 machine are even not loaded, (maybe only 9p for mount shares but this nothing to do with CPU etc.)
im almost 100% sure this not involve in any performace changes

MilhouseVH:
I guess iperf for network, no idea how to test video performance other than comparing performance of some really "hard" videos (is cpu load the same, is it as smooth with same dropped/missed frames as before etc.). You could spend hours on this though and if you're sure the modules aren't loaded except when required then that sounds good enough for me.

piotrasd:
i saw only one in dmesg on zotac - 9p - for mount shares from host (but this is like any other filesystem, NFS etc.)

johnodon:
I also would like to see these changes implemented.  As far as I am aware, there would be no impact to baremetal OE users.

I also agree with piotrasd regarding 9p.  I am one of the users who saw horrible speeds while trying to use this protocol.  It was so bad that SMB/NFS was 15x - 20x faster.  However, many other users have not experienced these issues.

I am currently running OE v6 final in 3 different VMs where each has a Radeon HD6450 passed through.  They run flawlessly, however, due to the absence of virtio I must force the VMs to use SATA and e1000 protocols for HD controller and NIC respectively.  While this works fine, it is not an "out-of-the-box" solution.  It would be preferable to use virtio in this case.

John

lrusak:
@sraue any other objections here?

sraue:
is this 9p change needed (here) i dont like it (here) other then that (the kernel config) can go in (but only after rebasing)

sraue:
also i would like to have thin in nvidia:legacy project too to keep it in sync until we remove this finally

johnodon:
I agree that 9p should be excluded at this point as it is still too experimental and buggy.

piotrasd:
in new qemu 2.5 and kernel 4.4 i read they improve this (9p) but no problem if you dont want this, most important for us are modules for KVM, rebase you mean for kernel 4.4rc8 ? (ok i will delate mount command from init and rebase kernel config)

chewitt:
@piotrasd please squash commits

lrusak:
@piotrasd please drop @1721dbdf33d8628f3be1d2f2a6f28771b210e533 then rebase and squash commits

piotrasd:
I cannot rebase because all time I get info about modified bin firmwares from ati :/ I don't know why, and I'm getting this info even after fresh clone

lrusak:
@piotrasd see here for info about kernel config changes, http://wiki.openelec.tv/index.php/Development_workflow#Kernel_Config

chewitt:
@piotrasd - I'm guessing you're also a Mac user? - git clone pulls down "blah.bin" and "BLAH.bin" to the same filesystem folder/directory and since MacOS uses case-insensitive HFS+ the second file overwrites the first and git sees this as a change. To solve this you need to use a case sensitive filesystem, i.e. re-partition the boot disk to create space for a separate case-sensitive HFS+ (HFSX) volume to clone into, or (much easier and recommended) use a Linux VM for OE development things.

sraue:
@lrusak can you take over this, and create a new PR only with the kernel config changes? so this could be closed?

piotrasd:
is possible to change virtio-blk module to builtin ??
CONFIG_VIRTIO_BLK=y

i have problem run OE on KVM with virtio driver for disk (im still using SATA emulation)
even if i add .config/modules-load.d/virtio-blk.conf for load on start still im unable to boot :(

sraue:
please dont answer to closed issues, they will not watched!

PR #4567, opened 2016-01-03
fakeroot: Make build_dir specific, not project specific

MilhouseVH:
When using `BUILD_SUFFIX` (#3791) to build to an alternate build directory, it's possible to build the same project concurrently with `DEBUG=Y` and `BUILD_SUFFX=debug`, and with `DEBUG=N` and `BUILD_SUFFIX=`, saving a lot of time as 2 concurrent builds will complete in the same time as a single build (if you have enough cores).

However, the filename used by the fakeroot script is `$ROOT/.fakeroot.$PROJECT.$TARGET_ARCH`, eg. `./fakeroot.RPi2.arm`, which is not good news when building the same project to separate build directories (ie. `build.OpenELEC-RPi2.arm-7.0-devel` and `build.OpenELEC-RPi2.arm-7.0-devel-debug`) as the same fakeroot script file could be used - or overwritten - by either build.

Consequently it would be better if the name of the fakeroot script file is based on the build directory to which it corresponds, and not simply the project and target_arch.


lrusak:
Seems sane to me :+1:

PR #4568, opened 2016-01-03
[get]: Avoid concurrent downloads of the same package

MilhouseVH:
This might be a little contentious, but it's a problem I see quite often as I run parallel/concurrent builds.

The change in this PR will block any build that is trying to download a package that is already in the process of being downloaded by another build.

Without this PR, partial (ie. corrupted) downloads are likely when building in parallel.


vpeter4:
I was using similar approach in the past. Good to see this in.

PR #4441, opened 2015-11-15
Use pkg-config mechanism to synchronize compile flags and paths between VDR and VDR plugins instead of tons of patches.

hsteinhaus:
Doing so allows to use the default Makefiles supplied by VDR plugins without any patching. This is a big gain in maintainability, as a plugin update now usually done by incrementing the version number in the package.mk file - no more manual patch rectification after plugin makefile changes needed. 

Additionally, the pkg-config mechanism ensures absolutely consistent compile flags between VDR and plugins. Inconsistent flags are often a source for unloadable VDR plugins due to symbol resolve issues (and a reason for more OpenELEC plugin patches).

PR #4573, opened 2016-01-04
vdr-addon: fix some vdr plugins

glenvt18:
After [PR#4441](https://github.com/OpenELEC/OpenELEC.tv/pull/4441) vdr-epgsearh, vdr-plugin-epgfixer and vdr-live are broken. They are built without errors but fail to load on target. Their makefiles use XXXX-config (e.g. pcre-config) scripts instead of pkg-config and if a XXXX:host package is built XXXX-config outputs paths for the host libraries. As the result the plugins above are linked (dynamically) against the host libraries (can be checked with ldd) and failed to load. 
The proposed solution allows to run the host XXXX-config when building for a host and the target one when building for a target (with /usr/include and /usr/lib stripped).

### DIFF
--- a/packages/3rdparty/lib/cxxtools/package.mk
+++ b/packages/3rdparty/lib/cxxtools/package.mk
@@ -44,10 +44,39 @@ pre_configure_target() {
 
 
 post_makeinstall_host() {
-  rm -rf $TOOLCHAIN/bin/cxxtools-config
+  mv $ROOT/$TOOLCHAIN/bin/cxxtools-config $ROOT/$TOOLCHAIN/bin/cxxtools-config.host
+
+  cat >$ROOT/$TOOLCHAIN/bin/cxxtools-config <<EOT
+#!/bin/sh
+case "\$1" in
+--c*flags*|--libs*)
+$ROOT/$TOOLCHAIN/bin/cxxtools-config.\$DESTIMAGE "\$@" | xargs -n1 | grep -v -e^-I/usr/include\\$ -e^-L/usr/lib\\$ | xargs
+;;
+*)
+$ROOT/$TOOLCHAIN/bin/cxxtools-config.\$DESTIMAGE "\$@"
+;;
+esac
+EOT
+
+  chmod 0755 $ROOT/$TOOLCHAIN/bin/cxxtools-config
 }
 
 post_makeinstall_target() {
-  rm -rf $SYSROOT_PREFIX/usr/bin/cxxtools-config
   rm -rf $INSTALL/usr/bin
+
+  mv $SYSROOT_PREFIX/usr/bin/cxxtools-config $ROOT/$TOOLCHAIN/bin/cxxtools-config.target
+
+  cat >$ROOT/$TOOLCHAIN/bin/cxxtools-config <<EOT
+#!/bin/sh
+case "\$1" in
+--c*flags*|--libs*)
+$ROOT/$TOOLCHAIN/bin/cxxtools-config.\$DESTIMAGE "\$@" | xargs -n1 | grep -v -e^-I/usr/include\\$ -e^-L/usr/lib\\$ | xargs
+;;
+*)
+$ROOT/$TOOLCHAIN/bin/cxxtools-config.\$DESTIMAGE "\$@"
+;;
+esac
+EOT
+
+  chmod 0755 $ROOT/$TOOLCHAIN/bin/cxxtools-config
 }
--- a/packages/3rdparty/web/tntnet/package.mk
+++ b/packages/3rdparty/web/tntnet/package.mk
@@ -50,7 +50,41 @@ PKG_CONFIGURE_OPTS_TARGET="--disable-unittest \
                            --with-ssl=no \
                            --with-stressjob=no"
 
+post_makeinstall_host() {
+  mv $ROOT/$TOOLCHAIN/bin/tntnet-config $ROOT/$TOOLCHAIN/bin/tntnet-config.host
+
+  cat >$ROOT/$TOOLCHAIN/bin/tntnet-config <<EOT
+#!/bin/sh
+case "\$1" in
+--c*flags*|--libs*)
+$ROOT/$TOOLCHAIN/bin/tntnet-config.\$DESTIMAGE "\$@" | xargs -n1 | grep -v -e^-I/usr/include\\$ -e^-L/usr/lib\\$ | xargs
+;;
+*)
+$ROOT/$TOOLCHAIN/bin/tntnet-config.\$DESTIMAGE "\$@"
+;;
+esac
+EOT
+
+  chmod 0755 $ROOT/$TOOLCHAIN/bin/tntnet-config
+}
+
 post_makeinstall_target() {
   rm -rf $INSTALL/usr/bin
   rm -rf $INSTALL/usr/share
+
+  mv $SYSROOT_PREFIX/usr/bin/tntnet-config $ROOT/$TOOLCHAIN/bin/tntnet-config.target
+
+  cat >$ROOT/$TOOLCHAIN/bin/tntnet-config <<EOT
+#!/bin/sh
+case "\$1" in
+--c*flags*|--libs*)
+$ROOT/$TOOLCHAIN/bin/tntnet-config.\$DESTIMAGE "\$@" | xargs -n1 | grep -v -e^-I/usr/include\\$ -e^-L/usr/lib\\$ | xargs
+;;
+*)
+$ROOT/$TOOLCHAIN/bin/tntnet-config.\$DESTIMAGE "\$@"
+;;
+esac
+EOT
+
+  chmod 0755 $ROOT/$TOOLCHAIN/bin/tntnet-config
 }
--- a/packages/devel/pcre/package.mk
+++ b/packages/devel/pcre/package.mk
@@ -44,6 +44,24 @@ PKG_CONFIGURE_OPTS_TARGET="--disable-shared \
              --enable-unicode-properties \
              --with-gnu-ld"
 
+post_makeinstall_host() {
+  mv $ROOT/$TOOLCHAIN/bin/pcre-config $ROOT/$TOOLCHAIN/bin/pcre-config.host
+
+  cat >$ROOT/$TOOLCHAIN/bin/pcre-config <<EOT
+#!/bin/sh
+case "\$1" in
+--c*flags*|--libs*)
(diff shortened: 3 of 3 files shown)


vpeter4:
How does it fail to load?

Because I build the addon one week back for imx6 project. After addon installed I enabled all the plugins in settings and didn't see any error. How can I be sure plugins are really loaded?

CvH:
Could you specify *but fail to load on target* ?
Live works like exacted at RPi2.

glenvt18:
Fails to load means loading a plugin's shared library fails like this:

`Jan 04 20:30:22 OpenELEC vdr[6270]: [6270] loading plugin: /storage/.kodi/addons/service.multimedia.vdr-addon/plugin/libvdr-epgfixer.so.2.2.0`

`Jan 04 20:30:22 OpenELEC vdr[6270]: [6270] ERROR: libpcre.so.1: cannot open shared object file: No such file or directory`

libpcre.so.1 is linked dynamically (against host libpcre):

`OpenELEC:~/.kodi/addons/service.multimedia.vdr-addon/plugin # ldd libvdr-epgfixer.so.2.2.0`
`    linux-vdso.so.1 (0x00007ffd05b0c000)`
`    libpcre.so.1 => not found`
`    libstdc++.so.6 => /usr/lib/libstdc++.so.6 (0x00007f30367f3000)`
`    libgcc_s.so.1 => /usr/lib/libgcc_s.so.1 (0x00007f30365dd000)`
`    libc.so.6 => /lib/libc.so.6 (0x00007f303623e000)`
`    libm.so.6 => /lib/libm.so.6 (0x00007f3035f3e000)`
`    /lib64/ld-linux-x86-64.so.2 (0x00005569c449e000)`

but it should have been linked statically against target libpcre

`OpenELEC:~/.kodi/addons/service.multimedia.vdr-addon/plugin # ldd libvdr-epgfixer.so.2.2.0.fixed`
`	linux-vdso.so.1 (0x00007fff4e18d000)`
`	libstdc++.so.6 => /usr/lib/libstdc++.so.6 (0x00007fc5bea1a000)`
`	libgcc_s.so.1 => /usr/lib/libgcc_s.so.1 (0x00007fc5be804000)`
`	libc.so.6 => /lib/libc.so.6 (0x00007fc5be465000)`
`	libm.so.6 => /lib/libm.so.6 (0x00007fc5be166000)`
`	/lib64/ld-linux-x86-64.so.2 (0x0000555f5bcbc000)`

Was built for x86_64, oe-6.0 with a clean build tree:
`PROJECT=Generic ARCH=x86_64 scripts/create_addon vdr-addon`

glenvt18:
@vpeter4 Let me explain what happens at my build.
1. vdr-addon is built from scratch.
2. vdr-live is listed before vdr-epgsearch and vdr-plugin-epgfixer in the vdr-addon target dependency list and is built before them.
3. vdr-live has both pcre:host and pcre:target as dependencies and both are built. pcre:host installs pcre-config into toolchain/bin, pcre:target installs pcre-config into sysroot/usr/bin.
4. When vdr-epgsearch and vdr-plugin-epgfixer are being built (for target) toolchain/bin is prepended to PATH and make executes toolchain/bin/pcre-config which outputs host options like `-L...toolchain/lib -lpcreposix -lpcre` that is linking with the host libraries.

What is proposed.
Some packages like curl, cxxtools, freetype, libpng16, libpng, pcre, tntnet, xml2, xslt and maybe some other offer PKG_NAME-config scripts as an alternative to pkg-config and some other packages use those PKG_NAME-config scripts in their makefiles. For those packages (like pcre) that are meant to support both host and target builds different PKG_NAME-config scripts must be called depending on the build target - "host" or "target" (those PKG_NAME-config scripts though always run on host - we don't build anything on target). Each time a host or target instance of a package is installed a wrapper toolchain/bin/PKG_NAME-config script is created which runs either host or target PKG_NAME-config depending on $DESTIMAGE which is either "host" or "target". As it is not known which instance (host or target) will be installed first, the same wrapper script is created each time. Original PKG_NAME-config scripts are saved as PKG_NAME-config.host and PKG_NAME-config.target.

As an alternative solution we could prepend sysroot/usr/bin to PATH when building for target, but I'm not sure it is safe to do it.

BTW. When you build epgsearch or epgfixer you should also see `-DPCRE_STATIC` in the compiler command line if you use the right pcre-config. Could you check it pls.

sraue:
see https://github.com/OpenELEC/OpenELEC.tv/blob/master/packages/print/freetype/package.mk#L49 and line 50, this should be the way to work with the PKG_NAME-config packages. i just checked some of the above (and also sdl2-config) are wrong will try to fix

vpeter4:
glenvt18, there are some issues and I'm checking it. The problem is something like sraue proposed. But I think this should be done from build or install script like for [.la files](https://github.com/OpenELEC/OpenELEC.tv/blob/master/scripts/build#L364) and not packages itself - it would be fixed in one place for all packages.

glenvt18:
@srau This is ok with freetype:target as long as no package has freetype:host as a dependency. A :host package would also want to have it's XXXX-config in toolchain/bin. We have to isolate such (dual host/target) packages and fix them when a :host instance is requested be some other package's package.mk. Or create a universal wrapper script with fixes for particular packages (and symlink to it). Don't know which one is better.

@vpeter Still wonder how you managed to build yours. How about building from a chrooted enviroment?

vpeter4:
I was building Generic x64 on i386 host. I saw no problem and also plugin was loaded.
But today (or yesterday) I did something to my build and need to rebuild everything which will take some hours. Will write my findings.
Maybe you should come to IRC for immediate discussion.

glenvt18:
@vpeter I'm at IRC.

vpeter4:
So my build was finished. The conclusion is that because I'm building 64 bit addon on 32 bit host the pcre library from host build is just skipped because it is incompatible:
> warning: skipping incompatible /home/me/OpenELEC.tv.master.5.1.2016/build.OpenELEC-Generic.x86_64-7.0-devel/toolchain/lib/libpcreposix.so while searching for pcreposix

That's why I got working addon with all the plugins.

Not sure what is the best solution here. Maybe sraue will create something.

glenvt18:
Nothing odd here. And it should have been skipped while building for freescale too (was it?).
Could you pls run this from your OpenELEC/ dir:
```
objdump -Tt build*/vdr-live-0.3.0/libvdr-live.so | grep pcre
objdump -Tt build*/vdr-plugin-epgfixer-1ad4aaf/libvdr-epgfixer.so | grep pcre
ldd build*/addons/vdr-addon/service.multimedia.vdr-addon/plugin/*
```
As for the proposed solution - yes, its maybe not the most elegant one, it just does its job. Each package is free to supply its own XXXX-config. There is no standard for it. So, still some manual work is needed. Other options could be to have a function like `install_config_script` and call it from `post_makeinstall_xxxx` in a package.mk or to have `PKG_CONFIG_SCRIPT=pcre-config` and handle it in the build script or look for $PKG_NAME-config files in the bin dirs and handle it automatically (is it relieble?). Any other suggestions?

vpeter4:
Sure it is odd building for target but you are looking for libraries in toolchain/lib - this is for host builds.
I can't check for imx6 because don't have any master build (only 6.0).

http://pastebin.com/mT6BV2V7

Other suggestions - few of them. But I will wait sraue (don't like duplicate work like in the past).

glenvt18:
@vpeter4 Thanks. The dumps look ok.

```
build.OpenELEC-Generic.x86_64-7.0-devel/addons/vdr-addon/service.multimedia.vdr-addon/plugin/libvdr-iptv.so.2.2.0:
    libcurl.so.4 => not found

build.OpenELEC-Generic.x86_64-7.0-devel/addons/vdr-addon/service.multimedia.vdr-addon/plugin/libvdr-satip.so.2.2.0:
    libcurl.so.4 => not found

build.OpenELEC-Generic.x86_64-7.0-devel/addons/vdr-addon/service.multimedia.vdr-addon/plugin/libvdr-xmltv2vdr.so.2.2.0:
    libxml2.so.2 => not found
```
Do those plugins start without errors?  You can also check
```
ldd /storage/.kodi/addons/service.multimedia.vdr-addon/plugin/*
```
on target (more reliable).

Ok. Will wait for sraue. As for me, I would suggest PKG_CONFIG_SCRIPT=xxxx-config solution. It leaves some space for manual control. It also requires fixing some packages, but there are not that many of them. Also, copying or moving xxxx-config files from sysroot/usr/bin to toolchain/bin is not a good idea generally - there can be several target sysroots and only one toolchain/bin.

vpeter4:
ldd looks for libraries on a host I think. And I don't have libcurl.so.4 there.
But libcurl.so.4 is inside openelec image:
> $ find build*/image/system/ -name libcurl.so.4
> build.OpenELEC-Generic.x86_64-7.0-devel/image/system/usr/lib/libcurl.so.4

glenvt18:
It's fine. curl is supposed to be linked dynamically. It should work. So, because of the host/target architecture mismatch magic you have a working vdr-addon:)

stefansaraev:
those should be patched NOT to use *-config. back in time, when vdr switched to "new way" for building plugins, I did not like it and patched all plugins. this has recently changed (check git log)

actualy, everything that uses *-config should die in pain. my 2 cents.

vpeter4:
So the solution is simple (as it was before): use patch only to change calls from xxx-config to pkg-config.

stefansaraev:
I would revert changes to epgfixer and epgsearch and remove live and tntnet. live was never worth the trouble.

glenvt18:
At least this solution (patching) is the most reliable one.
@stefansaraev May be it's just enough to change xxxx-config to pkg-config than reverting a whole commit? 
Also
vdr-iptv: curl-config
vdr-plugin-restfulapi: cxxtools-config
Removing vdr-live - no webinterface at all?

sraue:
if we can change the way to pkg-config (*.pc) this would be the best solution, $PKG_NAME-config should die .... here i agree with @stefansaraev . i prefer to patch this packages so they work, the solution for $PKG_NAME-config i have in mind is more complex and will need some time... in the worst case we should revert the PR from before

glenvt18:
@sraue The dilemma here is that one have either to examine each package if it provides -config script and fix it or to examine if it uses -config and patch its Makefile. So, anyway, some manual work is needed. It looks like there are more packages that provide than those that use -config scripts.

glenvt18:
Besides makefiles there are things like these:
http://www.gnu.org/software/autoconf-archive/ax_path_generic.html
http://www.gnu.org/software/autoconf-archive/ax_lib_curl.html (see [source](http://git.savannah.gnu.org/gitweb/?p=autoconf-archive.git;a=blob_plain;f=m4/ax_lib_curl.m4))

PR #4574, opened 2016-01-04
projects/imx6/linux: rename Hummingboard on-board analog sgtl5000 audio codec

vpeter4:
In device tree there is
> model = "On-board Codec";

and this doesn't match any alsa config. That's why it is renamed to
> model = "imx-sgtl5000";

for which we already have imx-sgtl5000.conf file.
This enables to use analog sound output on Hummingboard.


sraue:
same for oe6 too?

vpeter4:
Yes, same for OE6.

PR #4577, opened 2016-01-04
RPi/RPi2: final docker options

lrusak:
```CONFIG_BLK_CGROUP``` was listed as optional but when runtime testing it I got
> Error response from daemon: Cannot start container System error: mountpoint for blkio not found

```CONFIG_EXT4_FS_POSIX_ACL``` and ```CONFIG_EXT4_FS_SECURITY``` are needed as well otherwise the fs layers cannot even be pulled.

I seem to also run into this issue https://github.com/docker/docker/issues/18125 every once in a while, but otherwise docker containers run after these changes.

I think this changes are harmless but one last ping for @popcornmix to take a look


popcornmix:
We've got those options in raspbian, so no objections here.

lrusak:
:+1: Docker support is final!

PR #4536, opened 2015-12-27
hid-hbgic-kbd: fixed key response values on some WeTekPlay RC buttons

justason:
That will fix the behavior of 3 buttons on the WetekPlay RC, which were disabled recently

Issue #4428, opened 2015-11-06
after commit: on 24.10.2015, SHA:2b6087dcc4d9dc0f7c5812f750152dbad8a3b65a, several buttons on WetekPlay are not working


justason:
Problem is in the patch file: projects/WeTek_Play/patches/linux/linux-054-hbgic-remote-0.2.patch
After applying, at least "DisplayRatio" button on RC is acting as "Back" button.
Removing the patch restores the normal operation of the Wetek Play remote.

sraue:
then the remote for the OE Box is wrong mapped... the problem is both systems are using the same RF receiver, and both remotes are printed different... to solve this we must find a way which works for both. at least you are able to remap via keyboard.xml

justason:
Not really.
The Kodi is using values, that are coming from the driver.
So altering the keyboard.xml is useless.

I have proposed a solution, that possibly will solve the issue:
case 0x0007002a: hbgic_map_key(KEY_F2); break;
case 0x0007003a: hbgic_map_key(KEY_F1); break;

that will utilize the "missing" buttons on a wetek remote.

chewitt:
@justason do you have an update on your proposed fix? - we didn't see a PR yet

justason:
cannot "git push":
git push origin linux-054-hbgic-remote-patch-wetek-fix
Username for 'https://github.com': justason
Password for 'https://justason@github.com':
remote: Permission to OpenELEC/OpenELEC.tv.git denied to justason.
fatal: unable to access 'https://github.com/OpenELEC/OpenELEC.tv.git/': The requested URL returned error: 403

chewitt:
It looks like you are attempting to submit directly to our repo where you (correctly) don't have commit rights. You need to fork and clone your own fork to PR from - not clone our repo.

chewitt:
I'm closing this ticket down. Further discussion can take place on https://github.com/OpenELEC/OpenELEC.tv/pull/4536. Thanks

PR #4578, opened 2016-01-05
Oe6 update

sraue:


sraue:
this fixes (in a different way) #https://github.com/OpenELEC/OpenELEC.tv/pull/4536 and https://github.com/OpenELEC/OpenELEC.tv/issues/4428

PR #4500, opened 2015-12-16
kodi: update to 17.0

lrusak:
As far as I know nothing else needs to be added/dropped. I will update as things change.

This can stay as a placeholder until openelec-7.0 is branched

PR #4581, opened 2016-01-05
platform: rename to p8-platform

lrusak:
# Do not merge yet!

Not sure if this is safe yet as libcec and kodi-platform may have to be updated as well.

I think the binary addons only depend on kodi-platform, not p8-platform. so I think it will be ok but I will wait till I do a full build with addons.


MilhouseVH:
Yes you'll need latest libcec to use p8-platform. You'll then need this patch: https://gist.github.com/MilhouseVH/315ce8767085f7e9a28d

I'd probably have moved the kodi/p8-platform to devel/p8-platform as it's not really a kodi package, is it?

The kodi binary addons also need to be updated to use p8-platform - this is yet to happen.

lrusak:
This could be a pain. we may have to run p8-platform and platform side by side for a while till all packages are updated.

sraue:
@MilhouseVH our grouping is only informative, packes can be boved to other folders, if yo think some packages are wrong grouped you can change and send PRs

MilhouseVH:
I'll run it past @lrusak tomorrow - this might be a good time to clean up the mediacenter folder (even if only for OCD reasons). No big deal though, and can probably happen after this PR, if at all.

MilhouseVH:
With latest `libcec`, we don't need:

packages/devel/libcec/patches/libcec-01-add-repeating-keypress.patch
packages/devel/libcec/patches/libcec-02-fix-crash-on-some-systems.patch

as they're already merged in libcec master, respectively:

https://github.com/Pulse-Eight/libcec/pull/111
https://github.com/Pulse-Eight/libcec/pull/133

lrusak:
closing, merged into #4500

PR #4582, opened 2016-01-07
Pkg config script fixes

sraue:
This is a first step to fix #4573 after #4441:

$PKG_NAME-config for target should be in $SYSROOT_PREFIX/usr/bin and not in $TOOLCHAIN/bin, there should be only the $PKG_NAME-config scripts for hostbuilds.

As a second step all addons which require this scripts must be patched to use pkg-config (*.pc) which should work already or to use the scripts from $SYSROOT_PREFIX/usr/bin with using something like ```$SYSROOT/$prefix/bin|lib|include``` instead ```$prefix/bin|lib|include``` and set SYSROOT on target builds fin the package.mk file


sraue:
should be backported to OpenELEC-6.0

sraue:
with the last commits i changed (hopefully) all vdr-packages to use pkg-config instead. it also adds a patch to tntnet from tntnet-2.3rc1 to add pkg-config support. most PKG_NAME-config scripts are not used from mainbuild and vdr-* addons. as a next step we should try to remove them and rebuild all....

sraue:
no, forgot some packages :-) will add to the PR asap

glenvt18:
@sraue Tried to build on oe-6.0. tntnet fails - http://pastebin.com/eHVyEJTC

sraue:
@glenvt18 will look in this

vpeter4:
I removed hunk #5 in file tntnet-pkgconfig-support.patch and it builds.

sraue:
@glenvt18 please try again, updated the tntnet commit

glenvt18:
@sraue tntnet builds ok now as well as the vpeter4's version. Looks like other things work too. Need some time to rebuild - will report asap.
BTW. If you change
`+LIBS += $(shell pkg-config --libs cxxtools cxxtools-http)`
to
`+LIBS += $(shell pkg-config --libs cxxtools-http cxxtools)`
in `vdr-plugin-restfulapi-pkgconfig.patch` changes introduced in e93f4b5887c18e7c67f3b67636b19c2092f85868 will not be needed.

sraue:
@glenvt18 updated the pkgconfig patch and buildscript for vdr-plugin-restfulapi, please test and report back

glenvt18:
@sraue vdr-live won't start - http://pastebin.com/2HULJKnt. Looks like there is no `-l pcre`. Investigating...

sraue:
huh, found another c/p error in this patch and fixed.... can you try again before i add your change? (simply the vdr-live commit and rebuild vdr-live)

glenvt18:
Works (with my change).  I also changed package.mk like this
```
--- package.mk.orig	2016-01-08 05:45:35.478888448 +0300
+++ package.mk	2016-01-08 05:37:41.555781755 +0300
@@ -34,9 +34,9 @@
 PKG_AUTORECONF="no"
 
 pre_configure_target() {
-  export CFLAGS="$CFLAGS -fPIC -L$SYSROOT_PREFIX/usr/lib -L$SYSROOT_PREFIX/lib"
-  export CXXFLAGS="$CXXFLAGS -fPIC -L$SYSROOT_PREFIX/usr/lib -L$SYSROOT_PREFIX/lib"
-  export LDFLAGS="$LDFLAGS -fPIC -L$SYSROOT_PREFIX/usr/lib -L$SYSROOT_PREFIX/lib"
+  export CFLAGS="$CFLAGS -fPIC"
+  export CXXFLAGS="$CXXFLAGS -fPIC"
+  export LDFLAGS="$LDFLAGS -fPIC"
 }
 
 pre_make_target() {
```
to get rid of -LXXX  (see http://pastebin.com/Hg02yUfv). We don't need -L...sysroot/usr/lib here even for the linker as gcc is configured with the "the right" sysroot.

sraue:
added the pcre and *FLAGS changes, please test and report back

vpeter4:
Seems ok by me.

glenvt18:
@sraue Already tested - everything's fine.

sraue:
so good to merge? @stefansaraev

stefansaraev:
I dont use most of this stuff now. if it works for you, merge away ;)

MilhouseVH:
I'm assuming this is responsible for the clean build issues I'm now experiencing when building RPi2 on Ubuntu x86_64.

This is the first clean build in a few days but now I'm experiencing multiple failures due to packages or libraries not being found.

Initially, `libaacs` and `libbdplus` failed to build due to `libgcrypt` and `gpg-error` prefix issues - these where fixed with this patch: http://sprunge.us/hFNS

I then ran into an issue when building `libhid` - it's looking for `libusb.so` in the wrong location:
```
/usr/lib/x86_64-linux-gnu/libusb.so: file not recognized: File format not recognized
```
I've tried adding `-L$SYSROOT_PREFIX/usr/lib` to `LDFLAGS` in `package.mk`, but it's made no difference.

tail end of build log: http://sprunge.us/hFWH
libhid `configure`: http://sprunge.us/GQhh
libhid `config.log`: http://sprunge.us/HOgY
libhid `libhid.pc`: http://sprunge.us/GiED

Right now I'm stuck, so any suggestions welcome. I'm happy to PR the accumulated fixes once a complete build is achieved.

PR #4583, opened 2016-01-07
cosmetics: rename 'unofficial' to 'community' builds

chewitt:
builds created by users in our ecosystem/community should be labelled as such .. it's a friendlier way of saying they are not official builds

### DIFF
--- a/config/functions
+++ b/config/functions
@@ -282,7 +282,7 @@ show_config() {
 
   dashes="==========================="
   config_message="$config_message\n $dashes$dashes$dashes"
-  config_message="$config_message\n Configuration for $DISTRONAME ($([ "$OFFICIAL" = "yes" ] && echo "official" || echo "unofficial"))"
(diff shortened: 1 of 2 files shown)


sraue:
maybe "Community Edition"?

MilhouseVH:
"Edition" doesn't really make much sense in this context, IMHO.

Personally I don't have a problem with "unofficial", but I guess the problem being solved here is how the builds are perceived rather than how they are presented - I can't imagine any builders have an issue with the "unofficial" tag, they are after all not official builds.

"community" is fine by me.

chewitt:
I've typically seen "Community Edition" used to distinguish an OSS release from a commercial non-OSS version so "community" on it's own is enough to label the build (as unofficial has). This is all about how things are perceived, and part of a plan to better engage with and harness the efforts of the wider builder community around us.

PR #4594, opened 2016-01-09
connman: update to 1.31

chewitt:
connman 1.31 adds a new connman-wait-online binary which is built/embedded but not used since we provide our own .service files and cm-online script. This new capability needs to be further explored by someone who understands our systemd sequencing and can read the new code. Reading connman documentation on this I suspect connman-wait-onilne addresses some long-standing network timing and sequencing cases that we see.

https://01.org/connman/blogs/pflykt/2015/connman-1.31 <= better description of the change


MilhouseVH:
A clean build of RPi2 with connman 1.31 shows the following at boot:

screen

Network is working though.

chewitt:
@MilhouseVH - in this version connman attempts to symlink /etc/resolve.conf to /var/run/connmand and this fails due to the read-only filesystem. It then falls back to using traditional resolv.conf which we link to /storage/.cache so everything works. Can you look at journald logs to confirm that's the error? - I'll look at how to change that behaviour.

chewitt:
http://git.kernel.org/cgit/network/connman/connman.git/commit/?id=598248402eb82ad870ca8ac4b1a4e5f6bec948dc <= reverting that commit via a patch should probably be enough?

MilhouseVH:
Yep, looks like the same issue: http://sprunge.us/HBPj

```
Jan 10 03:29:34 OpenELEC systemd[1]: Started Journal Service.
Jan 10 03:29:34 OpenELEC systemd[1]: Starting Flush Journal to Persistent Storage...
Jan 10 03:29:34 OpenELEC systemd[1]: Started Flush Journal to Persistent Storage.
Jan 10 03:29:34 OpenELEC systemd[1]: Starting Create Volatile Files and Directories...
Jan 10 03:29:34 OpenELEC systemd-tmpfiles[220]: [/usr/lib/tmpfiles.d/z_03_connman.conf:19] Duplicate line for path "/var/cache", ignoring.
Jan 10 03:29:34 OpenELEC systemd-tmpfiles[220]: symlink(/var/run/connman/resolv.conf, /etc/resolv.conf) failed: Read-only file system
Jan 10 03:29:34 OpenELEC systemd[1]: Started udev Coldplug all Devices.
Jan 10 03:29:34 OpenELEC systemd[1]: systemd-tmpfiles-setup.service: Main process exited, code=exited, status=1/FAILURE
Jan 10 03:29:34 OpenELEC systemd[1]: Failed to start Create Volatile Files and Directories.
Jan 10 03:29:34 OpenELEC systemd[1]: systemd-tmpfiles-setup.service: Unit entered failed state.
Jan 10 03:29:34 OpenELEC systemd[1]: systemd-tmpfiles-setup.service: Failed with result 'exit-code'.
```

I'll try reverting those commits and update here a bit later.

lrusak:
How about setting the environement variable ```CONNMAN_RUNSTATEDIR_RESOLVCONF="no"``` ?

MilhouseVH:
Thanks. There are two possible places where `connman` creates the symlink - this patch will avoid both: https://github.com/MilhouseVH/OpenELEC.tv/commit/60e6e8683c0945bb630e9e0f1057bbaadc9bb723

packages/network/connman/package.mk
```diff
@@ -68,7 +68,7 @@ PKG_CONFIGURE_OPTS_TARGET="WPASUPPLICANT=/usr/bin/wpa_supplicant \
 
 
 PKG_MAKE_OPTS_TARGET="storagedir=/storage/.cache/connman \
-                      statedir=/run/connman"
+                      statedir=/var/cache"
```
MilhouseVH: @sraue / @seo - if you know why we were creating two `resolv.conf` files please advise.

With the current setup, the system creates `/var/cache/resolv.conf` (sym linked from `/etc/resolv.conf`) and populates it prior to starting `connman` with "fallback" nameservers `8.8.8.8` and `8.8.4.4`. `connman` itself is using and updating it's own `/run/connman/resolv.conf` file, and apparently ignoring `/var/cache/resolv.conf`.

With this change, there is now only one resolv.conf for the entire system: `/var/cache/resolv.conf`.

This file is initially populated with the same fallback nameservers as before, but once a valid network configuration is loaded (ie. manual or dhcp etc.) then these nameservers are overwritten by `connman`.

Sane?

sraue:
i would make it cleaner, we only have one one resolv.conf with connman-1.30, i look in this now

MilhouseVH:
In 1.30 it doesn't look like `resolv.conf` is being written to `/run/connman`, so this is new behaviour in 1.31 and fixed by this PR (`statedir=/var/cache`).

sraue:
my solution (incl moving /etc/hosts, to be sane) would look like this: http://paste.fedoraproject.org/309137/37513145

sraue:
+1 with the updated pr

packages/network/connman/package.mk
```diff
@@ -18,7 +18,7 @@
 
 PKG_NAME="connman"
 # DO NOT UPGRADE!!
```
lrusak: Remove this ;)

lrusak:
@MilhouseVH can you test this new change?

MilhouseVH:
Managed to find time for a quite test on RPi2, all good. Haven't been able to test on Generic - unrelated build issues.

packages/network/connman/package.mk
sraue: i would keep it, we can update for prereleases, but we should keep the comment to see we must be carefully for releases

stefansaraev:
did you not hit this bugs when testing the bump ?

http://git.kernel.org/cgit/network/connman/connman.git/commit/?id=f61f547e64aaf466522cf2a92af2ab450dc1134f
http://git.kernel.org/cgit/network/connman/connman.git/commit/?id=ccfae23093a170ffac92bc3c633bcec5a00c974c

MilhouseVH:
I didn't, no, but then my dhcp leases never expire. I'll PR those two commits later today.

stefansaraev:
I meant whoever bumps connman. this is probably an regression after 1.30 bump. 

here what's common sense to test.
- v4: isc-dhcpd (with and without broadcast flag turned on) and dnsmasq - with short lease time (5 minutes for easy testing)
- v4: boot without dhcp server enabled in network, try switch to static.
- v6: slaac (radvd)
- v6: prefix delegation (radvd + wide-dhcpv6 or isd-dhcpv6)
- v6/v6: switch from dhcp to auto and back. 10 times. tripplecheck if namesevers/timeservers are stored correctly
- netboot. network mounts. network wait. 

if any of the above fails with either wifi or wired (but worked before) - dont bump it, if you cant test - dont bump it :)

PR #4597, opened 2016-01-09
vdr-addon bump 6.0.2

CvH:
update to latest rls of sat>ip and vnsi-server

not yet compile/runtime tested


lrusak:
Let me know if you have compiled/runtime tested

lrusak:
Please test now that #4602 has been merged

sraue:
we also need the same on master, master should be never behind 6.0

CvH:
@sraue 
I pr a bump for master - this is only a "safe bump" and too ancient for master

PR #4600, opened 2016-01-10
xorg: enable tcp-transport so '-tcp nolisten' works

chewitt:
https://github.com/OpenELEC/OpenELEC.tv/commit/567719dea26c12c4b4bafd9a3e5c4c604f9273f5 is incomplete without this additional change. You need to enable tcp transport before -tcp nolisten can silence the error.

### DIFF
--- a/packages/x11/xserver/xorg-server/package.mk
+++ b/packages/x11/xserver/xorg-server/package.mk
@@ -115,7 +115,7 @@ PKG_CONFIGURE_OPTS_TARGET="--disable-debug \
                            --enable-xshmfence \
                            --disable-install-setuid \
                            --enable-unix-transport \
-                           --disable-tcp-transport \
+                           --enable-tcp-transport \
                            --disable-ipv6 \
                            --disable-local-transport \
                            --disable-secure-rpc \


sraue:
hmmm do we really need to enable tcp transport to later disable this? dont like this and we had always disabled tcp transport in xorg....

chewitt:
It makes no change to the running system - but you have to enable it at build-time to disable it again with config (and thus silence the error message) at run-time. I've used this in ATV builds for a while now with no visible impact to anything.

sraue:
i know... but maybe we can avoid this warning in the logs in a different way? i really dont want tcp support at buildtime if we dont need/use this... then i like it more to live with the warning in the logs

stefansaraev:
what about removing -tcp nolisten from systemd unit ?

chewitt:
Then you get the error in the logs. The error is harmless but this is the only way to remove it - the original idea was from @stefansaraev as I spotted the same in ATV builds a while back. I'm not sure why there's an objection since the tcp listen function is not active (same with with none of this and the error)..

chewitt:
Can someone explain why building with tcp (and then disabling in config) is a bad idea. I've used this in ATV builds for months with no issues and clean(er) logs are a good thing. Apologies if this is some dumb question (we all know my level) but all i'm hearing is "just because" which isn't a good reason to exclude this. Thanks.

sraue:
we never enabled tcp in xorg, for more then 6 years... also we dont should enable a feature at buildtime to disable the same at runtime... thats not the right fix for a problem.

sraue:
better revert https://github.com/OpenELEC/OpenELEC.tv/commit/567719dea26c12c4b4bafd9a3e5c4c604f9273f5

PR #4605, opened 2016-01-11
vdr-addon bump 7.0.1

CvH:
- updated all plugins to latest version
- compile tested but NOT runtime yet


stefansaraev:
please use mkpkg for dvbapi. whats the point of adding this -upstream patch.

also. before bumping dvbapi here, make sure oscam is updated in unofficial. or you are in trouble.

CvH:
The upstream.patch includes the latest patch to master (btw overlooked that mkpkg for dvbapi is there, compare was quicker then creating a mkpkg - I update pr)

Both workarounds are invalid if hopefully https://github.com/OpenELEC/OpenELEC.tv/pull/4537 get merged.

```This probing was added only for transitional period to not break
connections with OSCams < svn rev 9869.```
OE repo has 10607 "should" be okay but update oscam won't hurt either.

stefansaraev:
ok. 10607 is fine. thanks

PR #4608, opened 2016-01-11
Quoting to preserve whitespaces in banner

dolorosus:


dolorosus:
After applying this, change #4607 can be reverted to prevent us from ugly banners.

MilhouseVH:
Could you also do a backport for openelec-6.0 branch (this would allow #4608 to be reverted).

lrusak:
Can you include a revert of @dd2e7472a70a35fa3402a0d787189aea3f362bb1 before your commit?

Then rebase your commits to remove dolorosus/OpenELEC.tv@7ddd71a307871acd17a050106a48ca956365e4ee

PR #4609, opened 2016-01-11
Quoting variables to preserve whitespaces in banner

dolorosus:


lrusak:
Can you include a revert of @7e5f2557b3a00e8c2b6715f03bd5449526ecca61 before your commit?

PR #4617, opened 2016-01-13
linux: add support for TechnoTrend CT2-4650 version 2

trsqr:
This patch is probably going into kernel 4.6, so this OE patch
is obsolete after that.


sraue:
will it work on kernel 4.1 too? (for OE-6?)

trsqr:
The patch itself won't work with OE6, but I've got an OE6 version too that I can PR against openelec-6.0 branch if that's ok?

chewitt:
@trsqr submit the 6.0 patch too .. there's a chance it will still make 6.0.1

Issue #3554, opened 2014-10-18
[Request] mt7610u wireless driver


weha:
Would it be possible to add the driver for this usb stick?
It's the usb2.0 version of the T4U stick, half the speed, half the price.

According to this https://wikidevi.com/wiki/TP-LINK_Archer_T2U it should use the mt7650u_sta driver but it doesn't seem to be in openelec at this time (4.2.1).

Thank you!

cvlli:
hello @weha 
I found this solution for your problem. Keep in mind you need to do all the steps. 

First of all, you need to update/upgrade your linux version. 
apt-get update
apt-get dist-upgrade

A reboot may be indicated before you continue.So:
reboot

Then you need to download the driver from MediaTek, so you may do something like :
cd /usr/src
wget http://s3.amazonaws.com/mtk.cfs/Downloads/linux/mt7610u_wifi_sta_v3002_dpo_20130916.tar.bz2
tar xjf mt7610u_wifi_sta_v3002_dpo_20130916.tar.bz2
cd mt7610u_wifi_sta_v3002_dpo_20130916

Before you can compile the driver you need to edit "/usr/src/mt7610u_wifi_sta_v3002_dpo_20130916/include/os/rt_linux.h" at line ~279. There, replace the word int with kuid_t and kgid_t, or else, compilation will abort with an error. You should have something like this:

    typedef struct _OS_FS_INFO_
 {
    kuid_t              fsuid;
    kgid_t              fsgid;
    mm_segment_t    fs;
 } OS_FS_INFO;

Now you need to make the USB adaptor identifier known to the driver, so use lsusb.

lsusb 

The output will look like this:

Bus 001 Device 004: ID 0df6:0075 <<Reltek...>>

Notice the ID !

In the source directory of the module, you need to edit "common/rtusb_dev_id.c" by adding :

{USB_DEVICE(0x0DF6,0x0075)}, /* MT7610U */

The next thing we need to change is "os/linux/config.mk" :

HAS_WPA_SUPPLICANT=y
HAS_NATIVE_WPA_SUPPLICANT_SUPPORT=y

And lastly, edit the file "os/linux/config.mk" (line ~663) by changing 

CHIPSET_DAT = 2860 

to :

CHIPSET_DAT = 2870

Now you can do:

make
make install

modprobe mt7650u_sta

sudo iwlist ra0 scan

You are done ! I hope this helped you.

mrdominuzq:
[...]

PR #4625, opened 2016-01-15
mt7610u: add driver package

chewitt:
Initial support for the mt7610u chipset to solve #3554 - I don't have one of these devices so please do not merge until we see some users on the issue report confirming it works.

### DIFF
--- a/distributions/OpenELEC/options
+++ b/distributions/OpenELEC/options
@@ -60,7 +60,7 @@
 # for a list of additinoal drivers see packages/linux-drivers
 # Space separated list is supported,
 # e.g. ADDITIONAL_DRIVERS="DRIVER1 DRIVER2"
-  ADDITIONAL_DRIVERS="RTL8192CU RTL8192DU RTL8192EU RTL8188EU RTL8812AU dvbhdhomerun"
+  ADDITIONAL_DRIVERS="RTL8192CU RTL8192DU RTL8192EU RTL8188EU RTL8812AU mt7610u dvbhdhomerun"
 
 # build and install bluetooth support (yes / no)
   BLUETOOTH_SUPPORT="yes"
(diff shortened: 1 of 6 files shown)


sraue:
dont like its pulled from a github repo dedicated for OE only with no real commits... better is to repack the original sources or use a repo which really fixes stuff in this driver... other then that we should search for overlapping pciids from in-kernel mt7601 driver

sraue:
other then that it looks here like mediatek provides the same driver for MT7601 and MT7610... so if this driver comes from there we have to remove all MT7601 usb ids...

chewitt:
The mediatek site shows mt7601u release notes and download links but Google shows plenty of reports from mt7610u users building the driver and complaining it doesn't work, as it's the wrong driver. Both are forked from the same original source so they look similar and mediatek are lazy for not combining them into a single package, but they are different and AFAIK there are no USB ID overlaps. The only mt7610u sources I could ever find were from TP-LINK and wouldn't compile, but it looks like our user patched it up to solve that. The driver is not OE specific and IMHO is no better or worse than anything else you can find. https://github.com/kuba-moo/mt7601u/issues/6 <= this is the author of the in-kernel mt7601u driver stating they are different.

sraue:
is there another source then https://github.com/sohaib17/mt7610u which is a repo with 2 commits and done for openelec? whats the original source?

chewitt:
http://www.mediatek.com/AmazonS3/Downloads/linux/mt7610u_wifi_sta_v3002_dpo_20130916.tar.bz2

MT website now has links to the ^^ original driver (but still shows wrong release notes). It appears to be the same driver our user put in his git repo. This needs patches to build on a current kernel. I asked the user for links to the patches changes they applied. NB: his repo was created for him to use on OE but it is not no an OE specific driver, it builds fine on Ubuntu as well.

chewitt:
https://bitbucket.org/sanrath/mediatek_mt7610u_sta_driver_linux-64bit <= looks like a good source of patches for this driver. The only aspect of this driver that concerns me is the wireless config in the .dat file; commits show people making country-specific changes and OE should ideally have a 'one size fits all' configuration. I'll go back and have a look at what the mt7601u driver used before it was merged with the kernel.. as it's the same file used for the same purpose.

chewitt:
I'll find time for this between 6.95.1 and 7.0.0 release. If it's going to be used as an excuse to delay the 6.95.1 beta I will close the PR and submit it again later.

PR #4627, opened 2016-01-15
ffmpeg: Stop disabling ARMv5TE and ARMv6T2 - needed by ASM optimisations

MilhouseVH:
Disabling ARMv6T is preventing some ARM ASM from being used (ie. cabac optimisations in HEVC code) as `HAVE_ARMV6T2_INLINE` is not enabled, which is hurting performance.

Unless there's a specific reason to disable ARMv5TE then might as well leave it available as well.


popcornmix:
We are currently not making use of arm assembly code like:
https://github.com/FFmpeg/FFmpeg/blob/master/libavcodec/arm/cabac.h#L23

PR #4628, opened 2016-01-15
Update WeTek Play bootloader

### LINKED ISSUE #4550: Boot from mSD on WeTeK OpenELEC Edition not working
@codesnake knows this already, but I wanted to make this official. 

But well, some info about the issue: 
It's not possible to boot OpenELEC from mSD card when OpenELEC is installed to NAND
(e.g. it's not possible to use test builds of OE on mSD, when a (stable) build is installed on NAND).
It get's, somehow, wrong bootargs.
Enigma Images are booting fine but not OE.

At the moment I cannot provide anything more, 
what codesnake doesn't know already.

codesnake:
New bootloader makes possible to boot OE from SD-card when another OE is already installed in NAND memory (fixes #4550).


lrusak:
Do we want this in OE 6.0 too?

codesnake:
Yes, would be good.

PR #4630, opened 2016-01-17
Oe6 update

sraue:


codesnake:
Note that OpenELEC-settings in this PR still has a bug with incorrect path in do_restore function.

stefansaraev:
argh. did you also just broke it in 6.0 branch?

PR #4631, opened 2016-01-17
Package updates

sraue:


chewitt:
@sraue please add the USB ID's from http://openelec.tv/forum/88-wlan/79629-8812au-wifi-module#155780 to the 8812au update

sraue:
this device id is there... no logs before and after the module loading - no idea what todo

chewitt:
I've asked the forum user to check a milhouse build. If it doesn't show up there I'll investigate.

PR #4635, opened 2016-01-19
kodi: add patch so pulseaudio is at the end of the list

### LINKED ISSUE #4585: With recent inclusion of Pulse Audio in Master AAC2.0 isn't working properly
I noticed last night after a clean re-compile and upgrade to a master copy that files with AAC2.0 were silent and not playing audio. I tried a 5.1 file without any issues and was able to revert to HDMI0 and the same file played just fine. 

I have a 5.1 setup and a 2.0 setup and both exhibited the same problems. 

Debug log from the 5.1 setup below.

http://paste.ubuntu.com/14431337/

lrusak:
Thanks @fritsch for this.

This makes it so that pulseaudio output isn't selected automatically so we don't have people accidentally using it.

This fixes #4585 

keep open till @fritsch gives the go ahead


sraue:
lrusak please merge asap, if it works

lrusak:
is @fritsch happy with his patch?

fritsch:
I am not happy with the approach as you know a real PA Sink does much more, also in case of buffer handling / delay and so on. 

But that will at least care that users don't default to the "pulse" device in the future. Get it tested and see which issues come up.

lrusak:
well perhaps down the road I will switch it back to using a dedicated PA sink in kodi, for now this works.

PR #4637, opened 2016-01-20
xf86-video-ati: use DRI3

lrusak:
I'd like to test this out, not sure if there is any negative impact

see, http://www.phoronix.com/scan.php?page=article&item=radeon-dri3-perf&num=1

```
grep -i dri3 /var/log/Xorg.0.log
```
should show something similar to:
>[1550498.698] (**) RADEON(0): Option "DRI3" "on"
>[1550498.864] (**) RADEON(0): DRI3 enabled

### DIFF
--- a/packages/graphics/mesa/package.mk
+++ b/packages/graphics/mesa/package.mk
@@ -67,7 +67,7 @@ PKG_CONFIGURE_OPTS_TARGET="CC_FOR_BUILD=$HOST_CC \
                            --disable-gles1 \
                            --disable-gles2 \
                            --enable-dri \
-                           --disable-dri3 \
+                           --enable-dri3 \
                            --enable-glx \
                            --disable-osmesa \
                            --disable-gallium-osmesa \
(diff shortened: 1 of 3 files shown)


Klojum:
Result on AMD A4-5300:
[1567747.349] (**) RADEON(0): Option "DRI3" "on"
[1567747.627] (**) RADEON(0): DRI3 enabled
No visible negative impact when enabled.

fritsch:
Guys your mesa is built with:
```
--enable-dri \
 --disable-dri3 \
```

What impact did you think of?

lrusak:
hmmm not sure how I didn't see this. 

xorg-server is built with dri3 support already.

fritsch:
And also take care: intel ships with dri3 off by default as they have sever bugs in the dri3 chain, via: https://bugs.launchpad.net/ubuntu/+source/xserver-xorg-video-intel/+bug/1401784

lrusak:
Yea I'll leave it off for intel. I don't have intel hardware to test with so that's not my territory.

sraue:
please hold this back until OE 8.0, i had seen problems with dri3 in the past (intel gpu), thats why its still disabled (and yes i think it was a problem in mesa, so we cant enable this for other gpus)

PR #4640, opened 2016-01-21
openvpn: restore openvpn to core builds

chewitt:
Removing OpenVPN from the OE settings app was necessary (as the configuration GUI never made sense) but moving the binary to the unofficial repo only added confusion for users who want or need to use a VPN service in daily-use. The goals of this PR are:

a) Make the process of setting-up an OpenVPN connection a little easier
b) Make popular replacement GUI add-ons (e.g. Brian Hornsby's) work OOB
c) Make OE a static target for VPN service providers to write VPN GUI add-ons around
d) Ship an example systemd service for VPN autostart (based on @Frank-NL forum work)

Those who have no need of a VPN regard them as frippery that belongs in a repo. Those who need a VPN for daily-use regard it as core networking. I personally sit 50/50 between both sides, but I believe OE should evolve to reduce user-experience complexity and 380k is not a big penalty. OE minimalists will be against the idea - but I submit it for debate anyway.


wiwwo:
I really can't understand why (open)VPN functionality has been removed from openElec.
You are selling WeTek, that i was willing to buy, but since i live aboard that would be useless to me cause i won't be watching my homeland tv. And if i'd bought it i would have felt defrauded!
Apart from that, VPNs are *basic* requirements in the post-Snowden era!

sraue:
i would like to ask @lfiebach if we can readd (a better) config dialog to our settings addon. also was there not connman support for openvpn which we should enable too?

lrusak:
@sraue the problem with connman openvpn support is that is uses a non standard config file (and there really isn't any documentation for it).

It's better to be able to just run a .ovpn file with inline keys.

Unless we build in all the options into the config, the .ovpn method is the best.

chewitt:
@lrusak @sraue connman can use a config file as I persuaded connman devs to add that capability. The issue is that /storage/.cache/connman/ service files use vpn_servername_opvn naming so even if the user has a self-contained .conf/.ovpn file they need to open it up to discover the server name (a 'remote' in openvpn-speak) then enter this name via the GUI. It was a good thing to remove the connman GUI as it didn't follow OpenVPN conventions, but I'm not sure we need to provide a replacement. There are existing well-featured OpenVPN config add-ons and OE users have already started to write provider-specific ones (which would be easier to use if we remove 2 addon dependencies). I'd prefer to leave OE settings alone and encourage VPN providers and our user-ecosystem to fill that gap. We need to make the plumbing normal first though, which is what this PR is about.

sraue:
why we cant have both ways? if not setup via the gui (or connman method) simply support the ovn files?

chewitt:
The "connman method" is complicated to configure and makes no sense. This PR _does_ allow use of a simple ovpn (conf) file from the console or popular config add-ons. It also supports auto-starting the connection at boot via a systemd service (the example file has been updated for connman-wait-online changes). TL;DR - this PR allows use of OpenVPN with ONE (config) add-on vs. the current situation which requires THREE (unofficial repo, openvpn binary, config) add-ons.

chewitt:
How is this a blocker?

chewitt:
I am testing update to 2.3.10 which needs patches - do not merge

chewitt:
PR updated to 2.3.10 and tested working

PR #4643, opened 2016-01-21
fix PKG_DEPENDS_TARGET to include libjpeg-turbo

steffenhoenig:
won't compile otherwise


lrusak:
libjpeg-turbo is built before v4l-utils because it is a dependency of an earlier package. So this shouldn't be needed unless building packages out of the usual order.

steffenhoenig:
Well, If you modify the distribution/$distri/options file, like I did to "tinker" with the building environment, this is certainly the case. I didn't add or removed a package though

MilhouseVH:
If there is a known package dependency we should explicitly include it in package.mk, otherwise any change to the build order will likely bite us in the arse.

lrusak:
Unless you set MEDIACENTER="no", but why would you do this and keep REMOTE_SUPPORT="yes"

steffenhoenig:
Well, because one can?

steffenhoenig:
like @MilhouseVH said, if the dependency is obvious I'd go for the explicit way, but that's personal believe I guess

MilhouseVH:
If the dependency is conditional then the dependency should still be in the `package.mk`, but conditional, ie.

```
[ "$MEDIACENTER"="yes" ] && PKG_DEPENDS_TARGET="$PKG_DEPENDS_TARGET libjpeg-turbo"
```
(or whatever)

The point is, we should still be including the known dependency when required and not leaving it to chance.

stefansaraev:
what @MilhouseVH said. known (hard) dependencies should be listed. 

optional dependencies should be either disabled by configure switch (if one available and dependency not needed) or listed in PKG_DEPENDS* (if configure switch not available). predictable builds are a must.

sraue:
wait... why we need jpeg support for v4l utils where we only need ir-keytable? makes no sense for me... if there is a buildproblem it should be fixed in another way

sraue:
please try with --without-jpeg for configure and revert this commit again if this works

vpeter4:
Seems --without-jpeg is doing it's job.

PR #4649, opened 2016-01-24
repository.unofficial.addon.pro: change kodi minversion to allow beta…

lrusak:
…/rc releases

Should be bumped to 16.0 after kodi final is out.


sraue:
dont really think we shouldand must bump this to 16.0, except there are some changes which makes the betas incompatible with 16.0...

PR #4653, opened 2016-01-25
Pulseaudio update to 8.0 and switch back to AEPulseSink

lrusak:
as per @fritsch recommendations

This removes the alsa-->pulse bridge and creates a dedicated pulse sink in kodi.

Apparently this is a better solution. Maybe @fritsch can say why this is the best way to do it.

I will probably need to create another patch so that the Pulseaudio: dummy output device isn't shown.

With this PR the audio output looks like this
**RPi2**:
RPi audio settings

and **Generic**:
Generic audio settings


fritsch:
Looks good. Someone other distros can copy :-) Please add the "usability patch" we discussed and I'd say - nicely done. 

Btw. just for interest, can you show me where you tell PA to ignore all other devices besides Bluetooth?

lrusak:
I’m not loading the pa module ```module-udev-detect``` see, @5782ed64053805271fef2da1ffc86777e4f53519

sraue:
done?

MilhouseVH:
No reported issues so far...

lrusak:
:+1: should be good to go. Thanks for asking.

I hope this is the last time I have to muck with pulseaudio

fritsch:
Hope dies last - but you are doing a good job here :-)

2016-01-29 11:05 GMT+01:00 Lukas Rusak <notifications@github.com>:

> [image: :+1:] should be good to go. Thanks for asking.
>
> I hope this is the last time I have to muck with pulseaudio
>
> —
> <https://github.com/OpenELEC/OpenELEC.tv/pull/4653#issuecomment-176677234>
> .
>



-- 
                   Key-ID:     0x1A995A9B
                   keyserver: pgp.mit.edu

PR #4656, opened 2016-01-27
various build fixes

lrusak:
I think we want all of these.

@stefansaraev do we want the stripping changes too? 
ie:
stefansaraev/TB@8677d037bb9c1af54f6ff6e861c4b669c32cd26a
stefansaraev/TB@1b40e92ba23866985588d50f8c4e3cb162611d9b
stefansaraev/TB@cc2906b7472c4bd7e69796b046821c0b56c24956
etc.

I'm asking because I'm not sure why you chose to add a new function `strip_debug` to some packages. :)

### DIFF
--- a/packages/debug/gdb/package.mk
+++ b/packages/debug/gdb/package.mk
@@ -18,7 +18,7 @@
 
 PKG_NAME="gdb"
 PKG_VERSION="7.10.1"
-PKG_REV="1"
+PKG_REV="2"
 PKG_ARCH="any"
 PKG_LICENSE="GPL"
 PKG_SITE="http://www.gnu.org/software/gdb/"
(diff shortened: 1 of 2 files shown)


stefansaraev:
https://github.com/lrusak/OpenELEC.tv/commit/a47eba03d13c2fa6b906a4846264a8b0392a4c2a does nothing without https://github.com/stefansaraev/TB/commit/8677d037bb9c1af54f6ff6e861c4b669c32cd26a - but if you pick it, you will need more than current stripping changes. but yeah, (parts of) glibc should not be stripped

lrusak:
I removed the stripping stuff from this PR and will include it in another as it doesn't really belong here.

The rest of this should be sane :+1:

sraue:
i would add script/addons as a seperate PR, this i want to check soon, the other both are ok to merge asap

lrusak:
done

packages/debug/gdb/package.mk
```diff
@@ -18,7 +18,7 @@
 
 PKG_NAME="gdb"
 PKG_VERSION="7.10.1"
-PKG_REV="1"
+PKG_REV="2"
```
stefansaraev: uh. not needed but doesnt hurt.

Issue #4483, opened 2015-12-08
Extra systemd mounts prevent systems rebooting after #4455


MilhouseVH:
After #4455, if you add a systemd mount and try to reboot, the system will hang while rebooting.

Add an nfs mount (following is an example, change paths etc. to suit circumstances):

```
[Unit]
Description=test nfs mount script
Requires=network-online.service
After=network-online.service
Before=kodi.service

[Mount]
What=192.168.1.66:/mnt/sdb2/PVR
Where=/storage/pvr
Options=
Type=nfs

[Install]
WantedBy=multi-user.target
```

Once the mount service has started and mounted the NFS share, try to reboot the system - It will stop all the services, then hang before rebooting.

If the connman-1.30 bump is reverted, the system does reboot normally. If there is no systemd mount, the system will also reboot normally with connman-1.30.

However, with connman-1.30 AND systemd mount, the system will hang at reboot.

mlade77:
Btw. in my case that happens only when I am using WiFi dongle. 
If I use cable that doesn't happen.
https://github.com/OpenELEC/OpenELEC.tv/issues/4504

the-dreamer:
in my case i have this problems only when i have more than one mount too use. my current workaround seems to be adding some mounts directly in kodi. my pvr / recordings folder is mounted on startup. the other mounts (music & pictures) directly in kodi without mount script. 

reboot/shutdown hangs seems to be gone. tried with 01.01.2016 milhouse build and OE7 fritsch egl build. 
btw. i saw in my scripts leftovers from xbmc which i have renamed now.

MainUser1:
In my case this also happens when using a wired connection.

I think it should be noted, that network booting OpenELEC from NFS also breaks reboot / shutdown.
There doesn't seem to be a usable workaround in this situation.

I'm more than willing to help if there is something I can do to track this issue down.

MilhouseVH:
If you have this problem can you test build #0205 as this build contains a potential fix, and report your results:

RPi1/RPi2: http://forum.kodi.tv/showthread.php?tid=250817
x86: http://forum.kodi.tv/showthread.php?tid=238393

chewitt:
closing as resolved in #4659

PR #4659, opened 2016-01-30
connman-wait-online and connman.service changes

### LINKED ISSUE #4475: the storage partition doesn't get unmounted cleanly on shutdown
Installed a build of the master branch from yesterday to a sd card for the wetek play. After a shutdown it will boot into Android. If I put the sd card to a linux system the journal of the storage partition gets repaired and after this the wetek will boot into OpenELEC again.

 Gerald

lrusak:
This allows us to use the connmand-wait-online tool that connman provides.

I have moved `mediacenter/kodi/system.d/kodi-waitonline.service` to `network/connman/system.d/connman-wait-online.service` as this is a better place for it.

Kodi now wants `network-online.target`, this cleans things up a bit.

I have also adjusted the connman.service to conform more closely to upstream. This seems to fix the systemd failed to unmount /flash and /storage errors. It may also fix #4475 and #4483 but will need more testing. It worked for me in a quick NFS booting test.

@MilhouseVH are you able to include this in your builds?


MilhouseVH:
Confirming that this does seem to fix #4483 (the problem is easy to reproduce on a build from December with a single systemd NFS mount, but more difficult to reproduce in a more recent test build - most times it would reboot but occasionally it would not). 

I'll include this PR in future test builds and see if we can get additional testing results.

MilhouseVH:
OK I'm going to have to give this a :-1:

I had in fact been testing the following, which works with the NFS mount service from #4483 and fixes the reboot issue:

```
[Unit]
Description=Connection service
DefaultDependencies=false
Conflicts=shutdown.target
Before=network.target multi-user.target shutdown.target
Wants=network.target

[Service]
Type=dbus
BusName=net.connman
Restart=on-failure
EnvironmentFile=-/run/openelec/debug/connman.conf
ExecStartPre=-/bin/sh -c "echo -e 'nameserver 8.8.8.8\nnameserver 8.8.4.4' > /etc/resolv.conf"
ExecStartPre=-/bin/sh -c "cat /storage/.cache/hostname > /proc/sys/kernel/hostname"
ExecStartPre=-/bin/sh -c "cat /storage/.config/hosts.conf > /etc/hosts"
ExecStartPre=/sbin/ifconfig lo 127.0.0.1 netmask 255.0.0.0 up
ExecStart=/bin/sh -c ". /usr/lib/connman/connman-setup; exec /usr/sbin/connmand -nr $CONNMAN_MAIN $CONNMAN_DEBUG"
StandardOutput=null
RestartSec=2
StartLimitInterval=0

[Install]
WantedBy=multi-user.target
```
(create the above file as `/storage/.config/system.d/connman.service`)

However when using this PR (and without `connman.service`) the mount service from #4483 no longer works - there is no longer a `network-online.service` so users will need to modify their existing mount services to use a different `Requires`/`After` (any suggestions?), which isn't ideal. But if that's the price to pay for reliable rebooting, then maybe it's worth it.

lrusak:
@MilhouseVH I have updated this PR, please test again.

I changed `connman-wait-online.service` back to `network-online.service` as there is no reason for this to change and it would cause breakage (without using an alias).

I also kept `kodi-waitonnetwork.service` as discussed.

So:
- The "wait on network" option in the OpenELEC add-on settings effects `kodi-waitonnetwork.service`
- Any `*.mount` files should still use `network-online.service`
- The service files and targets have been adjusted to use network-online.target

MilhouseVH:
I've been testing the latest updates - looking better, thanks.

Existing mount services will now mount reliably whether OE Settings "Wait for network" is enabled or not. :+1:

However I'm still getting occasional hangs at reboot - I think `Conflicts=shutdown.target` is required in the mount service to avoid the hangs - more testing required to confirm this however.

chewitt:
It would be good to test Kodi with a MySQL database as that's the main reason the "wait for network" hack exists. If connman is now waiting properly for drivers to load and the network to come up it would be nice to remove this (at some future point, and after much testing).

MilhouseVH:
Unfortunately I'm still experiencing hangs at reboot when using a mount service and this PR:
```
[Unit]
Description=test nfs mount script
Requires=network-online.service
After=network-online.service
Before=kodi.service
Conflicts=shutdown.target

[Mount]
What=192.168.0.3:/mnt/share/media
Where=/storage/mymedia
Options=
Type=nfs

[Install]
WantedBy=multi-user.target
```
The above includes `Conflicts=shutdown.target`, but it doesn't seem to have any effect - hangs occur at reboot with or without this line.

packages/mediacenter/kodi/system.d/kodi-waitonnetwork.service
```diff
@@ -1,16 +1,19 @@
 [Unit]
 Description=Wait on network
+Requisite=connman.service
 After=connman.service
-Before=kodi-autostart.service kodi.service
+Before=network-online.target
```
stefansaraev: why?

packages/mediacenter/kodi/system.d/kodi-waitonnetwork.service
```diff
@@ -1,16 +1,19 @@
...
+DefaultDependencies=no
```
stefansaraev: hum.

lrusak:
@MilhouseVH I've had time to test the nfs .mount issue and it seems to be pretty random. I've only run into the failed unmounting a couple times but am unable to figure out what is causing it.

you shouldn't need the `Conflicts=shutdown.target` line because mounts automatically conflict on umount.target which should be more than enough.

I'm not really sure how to proceed.

lrusak:
So after a bunch of reading and testing it seems that adding `After=dbus.service` fixes the NFS mount issue.

https://bugs.launchpad.net/ubuntu/+source/dbus/+bug/1438612

chewitt:
@MilhouseVH - holding on merge until you give us positive field reports, thanks :)

MilhouseVH:
Good idea. Will get this in RPi and Generic builds tonight. Maybe merge Friday or Saturday if reports are good?

chewitt:
that's the general idea.. keep us posted

MilhouseVH:
I've had a single positive report that this fixes the hang on reboot when using systemd mounts - would have preferred more but... meh.. Personally speaking it also fixes it for me, as I can reliably reproduce the hang while rebooting with a systemd mounted filesystem, as can @lrusak. Not seen any negative reports in relation to this PR, so I'd say merge it.

PR #4660, opened 2016-01-30
xf86-video-nvidia: update to xf86-video-nvidia-352.79

MilhouseVH:
http://www.geforce.com/drivers/results/97645


chewitt:
The release notes state this adds support for new GPU devices, so are there any udev rule changes needed? ..e.g. the list of device ID's that are used to select the non-legacy driver.

MilhouseVH:
I don't know, is that something which runs as part of the build or is it a static list that needs updating with each new driver? I included this driver in my last Generic test build #0129:

http://forum.kodi.tv/showthread.php?tid=238393&pid=2229796#pid2229796

chewitt:
packages/x11/driver/xf86-video-nvidia/scripts/make_nvidia_udev.py needs to be run to generate a new 96-nvidia.rules then diff against the existing file to see if there's anything new added.

lrusak:
Yes, as @chewitt said. the [script](https://github.com/OpenELEC/OpenELEC.tv/blob/master/packages/x11/driver/xf86-video-nvidia/scripts/make_nvidia_udev.py) needs to be run.

From the root OpenELEC directory you simply need to run it
```
python packages/x11/driver/xf86-video-nvidia/scripts/make_nvidia_udev.py
```
You made need some python packages in order to run it (python-lxml python-requests). It should be compatible with python 2 or 3.

MilhouseVH:
Would it be possible for this python process to become part of the build, rather than for it to be run manually whenever something changes?

I've added a second commit to include the updated rules.

lrusak:
Nope, as it pulls info from a website I'd rather not. It just needs to be run whenever the package version changes.

MilhouseVH:
OK will try and remember next time. :)

lrusak:
If you want to include a comment next to or above the PKG_VERSION to remember to run the script that would be great :)

MilhouseVH:
How's that?

PR #4669, opened 2016-02-02
Various package updates

lrusak:
~~libcec should be safe.~~ nevermind, needs p8-platform

Curl update might need some more testing then what I was able to provide.

pixman should also be safe.

Will look for other updates to include as well.


sraue:
please merge, check also libressl, kernel, cmake, ffmpeg, libnfs

PR #4673, opened 2016-02-03
More package updates

lrusak:
The only one I'm concerned about here is the libnfs bump. It is currently being worked in the kodi depends https://github.com/xbmc/xbmc/pull/9016

All others are just maintenance updates.

@sraue do you have any more to add?


MilhouseVH:
I've had libnfs master in my builds for quite a while - there's not much new in libnfs 1.10.0, apart from the version change, that hasn't been in my builds for a few weeks/months.

lrusak:
Should be safe to merge then :+1:

Issue #4674, opened 2016-02-03
Build of master fails


grin1:
>BUILD    debug (target)
  INSTALL    debug (target)
      INSTALL    gdb (target)
basename: extra operand build.OpenELEC-WeTek_Play.arm-7.0-devel/image/system/lib/modules/RTL8192CU'
Try basename --help' for more information.
make: *** [release] Fehler 1

What is the problem here?

stefansaraev:
remove it from ADDITIONAL_DRIVERS

@sraue @lrusak you better remove it too. wetek doesn't really need this 3rdparty rtl piece of nonsense.

sraue:
@stefansaraev where it does help to remove this from ADDITIONAL_DRIVERS?

@grin1 please paste the output from ``` ls -lah build.OpenELEC-WeTek_Play.arm-7.0-devel/image/system/lib/modules```

grin1:
> drwxr-xr-x 3 grin1 users 4,0K Feb  3 14:41 3.10.94
drwxr-xr-x 2 grin1 users 4,0K Feb  3 14:41 RTL8188EU
drwxr-xr-x 2 grin1 users 4,0K Feb  3 14:41 RTL8192CU
drwxr-xr-x 2 grin1 users 4,0K Feb  3 14:41 RTL8192DU
drwxr-xr-x 2 grin1 users 4,0K Feb  3 14:41 wetekdvb

kdietrich:
Can confirm. Submitted a PR.

codesnake:
@stefansaraev Someone still use USB Wi-Fi sticks instead of built-in Wi-Fi on WeTek Play.

stefansaraev:
yea sure.

codesnake:
Strange, just rebuilt the WeTek Play project from master without any problems.

chewitt:
I'm closing this ticket as I built both WP/WC from current master yesterday so this is a local build issue on your environment. Make sure you are doing clean (not incremental) builds and remove ~/.ccache* - if you have further issues open a thread in the forums. Thanks.

PR #4675, opened 2016-02-04
projects/WeTek_Play: Remove RTL drivers

kdietrich:
WeTeK build fails with them and @stefansaraev confirmed that they are not needed. See https://github.com/OpenELEC/OpenELEC.tv/issues/4674

### DIFF
--- a/projects/WeTek_Core/options
+++ b/projects/WeTek_Core/options
@@ -123,7 +123,7 @@
   # for a list of additinoal drivers see packages/linux-drivers
   # Space separated list is supported,
   # e.g. ADDITIONAL_DRIVERS="DRIVER1 DRIVER2"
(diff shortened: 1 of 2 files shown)


chewitt:
please also remove them from the WeTek_Core project and squash the commits

kdietrich:
Okay now?

chewitt:
Can you set the commit message to "wetek: remove RTL drivers from core/play projects" .. it's nitpicking but we try to keep file/paths out of the message.

kdietrich:
Makes sense, done.

sraue:
No, there is no reason to remove this drivers. and they are needed by this users who uses this hardware

sraue:
closing this, this is not the solution to fix #4674

PR #4676, opened 2016-02-04
build: Take a lock on the package download directory

MilhouseVH:
This avoids dropping `.lock` files all over the `sources` directory which is a bit messy - now uses the directory itself as the lock resource (read access).


lrusak:
:+1:

PR #4680, opened 2016-02-06
Package updates

sraue:
dbus-glib and 	timezone-data should be backported to 6.0 branch

### DIFF
--- a/packages/devel/dbus-glib/package.mk
+++ b/packages/devel/dbus-glib/package.mk
@@ -17,7 +17,7 @@
 ################################################################################
 
 PKG_NAME="dbus-glib"
-PKG_VERSION="0.104"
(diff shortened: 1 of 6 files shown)


packages/devel/dbus-glib/package.mk
```diff
@@ -17,7 +17,7 @@
 ################################################################################
 
 PKG_NAME="dbus-glib"
-PKG_VERSION="0.104"
+PKG_VERSION="0.106"
```
chewitt: this also requires a URL update, see https://github.com/OpenELEC/OpenELEC.tv/pull/4687

MAFLO321: now URL update is in #4691

PR #4682, opened 2016-02-06
drop package: media_build

CvH:
useless if not updated very frequently + recreate the OE images


codesnake:
I was going to make use of this for WeTek Play and Core, because some users request to support some tuners not supported by kernel 3.10, but you was faster :smile:

CvH:
@codesnake this wouldn't help either, without specific patches it wouldn't compile for WeTek (already tested a big number of versions)

sraue:
@codesnake if you need this you can revert/readd and update .... also if its usefull for wetek we should include per default in this builds?

codesnake:
@sraue I will try to build it and in case of success will re-add the package and include it to  ADDITIONAL_DRIVERS for Play and Core.

PR #4684, opened 2016-02-07
Increase pre_allocated_buffer to fix buffering issues.

kevmegforest:
Also activate verbose procfs for better debugging and setting dynamic settings.

This is useful for people with HDA audio with multichannel analog output that needs special asound.conf.


chewitt:
Can you be more explicit on what buffering issues this resolves please. Provide links to forum posts etc. with logs and diagnostics.

projects/Generic/linux/linux.x86_64.conf
```diff
@@ -3147,7 +3147,7 @@ CONFIG_SND_DYNAMIC_MINORS=y
 CONFIG_SND_MAX_CARDS=32
 # CONFIG_SND_SUPPORT_OLD_API is not set
 CONFIG_SND_PROC_FS=y
-# CONFIG_SND_VERBOSE_PROCFS is not set
+CONFIG_SND_VERBOSE_PROCFS
```
fritsch: Are you sure you need that one?

lrusak:
@chewitt http://openelec.tv/forum/68-audio/79877-dshare-alsa-plugin-audio-dropout#157199

it doesn't hurt to have it

stefansaraev:
this is ok if you remove CONFIG_SND_VERBOSE_PROCFS
```
echo 2048 > /proc/asound/card0/pcm0p/sub0/prealloc
```
maybe immediate fix for your issue. fyi debian and ubuntu kernels set it to 64
1024 should be enough.

kevmegforest:
I don't know what's the best way to go.
Removing CONFIG_SND_VERBOSE_PROCFS will not permit to do 
`echo 2048 > /proc/asound/card0/pcm0p/sub0/prealloc`
It will permit to keep the 2048 by default for everyone else if you keep the other change. 

The other way around is keeping CONFIG_SND_VERBOSE_PROCFS while keeping the default at 64.

I think it's better to get those 2. More control and info of the sound hardware can't be bad to have around in the kernel. And the 2048 settings doesn't really affect a lot of memory too, while reducing the chance of buffer issues in some setups.

stefansaraev:
then it's ok to have both.

projects/Generic/linux/linux.x86_64.conf
MilhouseVH:
```
+CONFIG_SND_VERBOSE_PROCFS
```
Shouldn't this be `=y` or something?

And what about other platforms (eg. RPi etc.) - is this option relevant there too? Aren't we still maintaining Nvidia_Legacy for the time being?

kevmegforest: I can add it tomorrow also.
For the other platforms, I have no idea if it's important for the
snd-hda-intel module since I don't think it's included in non x86 platform
. But the verbose sound fs can be useful in those too.

kevmegforest:
I've added all the changes in all projects just as asked by @MilhouseVH 
I've also added the =y which was missing in the original PR.

I think it's done.

PR #4692, opened 2016-02-07
enable sata

galeots:
compile orange pi plus the sata and the power went out!


sraue:
if you want orangePi support in OpenELEC send a proper PR and maintain it (and it has to work with kodi without (much) patches):
closing this for now

PR #4697, opened 2016-02-10
config/noobs: update slides

lrusak:
This updates the noobs slides.

I also added *xcf files so the slides can easily be edited by anyone using gimp (or photoshop)

I made the system partition size 256MB


MilhouseVH:
Would a 512MB FAT partition be a problem? I can understand there may be potential reasons why it's not required (NOOBS users being least likely to get involved with debug testing etc. etc.) but a consistent FAT size for all platforms would be one less headache.

lrusak:
I'll create a new PR

sraue:
partition_size_nominal should be the size which is needed as minimum... so noobs can calc all the partition sizes for all distros which will be installed. we dont should make this to large...

sraue:
also "uncompressed_tarball_size": 120, should almost fit with the uncompressed result

PR #4699, opened 2016-02-10
config/noobs: use 512MB parition size for consistency

lrusak:
Thanks @MilhouseVH ;)


sraue:
i am against this.... why this is merged so fast (and other PRs must sit some days)?

PR #4701, opened 2016-02-10
scripts/mkimage: create disk image without sudo

vpeter4:
Mtools is used to format and copy files to vfat partition.
Populatefs is used to copy files to ext4 partition.

Tested on imx6 with u-boot, Generic with BIOS and EFI boot.

To avoid messing around with ext4 partition we could also provide "empty" disk image file. Both partitions would be already formatted and storage partition would already contain resize file. Then mkimage script would only copy kernel, system, bootloader files and install bootloader.
But current approach is more flexible for future use.


lrusak:
Awesome!

lrusak:
@vpeter4 are you able to provide some sample disk images for testing?

vpeter4:
Don't have anything at hand. But will build some for a test run over night.

PR #4705, opened 2016-02-11
systemd: update to 229

lrusak:
Builds ok, nothing really in the changelog is applicable to use.
https://github.com/systemd/systemd/blob/master/NEWS#L3...#L220

Please keep open till some runtime testing has been done.


lrusak:
Runtime tested and it is working

stefansaraev:
http://sprunge.us/AHCI it seems systemctl disable does not clean up *.wants/

stefansaraev:
well, now I see why it fails. you can't do systemctl disable /path/to/unit now

```
systemctl disable "$1.service"
```

and removing the "rm -f" workarounds should be enough.

PR #4706, opened 2016-02-11
update packages

CvH:
referring to OpenELEC/OpenELEC.tv/issues/4526
- libopenmpt compiles but portaudio is now enforced (have to disable it through conf) - I have no idea how to check if this works runtime
- rtmpdump compiles (sources are uploaded)

### DIFF
--- a/packages/audio/libopenmpt/package.mk
+++ b/packages/audio/libopenmpt/package.mk
@@ -17,12 +17,13 @@
 ################################################################################
 
 PKG_NAME="libopenmpt"
-PKG_VERSION="0.2.4764"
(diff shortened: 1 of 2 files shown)


sraue:
@MilhouseVH can you test (rtmpdump) please?

MilhouseVH:
@popcornmix has been updating rtmpdump to fa8646d by way of UpdateToLatest.diff in newclock5 - this has been in use since December so should be fine:

https://github.com/popcornmix/xbmc/commit/5b7c6a060e99eea8ee8363d13fdd839fa0a0593c

PR #4728, opened 2016-02-19
Add entropy generation

### LINKED ISSUE #4333: 5.95.5: WiFi Set to "Inactive" on Reboot

lrusak:
This hopefully fixes #4333 

Waiting to hear back from @MattDevo 

If this works it should go in for openelec-6.0 aswell


MattDevo:
checking now against OE master, had to make a slight modification to compile with OE-6.0 (after which it worked) so I'll push and PR those shortly

MattDevo:
@lrusak I needed to remove the [install] section entirely from add-random-at-shutdown.service in order for it to compile for both OE-6.0 and master.  Without that section removed, the build process kept creating /build.xyz/image/usr/lib/systemd/system/poweroff.target as a directory, and then trying to overwrite it with a file of the same name which caused the build process to fail

packages/sysutils/entropy/system.d/add-random-at-shutdown.service
```diff
@@ -0,0 +1,12 @@
...
+DefaultDependencies=no
+Before=systemd-poweroff.service systemd-reboot.service systemd-halt.service
+
+[Service]
+Type=oneshot
+ExecStart=/usr/lib/entropy/add-random-at-shutdown
+RemainAfterExit=yes
+
+[Install]
+WantedBy=poweroff.target reboot.target halt.target
```
stefansaraev: our enable_service supports only one WantedBy.

MattDevo:
@stefansaraev all good when limiting to one WantedBy - question is which is best ?

stefansaraev:
none. https://github.com/OpenELEC/OpenELEC.tv/blob/master/config/functions#L227 must be fixed.

```
enable_service () {
  local unit="$1"
  local unit_dir="/usr/lib/systemd/system"
  local target="$2"
  local target_dir=$INSTALL

  [ -f "$target_dir/$unit_dir/$unit" ] || exit 1
  if [ -z "$target" ] ; then
    for target in `grep '^WantedBy' $target_dir/$unit_dir/$unit | cut -f2 -d=` ; do
      if [ -n "$target" ]; then
        mkdir -p ${target_dir}/$unit_dir/${target}.wants
        ln -sf ../${unit} ${target_dir}/$unit_dir/${target}.wants/
      fi
    done
  fi
}
```

not tested but should be enough to support both WantedBy=svc1 svc2 and multi WantedBy lines.

MattDevo:
@stefansaraev yup, that works with all 3 targets

stefansaraev:
this entropy thing smells btw, I am not a crypto expert but what about CONFIG_CRYPTO_JITTERENTROPY=y in kernel config ?

MattDevo:
@stefansaraev believe we'd need to include/make use of rng_tools in order to exploit hardware rng capabilities

chewitt:
Enabling and using HW RNG features in the kernel improves the quality of randomness but doesn't solve the lack of system activity and input to RNG generation that results in a low rate of entropy accumulation and a long delay before /dev/random unlocks. This is a perfectly valid approach to low-activity.

MattDevo:
added the fix foe enable_service: https://github.com/OpenELEC/OpenELEC.tv/pull/4730

PR #4743, opened 2016-02-21
amlogic cec: major rework

gdachs:
Major rework. Nearly independent from the low level functions of the original driver. Unified for Meson 6 and Meson >=8.

To users outside the OpenELEC team:
If it doesn't work for you, check that you own a Wetek Play/Core. Currently nothing else supported.
I will close all issues that doesn't contain a link to the output of journalctl. 
If you don't know how to do that, don't create an issue. Your level of experiences would not allow me to work with you on it.


kszaq:
@gdachs With this patch everything is working great on my WeTek Play (-> LG AVR -> Samsung TV), except for the fact that turning CEC on/off requires a reboot, but that's a very minor issue.

Great job!

gdachs:
Good that it is working for you. CEC on/off is in no way influenced by my driver.

PR #4744, opened 2016-02-21
mkpkg_kodi: git clone without history

CvH:
- saves a lot bandwith + faster


stefansaraev:
mkpkg_kodi supports branches. this will break it.

CvH:
Not sure what you referring.
branches like ```mkpkg_kodi master``` or ```mkpkg_kodi Jarvis``` works
tags like ```mkpkg_kodi 16.0-Jarvis``` also work

stefansaraev:
okay right. we do not switch branches in that script. sorry for the noise. +1

PR #4758, opened 2016-02-23
kodi-language-addons: add languages

lrusak:
this adds languages to kodi so they can be used without having to initially install them.

see lrusak/service.openelec.settings@a58e77104e03c37fd300091a82a0c7a5e7d78f69

Not sure I really like this though as it is quite large, not sure about the compressed size.
```
lrusak@buildserver ~/official$ du -lh -d 1 build.OpenELEC-Generic.x86_64-7.0-devel/kodi-language-addons-ce9947c/.install_pkg
18M	build.OpenELEC-Generic.x86_64-7.0-devel/kodi-language-addons-ce9947c/.install_pkg
```


sraue:
thanks much, will test today and report back

MilhouseVH:
I thought we agreed on slack that this could be made optional? Don't really want to include every language in nightly test builds when it's only something that's needed on the very first (usually disk image) install.

sraue:
please read what i wrote in slack

MilhouseVH:
```
sraue
10:35 we could have this configurable in the options file, so you can build without this via ~/.openelec/options
```

Am I missing something? I don't see how this is optional.

gdachs:
Have thought the option is to not have a ,openelec directory, or no option file inside it

PR #4769, opened 2016-02-24
Fix wetek IR remote

### LINKED ISSUE #4759: Amlogic amremote broken after glibc-2.23
I have an Android TV stick, the UStick MK809 Plus, which I'm cooking my own OpenELEC builds for. This stick comes with the same remote controller that most of these similar sticks comes with. It's using the remotecfg utility from the amremote package to load the key config.

This has worked fine, up until the upgrade to glibc-2.23, which seems to cause this utility to not work anymore, resulting in the remote controller not functioning. I can't see anything terrible wrong in the logs, and the utility seems to load the remote config file when run. So something quite odd is happening here.

For now, as a workaround, I'm using a prebuilt static linked remotecfg binary from the stock Android rom that came with the stick. With this the remote controller works again.

codesnake:
While this pull request fixes #4759, it may be an issue in glibc 2.23 - see https://github.com/codesnake/amremote/commit/1ecd52a4d6b02d2dfe276f3f688392171257f08f


codesnake:
@sraue You are so fast. I was going to fix the commit message "Update excess PKG_URL variable in package.mk" to "Remove excess PKG_URL variable in package.mk" and see the PR is already merged :).

sraue:
sorry, just seen and wanted a fix :-)
thanks for your work

PR #4778, opened 2016-02-25
Add all mtools during host install step.

aluft:
Add all mtools during host install step.
Especially mattrib which is used by syslinux.mtools during make image

When absent, this error is observed:
make image

> image: installing extlinux to part1...
> sh: mattrib: command not found
> syslinux.mtools: warning: failed to set system bit on ldlinux.sys
> sh: mattrib: command not found
> syslinux.mtools: warning: failed to set system bit on ldlinux.c32
> image: copying files to part1...

### DIFF
--- a/packages/tools/mtools/package.mk
+++ b/packages/tools/mtools/package.mk
@@ -36,5 +36,5 @@ makeinstall_host() {
   $STRIP mtools
 
   mkdir -p $ROOT/$TOOLCHAIN/sbin
-  cp -P mtools mformat mcopy mmd $ROOT/$TOOLCHAIN/sbin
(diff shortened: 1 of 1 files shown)


aluft:
As requested last commit removes makeinstall_host which results in the mtools being installed into toolchain/bin instead of toolchain/sbin, both are located by the make image that was causing issues.

vpeter4:
So with this commit make image works without errors?

aluft:
yes, here is what I did:
removed the toolchain symbolic links from sbin
scripts/clean mtools
scripts/build mtools:host
make image
Then installed the generated img.gz to USB device and flashed to a clean emmc based device

vpeter4:
Thank you.

PR #4786, opened 2016-02-27
Oe6 update

sraue:


MilhouseVH:
It's a freaking maintenance release! :(

:-1:

lrusak:
I'd rather not :-1:

sraue:
look at the commits, 1/3 buildsystem fixes which dont harm the build, 1/3 package updates (minors and security fixes) and aml cec support... what you dont like? (most packages in sync with master, which already runs for many users without known complains)

MilhouseVH:
Did you ever think about communicating this first with your team members?

chewitt:
-100 from me .. not that it will make any difference

sraue:
yes everytime... bugfix releases contains bugfix (x.y.Z) package updates, this we do since 6 years this way

MilhouseVH:
Perhaps if it wasn't so last minute it would be easier to digest, the fact you don't communicate with anyone in the project and then drop this motherload of changes just makes it easier to walk away from this cluster fuck.

MilhouseVH:
Are you censoring github now? Special.

PR #4787, opened 2016-02-27
add generic aarch64 support

lrusak:
This should get most of the relevant packages to build for aarch64 and will be helpful when to upstream this while I work on the Odroid C2

I have some other amlogic specific aarch64 patches in another branch.

@sraue take a look at the config/arch.arm64 file and see if you are ok with this.

Not sure if the `TARGET_CPU_FLAGS="+crc+crypto"` line is really needed but it did no harm for me while building.
https://gcc.gnu.org/onlinedocs/gcc/AArch64-Options.html

maybe @codesnake can have a look too.

### DIFF
--- a/config/arch.aarch64
+++ b/config/arch.aarch64
@@ -1,0 +1,43 @@
+# determines TARGET_CPU, if not forced by user
+  if [ -z "$TARGET_CPU" ]; then
+    TARGET_CPU=cortex-a53
+  fi
+
+# TARGET_CPU:
+# arm2 arm250 arm3 arm6 arm60 arm600 arm610 arm620 arm7 arm7m arm7d
+# arm7dm arm7di arm7dmi arm70 arm700 arm700i arm710 arm710c arm7100
+# arm720 arm7500 arm7500fe arm7tdmi arm7tdmi-s arm710t arm720t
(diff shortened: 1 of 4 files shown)


codesnake:
It is almost the same as I have for WeTek Play+ :). So tested and works. Though I didn't use "+crc+crypto" options. I think it doesn't make any serious difference.

stefansaraev:
ARCH should be aarch64, not arm64.

codesnake:
@stefansaraev In kernel it is arm64

stefansaraev:
I know, I set kernel arch in config/arch.*

afaik arm64 is specific to apple, aarch64 is the name that you use for non apple hardware.

lrusak:
The problem with that is that the kernel doesn't know aarch64, so when we adjust the arch in `linux/package.mk` to be TARGET_ARCH, then it fails. I'm not sure the best solution, otherwise just using arm64 works fine.

stefansaraev:
https://github.com/stefansaraev/TB/commit/3670892b904460497a9bebbdd154e2a377db22dd

sraue:
@lrusak, ready so far or needs to be changed after #4797 ?

lrusak:
It's good to go, I already updated it

PR #4815, opened 2016-03-04
libdvbcsa: improved version

glenvt18:
Hi folks.
I've made a [version](https://github.com/glenvt18/libdvbcsa) of [libdvbcsa](http://www.videolan.org/developers/libdvbcsa.html) which is 2-3x faster than the original from videolan on both ARM and x86_64 platforms. Unfortunately, I couldn't get it merged upstream. But this version is well tested and maintained. If this is OK with you, you can merge this PR. I think this is better than keeping a batch of patches to the original. It would allow to process high bit rate TS streams on RPi2 and Atoms. There is also a AArch64/NEON dev [branch](https://github.com/glenvt18/libdvbcsa/commits/arm64) looking for testers.

### DIFF
--- a/packages/multimedia/libdvbcsa/package.mk
+++ b/packages/multimedia/libdvbcsa/package.mk
@@ -17,12 +17,12 @@
 ################################################################################
 
 PKG_NAME="libdvbcsa"
-PKG_VERSION="1.1.0"
+PKG_VERSION="f988715"
 PKG_REV="1"
 PKG_ARCH="any"
 PKG_LICENSE="LGPL"
-PKG_SITE="http://www.videolan.org/developers/libdvbcsa.html"
-PKG_URL="http://download.videolan.org/pub/videolan/libdvbcsa/${PKG_VERSION}/libdvbcsa-${PKG_VERSION}.tar.gz"
+PKG_SITE="https://github.com/glenvt18/libdvbcsa"
+PKG_URL="$DISTRO_SRC/$PKG_NAME-$PKG_VERSION.tar.xz"
 PKG_DEPENDS_TARGET="toolchain"
 PKG_PRIORITY="optional"
 PKG_SECTION="lib"
@@ -33,3 +33,16 @@ PKG_IS_ADDON="no"
 PKG_AUTORECONF="yes"
 
 PKG_CONFIGURE_OPTS_TARGET="--disable-shared --enable-static --with-sysroot=$SYSROOT_PREFIX"
+
+if echo "$TARGET_FPU" | grep -q '^neon'; then
+  PKG_CONFIGURE_OPTS_TARGET="$PKG_CONFIGURE_OPTS_TARGET --enable-neon"
+elif [ "$TARGET_ARCH" = x86_64  ]; then
+  if echo "$PROJECT_CFLAGS" | grep -q '\-mssse3'; then
+    PKG_CONFIGURE_OPTS_TARGET="$PKG_CONFIGURE_OPTS_TARGET --enable-ssse3"
+  elif echo "$PROJECT_CFLAGS" | grep -q '\-msse2'; then
+    PKG_CONFIGURE_OPTS_TARGET="$PKG_CONFIGURE_OPTS_TARGET --enable-sse2"
+  else
+    PKG_CONFIGURE_OPTS_TARGET="$PKG_CONFIGURE_OPTS_TARGET --enable-uint64"
+  fi
+fi
+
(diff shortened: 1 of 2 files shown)


chewitt:
How do we know that you're going to track upstream and release updates reliably? The optimisations are sounding interesting but we'd prefer to patch upstream original code so that if you disappear and a future version breaks something we simply remove your patches and still have something that works - albeit less optimised. What was the upstream objection to merge?

glenvt18:
@chewitt I understand your worries. That's the reason why I haven't submitted this for a while. As for the talks with upstream, you can find it on the mailling list, for example [here](http://vlc-devel.videolan.narkive.com/iRxTs4C9/patch-00-16-libdvbcsa-improve-performance) is a nice view. As you can see there were no serious objections and half of the patches were not reviewed at all. My guess is that this project is nearly abandoned - there is no maintainer, no mailling list (as you see I posted to vlc player's mailing list - there is no videolan thing for about 5 years), and it's a risky thing for vlc guys to merge 16 commits to a side project without diving deeply into it. So, for now there two alternatives: either stick to the almost frozen original version which even doesn't support SIMD well (10x slower with sse2!) or use this one. If I disappear (and remove my github account), you will still have your copy at sources.openelec.tv or you can revert this commit and switch to the videolan's one - it's like removing patches. BTW, besides "release" version of the videolan's library there is also an svn one (which I forked from). This one is even slower. Having a slow CSA library on a RPi2 or, even Atom makes vdr-addon with sw descrambler processing heavy HD streams useless - video playback smoothness suffers a lot. Not merging this PR, I would rather stick to FFdecsa (at least on x86_64), which is the origin of libdvbcsa. About patching. It can be done easily, though there will be a lot of them. It's very unlikely you will ever see another videolan's version. But I don't see much sense in doing so - there will be more patches. And if I break something, reverting commits is as easy as removing patches. This kind of software is hard to write, but quite easy to test - just run tests (in fact, they are run after each change, on every target). If you have it in a broken state, it's very hard (if ever possible) to find the bug. And, as you can see I added aarch64 support to my fork several hours ago (and I don't even have a RPi3 yet), so this fork is really maintained. In a sense, it's the "upstream" de-facto. But, again, if your policy is strongly against using forks than are not "official" or upstream, nothing can be done.

chewitt:
@glenvt18 thanks for the detailed explanation. It will be discussed.

sraue:
added with https://github.com/OpenELEC/OpenELEC.tv/pull/4840 can you test?

glenvt18:
@sraue I've built your `package-updates` branch for x86_64 and RPi2, tested and benchmarked. Everything is fine except one issue: a LTO (`-flto`) related bug with gcc 5.3 for x86_64 which makes `testenc` and `testdec` tests fail. With gcc 4.9 and OpenELEC 6.0 they pass. Legacy (videolan's) libdvbcsa is also affected by this bug. Those tests are not related to the parallel mode of operation, which is used in OpenELEC, and `testbitslice` (the most important test) passes. I also built and tested vdr_addon - it works. So, the library itself is OK. I'll look into this issue (it's really weird!), but at this moment I need your advice. This library itself uses heavy optimization options and they are enough. I don't like those internal inline functions (and it looks like the bug is connected to one of them) spreading to the application code. So, maybe it's a good idea to clean CFLAGS up before `./configure`  removing `-flto`, and some other options? What do you think of it? BTW, benchmarks show that without `-flto` it's 1-2% faster.

sraue:
building with LTO disabled in https://github.com/OpenELEC/OpenELEC.tv/commit/75d95e8a25069c9f8799a643c33c29bdb8192e9a

glenvt18:
@sraue Everything works fine now.

stefansaraev:
just throwing my 2 cents here.

I dont mind having some "non-upstream" libdvbcsa patches, as long everything works. it is nice to have everything upsteam, but I understand it's not always possible.

even if the author of those patches suddenly disappears and/or does not have time to maintain his work, the risk for openelec is low. libdvbcsa seems to be in "maintenance mode" and I really dont expect breaking changes that would require rebasing everything, even if that happens, those patches can be dropped, and we still have them in git history if someone else wants to pick and rebase later.

ffdecsa is a no-go, libdvbcsa is the only thing we have now. it is needed for vdr/dvbapi, tvheadend, and one or two unofficial addons.

sraue:
added in https://github.com/OpenELEC/OpenELEC.tv/commit/c37c22265b6a151c086e7ae691da70b68386297a

PR #4817, opened 2016-03-04
Enable SSL for MySQL

sopparus:
Tested both with SSL and without, it works fine.
Kodi on windows and other distros have for some time now being able to use ssl with mysql but not OpenELEC.

Thanks!


sopparus:
ok, done

PR #4818, opened 2016-03-04
Various $HOME/.openelec (vendor) fixes

zalaare:
Allow the use of $HOME/.openelec for all packages, settings, etc.
This fixes issues with the linux package, and allows custom u-boot.

### DIFF
--- a/config/functions
+++ b/config/functions
@@ -55,7 +55,11 @@ setup_toolchain() {
 }
 
 kernel_path() {
-  if [ -e $ROOT/projects/$PROJECT/packages/linux/package.mk ]; then
+  if [ -e $HOME/.openelec/projects/$PROJECT/packages/linux/package.mk ]; then
+    . $HOME/.openelec/projects/$PROJECT/packages/linux/package.mk
+  elif [ -e $HOME/.openelec/packages/linux/package.mk ]; then
+    . $HOME/.openelec/packages/linux/package.mk
+  elif [ -e $ROOT/projects/$PROJECT/packages/linux/package.mk ]; then
     . $ROOT/projects/$PROJECT/packages/linux/package.mk
   else
     . $ROOT/packages/linux/package.mk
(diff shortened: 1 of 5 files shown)


sraue:
please squash all 3

zalaare:
Squashed 'em

lrusak:
Why do we want to allow packages to sit outside of the OE tree? This basically makes it so people don't have to make commits anymore (which is bad for contributing to upstream aka us).

zalaare:
For what its worth:

I have been told that any device that supports "natively" an "official" kodi package will not be accepted upstream.  I don't know if this statement was true at one point and no longer is, was never true, or is still true, however this seems like a pretty good standpoint and would keep most "devices" from constantly being asked for as "official" support.  This means in my case that all the HardKernel Odroid flavors will not be accepted upstream as they all run Android natively and or come with Ubuntu (native kodi packages are available for both OS).

No one should expect the OpenELEC team to ever support the hundreds of devices that could develop were this standpoint rendered moot; I know I wouldn't expect someone to provide patches and support for devices they don't have or use anyway.  Not to mention that these devices constantly run custom Linux kernels, and custom U-Boot setups.

I personally did not create this new $HOME/.openelec approach, but I immediately see a benefit for people like me who cannot code very well, but can figure out the minimum needs to get an OpenELEC build running on a new device.  This allows me to build OpenELEC for my needs without wasting upstream resources better spent elsewhere.  If a "project" seems stable and successful, then mv $HOME/.openelec/projects/$PROJECT $PROJECT_DIR/$PROJECT seems pretty easy to commit down the road.

Just my 2 cents.

lrusak:
This is exactly what I would like to avoid. I'd rather not see so many forks.

I'd rather see people PR'ing against the upstream repository so we can have as many people working on the same project as possible.

I'd like to see more projects in the main OpenELEC tree (even if we don't officially support them), however, this isn't solely my decision to make. I'm not saying that every device needs to be supported (because that's impossible) but I'd like to seem ore supported with the possibility of becoming "official" builds.

If people keep all their changes to themselves this hurts the project as a whole and honestly with git it is trivial to maintain a branch that tracks upstream (see my wayland, retroplayer, odroid branches).

Yes things become a little messy when there are 10 different kernel versions and u-boot versions, but it's better than trying to search around for the right fork when everything can be had in one repository.

my 2 cents.

PR #4828, opened 2016-03-05
linux: patch to fix EDID parsing when ELD data spans multiple blocks

MattDevo:
Signed-off-by: Matt DeVillier <matt.devillier@gmail.com>


MattDevo:
patch taken from and addresses bug from https://bugs.freedesktop.org/show_bug.cgi?id=94197

Issue #4703, opened 2016-02-11
fix Python 2.7.10


sraue:
https://github.com/OpenELEC/OpenELEC.tv/pull/4500#issuecomment-169562986 showed Python is broken after the update to 2.7.10. even if this works in most cases there is still a issue with including host *flags.
we have 2 choices, fix or revert to python 2.7.3 (fixing preferred)

lrusak:
Can you describe the issue better? I tried looking into this today but couldn't find anything wrong.

sraue:
fixed in https://github.com/OpenELEC/OpenELEC.tv/pull/4833

PR #4833, opened 2016-03-07
Python fixes

sraue:
some fixes and cleanups (and using patches from buildroot) see also https://github.com/OpenELEC/OpenELEC.tv/issues/4703


sraue:
@MilhouseVH please include in your next build and wantch for python problems (i dont expect any). @lrusak we should rebuild all python related addons and packages and test at least for build issues. for this we need to watch the compile output (wrong -I / -L dirs in host and target builds)

sraue:
this whole should fix https://github.com/OpenELEC/OpenELEC.tv/issues/4703 does some cleanups and removes some leftovers

stefansaraev:
thanks. this fixes PIL after python 2.7.1x for me

PR #4836, opened 2016-03-08
Yet more $HOME/.openelec fixes

zalaare:
I did a grep on the packages, scripts and configs looking for PROJECT_DIR and added  the appropriate $HOME/.openelec additions where none existed yet.  This should bring the $HOME/.openelec approach to feature parity with $PROJECT_DIR.

I tested x86_64 Generic w/o any $HOME/.openelec to make sure nothing broke.


MilhouseVH:
Wouldn't it be easier to add a helper function to `function` that returns the correct package directory for a file?

eg.
```
getpackagedir()
{
  if [ -f $PROJECT_DIR/$PROJECT/$PKG_NAME/$1 ]; then
    echo "$PROJECT_DIR/$PROJECT/$PKG_NAME"
  elif [ -f $HOME/.openelec/projects/$PROJECT/$PKG_NAME/$1 ]; then
    echo "$HOME/.openelec/projects/$PROJECT/$PKG_NAME"
  elif [ -f $HOME/.openelec/$PKG_NAME/$1 ]; then
    echo "$HOME/.openelec/$PKG_NAME"
  else
    echo "$PKG_DIR/config"
  fi
}
```
then in each `package.mk`, rather than have this:
```
  if [ -f $PROJECT_DIR/$PROJECT/$PKG_NAME/$PKG_NAME.$TARGET_ARCH.conf ]; then
    KERNEL_CFG_FILE=$PROJECT_DIR/$PROJECT/$PKG_NAME/$PKG_NAME.$TARGET_ARCH.conf
  elif [ -f $HOME/.openelec/projects/$PROJECT/$PKG_NAME/$PKG_NAME.$TARGET_ARCH.conf ]; then
   KERNEL_CFG_FILE=$HOME/.openelec/projects/$PROJECT/$PKG_NAME/$PKG_NAME.$TARGET_ARCH.conf
  elif [ -f $HOME/.openelec/$PKG_NAME/$PKG_NAME.$TARGET_ARCH.conf ]; then
   KERNEL_CFG_FILE=$HOME/.openelec/$PKG_NAME/$PKG_NAME.$TARGET_ARCH.conf
  else
    KERNEL_CFG_FILE=$PKG_DIR/config/$PKG_NAME.$TARGET_ARCH.conf
  fi
```
you just have:
```
  KERNEL_CFG_FILE="($getpackagedir() "$PKG_NAME.$TARGET_ARCH.conf")"
```

zalaare:
Absolutely.  I didn't want to get too "creative" since these are my first PR's.  I'll change it as you suggest.

MilhouseVH:
I'm not going to have the final say, and @sraue may well disagree, but given the choice of making ugly code even uglier, I'd go with "creative". :)

gdachs:
I think the idea of this .openelec dir was not to make it easier to allow forks to coexist with the upstream OpenELEC tree for all times. It was made to allow OpenELEC contributors (like me ;)) to easy generate pull request to OpenELEC, but still have the chance to make their own images that contains other skins,
special addons and maybe special remote mappings, without having them to exclude from the pull requests always.

sraue:
i agree wirh @MilhouseVH an dont disagree :-)

zalaare:
I didn't squash these because I thought it would be easier to see if I missed something.  I won't lie and say I setup a test-case scenario for every different file location throughout every possible project directory location, but I did test a few along the way.  I rebuilt x86_64 Generic w/o $HOME/.openelec to again verify nothing broke during the build process.

I got "creative" with the function but had to decide where to stop.  I could have used @sraue example and simply returned the directory path, but I opted to return the path + the file in question.  This is a little easier to use in my opinion, but if you note in [plymouth-lite](https://github.com/zalaare/OpenELEC.tv/blob/eb379b2d5df184fe069395c3eb4f00514172ada9/packages/tools/plymouth-lite/package.mk#L52) I ended up wrapping the return in a dirname to keep parity with the original package.mk.

Please feel free to critique.  I'm not all-knowing or perfect =)

packages/linux/package.mk
```diff
@@ -55,11 +55,8 @@ if [ "$BUILD_ANDROID_BOOTIMG" = "yes" ]; then
 fi
 
 post_patch() {
-  if [ -f $PROJECT_DIR/$PROJECT/$PKG_NAME/$PKG_NAME.$TARGET_ARCH.conf ]; then
-    KERNEL_CFG_FILE=$PROJECT_DIR/$PROJECT/$PKG_NAME/$PKG_NAME.$TARGET_ARCH.conf
-  else
-    KERNEL_CFG_FILE=$PKG_DIR/config/$PKG_NAME.$TARGET_ARCH.conf
-  fi
```
zalaare: I see a potential break here already :-1: I'm looking into it.

config/functions
```diff
@@ -217,6 +209,33 @@ strip_gold() {
...
 }
 
+get_projectdir() {
+  local d_list=( # main directory roots to search
+    $PROJECT_DIR/$PROJECT/packages
+    $HOME/.openelec/projects/$PROJECT/packages
+    $HOME/.openelec/packages
+    $PROJECT_DIR/$PROJECT
+    $HOME/.openelec/projects/$PROJECT
+    $HOME/.openelec
```
zalaare: I could potentially add $PKG_DIR to the main d_list variable thereby enabling searching inside the original PKG_DIR and possible removing the need to do an else.  Not sure if this makes it more confusing or not.

packages/linux/package.mk
```diff
@@ -55,11 +55,8 @@ if [ "$BUILD_ANDROID_BOOTIMG" = "yes" ]; then
...
+
+  KERNEL_CFG_FILE="$(get_projectdir $PKG_NAME.$TARGET_ARCH.conf)"
```
zalaare: I see a potential break here already :-1: I'm looking into it.

lrusak:
Please no. This is not needed at all :-1:

zalaare:
@MilhouseVH What I'm testing right now is similar.  I'm stripping everything before the slash and searching for that as well.  So busybox/busybox-target.conf search for that AND busybox-target.conf in the subdirectories.  This finds $PKG_DIR/config/busybox-target.conf this way.

so `BUSYBOX_CFG_FILE_TARGET=$(get_projectdir busybox/busybox-target.conf)`

Search order becomes:
$PROJECT_DIR/$PROJECT/packages/busybox/busybox/busybox-target.conf
$PROJECT_DIR/$PROJECT/packages/busybox/busybox-target.conf
$PROJECT_DIR/$PROJECT/packages/busybox/config/busybox/busybox-target.conf
$PROJECT_DIR/$PROJECT/packages/busybox/config/busybox-target.conf
$PROJECT_DIR/$PROJECT/packages/busybox/scripts/busybox/busybox-target.conf
$PROJECT_DIR/$PROJECT/packages/busybox/scripts/busybox-target.conf
$PROJECT_DIR/$PROJECT/packages/config/busybox/busybox-target.conf
$PROJECT_DIR/$PROJECT/packages/config/busybox-target.conf
$PROJECT_DIR/$PROJECT/packages/scripts/busybox/busybox-target.conf
$PROJECT_DIR/$PROJECT/packages/scripts/busybox-target.conf
$PROJECT_DIR/$PROJECT/packages/./busybox/busybox-target.conf
$PROJECT_DIR/$PROJECT/packages/./busybox-target.conf
$PROJECT_DIR/$PROJECT/busybox/busybox/busybox-target.conf
$PROJECT_DIR/$PROJECT/busybox/busybox-target.conf
$PROJECT_DIR/$PROJECT/busybox/config/busybox/busybox-target.conf
$PROJECT_DIR/$PROJECT/busybox/config/busybox-target.conf
$PROJECT_DIR/$PROJECT/busybox/scripts/busybox/busybox-target.conf
$PROJECT_DIR/$PROJECT/busybox/scripts/busybox-target.conf
$PROJECT_DIR/$PROJECT/config/busybox/busybox-target.conf
$PROJECT_DIR/$PROJECT/config/busybox-target.conf
$PROJECT_DIR/$PROJECT/scripts/busybox/busybox-target.conf
$PROJECT_DIR/$PROJECT/scripts/busybox-target.conf
$PROJECT_DIR/$PROJECT/./busybox/busybox-target.conf
$PROJECT_DIR/$PROJECT/./busybox-target.conf
$HOME/.openelec/projects/$PROJECT/packages/busybox/busybox/busybox-target.conf
$HOME/.openelec/projects/$PROJECT/packages/busybox/busybox-target.conf
$HOME/.openelec/projects/$PROJECT/packages/busybox/config/busybox/busybox-target.conf
$HOME/.openelec/projects/$PROJECT/packages/busybox/config/busybox-target.conf
$HOME/.openelec/projects/$PROJECT/packages/busybox/scripts/busybox/busybox-target.conf
$HOME/.openelec/projects/$PROJECT/packages/busybox/scripts/busybox-target.conf
$HOME/.openelec/projects/$PROJECT/packages/config/busybox/busybox-target.conf
$HOME/.openelec/projects/$PROJECT/packages/config/busybox-target.conf
$HOME/.openelec/projects/$PROJECT/packages/scripts/busybox/busybox-target.conf
$HOME/.openelec/projects/$PROJECT/packages/scripts/busybox-target.conf
$HOME/.openelec/projects/$PROJECT/packages/./busybox/busybox-target.conf
$HOME/.openelec/projects/$PROJECT/packages/./busybox-target.conf
$HOME/.openelec/projects/$PROJECT/busybox/busybox/busybox-target.conf
$HOME/.openelec/projects/$PROJECT/busybox/busybox-target.conf
$HOME/.openelec/projects/$PROJECT/busybox/config/busybox/busybox-target.conf
$HOME/.openelec/projects/$PROJECT/busybox/config/busybox-target.conf
$HOME/.openelec/projects/$PROJECT/busybox/scripts/busybox/busybox-target.conf
$HOME/.openelec/projects/$PROJECT/busybox/scripts/busybox-target.conf
$HOME/.openelec/projects/$PROJECT/config/busybox/busybox-target.conf
$HOME/.openelec/projects/$PROJECT/config/busybox-target.conf
$HOME/.openelec/projects/$PROJECT/scripts/busybox/busybox-target.conf
$HOME/.openelec/projects/$PROJECT/scripts/busybox-target.conf
$HOME/.openelec/projects/$PROJECT/./busybox/busybox-target.conf
$HOME/.openelec/projects/$PROJECT/./busybox-target.conf
$HOME/.openelec/packages/busybox/busybox/busybox-target.conf
$HOME/.openelec/packages/busybox/busybox-target.conf
$HOME/.openelec/packages/busybox/config/busybox/busybox-target.conf
$HOME/.openelec/packages/busybox/config/busybox-target.conf
$HOME/.openelec/packages/busybox/scripts/busybox/busybox-target.conf
$HOME/.openelec/packages/busybox/scripts/busybox-target.conf
$HOME/.openelec/packages/config/busybox/busybox-target.conf
$HOME/.openelec/packages/config/busybox-target.conf
$HOME/.openelec/packages/scripts/busybox/busybox-target.conf
$HOME/.openelec/packages/scripts/busybox-target.conf
$HOME/.openelec/packages/./busybox/busybox-target.conf
$HOME/.openelec/packages/./busybox-target.conf
$HOME/.openelec/busybox/busybox/busybox-target.conf
$HOME/.openelec/busybox/busybox-target.conf
$HOME/.openelec/busybox/config/busybox/busybox-target.conf
$HOME/.openelec/busybox/config/busybox-target.conf
$HOME/.openelec/busybox/scripts/busybox/busybox-target.conf
$HOME/.openelec/busybox/scripts/busybox-target.conf
$HOME/.openelec/config/busybox/busybox-target.conf
$HOME/.openelec/config/busybox-target.conf
$HOME/.openelec/scripts/busybox/busybox-target.conf
$HOME/.openelec/scripts/busybox-target.conf
$HOME/.openelec/./busybox/busybox-target.conf
$HOME/.openelec/./busybox-target.conf

MilhouseVH:
Yes, but those are all `$PROJECT` and `$HOME` directories, and if your file isn't found in those locations then you need to perform an `else` which uses a default file (in this case, a `$PKG_DIR` file).

By passing a "default" filename to the function (the `$PKG_DIR` filename) you avoid calling the function twice, and the code in `package.mk` is a bit cleaner.

That said, putting more functionality into the function makes the code in `package.mk` less clear so... it's just a suggestion.

zalaare:
In my test I added $PKG_DIR as the last place to look in for the file in question.  So if it doesn't find busybox-target.conf anywhere before $PKG_DIR/config, then it will still get it from $PKG_DIR/config/busybox-target.conf.  The only time an if is is needed is if its possible that there is NO file to be found.  For example [xorg-server/package.mk](https://github.com/zalaare/OpenELEC.tv/blob/VendorFixes/packages/x11/xserver/xorg-server/package.mk#L168-L170).

MilhouseVH:
I'd still allow for a default file when all else fails, but that's just me.

zalaare:
I can do either way.  Adding PKG_DIR to the search makes the name even more odd
`get_projectdir`
`get_project_file`
`get_file`

MilhouseVH:
The more directories you add to this, the more risk there is of inadvertently changing existing behaviour. Adding more directories is also not really an alternative to a default as the filename of the default file may not be the same as the file you are searching for.

Maybe call the function `find_a_file()`. :)

zalaare:
The more I think about this, the more I agree with you @MilhouseVH.  Less is more.  I'd like to keep toying with this function to limit the number of directories that are actually searched.  Adding a fallback makes it feel safer all around.

My last test showed 72 potential locations searched (though many are bogus and would never exist, it's still a waste of cpu time to look for them there)

zalaare:
I've implemented @MilhouseVH approach and limited the amount that the function searches through.  This seems much nicer than my first attempt.

I've squashed it up right nice.

config/functions
```diff
@@ -217,6 +201,41 @@ strip_gold() {
...
+    $HOME/.openelec/projects/$PROJECT/packages
+    $HOME/.openelec/projects/$PROJECT
+    $HOME/.openelec/packages
+    $HOME/.openelec
+  )
+  local s_list=( # subdirectories to search
+    $PKG_NAME
+    .
+  )
+  local result
```
MilhouseVH: `local result dir sdir`

config/functions
```diff
@@ -217,6 +201,41 @@ strip_gold() {
...
+  for dir in ${d_list[@]} ; do
+    for sdir in ${s_list[@]} ; do
+      if [ -f $dir/$sdir/$1 ]; then
+        result="$dir/$sdir/$1"
+        break 2
+      fi
+    done
+  done
+  if [ -n "$2" ]; then
```
MilhouseVH: Change to `if [ -z "$result" -a -n "$2" ]; then`

and you don't need to check any further for `-z $result` in lines 229 and 231, and line 231 can be a simple `else` since you know that `$2` won't exist.

sraue:
before going further:

*this whole* seems to me goes a bit to far, we can already use packages in ~/.openelec, so any special or own packages goes there. our default packages should not respect configs from this dir and we should not maintain this. if anyone needs packages there he should copy and the default package from packages/ to ~/.openelec/packages and adapt if needed the package there. a mix of packages with and without package.mk we should not support.
in the end any package.mk in $ROOT/packages should ideally not have references to ~/.openelec (and even $ROOT/projects/$PROJECT). this has to be done only in scripts/ and config/ while searching for $PKG_DIR

sraue:
even something like https://github.com/zalaare/OpenELEC.tv/blob/VendorFixes/packages/x11/xserver/xorg-server/package.mk#L150-L155 is not really clean and what i really want, there should be another solution... but this is nothing to discuss in this PR

PR #4840, opened 2016-03-11
Package updates

sraue:


MilhouseVH:
You wouldn't need the additional mkpkg's if you bothered to consider #4537.

sraue:
i added this mainly for packages where its missing actually (except libdvbcsa)

PR #4871, opened 2016-03-23
Openelec 6.0.1

Fosphatic:
I have beside wetek play as well an Tronsmart Vega S89-H device and was wondering if it's possible to get the latest update of openelec 6.0.1 as well for this S82 alike box with S802 cpu. The performance of this box is amazing but I have some issues with high bitrate mkv files what wetek play deffo plays without issues.


sraue:
something was wrong here....

Issue #4873, opened 2016-03-24
Kodi compile error with KODI_BLURAY_SUPPORT="no"


hoblins:
Hello,
with setting
`# build and install with BluRay support (yes / no)
  KODI_BLURAY_SUPPORT="yes"`

Kodi is not compiled. Error:

`In file included from DVDDemuxers/DVDDemuxStreamSSIF.h:24:0,
                 from DVDDemuxers/DVDDemuxFFmpeg.h:24,
                 from DVDFileInfo.cpp:43:
../OpenELEC.tv/build.OpenELEC-RPi2.arm-8.0-devel/kodi-67cf03a/xbmc/cores/VideoPlayer/DVDInputStreams/DVDInputStreamBluray.h:29:30: fatal error: libbluray/bluray.h: No such file or directory
compilation terminated.
../OpenELEC.tv/build.OpenELEC-RPi2.arm-8.0-devel/kodi-67cf03a/Makefile.include:97: recipe for target 'DVDFileInfo.o' failed
make[2]: *** [DVDFileInfo.o] Error 1`

Branch: current master (commit 51c9445)

sraue:
can you please check again with latest master, i dont see the error (anymore)

hoblins:
The error still occurs.
`In file included from DVDDemuxStreamSSIF.h:24:0,
                 from DVDDemuxStreamSSIF.cpp:21:
/home/OpenELEC.tv/build.OpenELEC-RPi2.arm-8.0-devel/kodi-67cf03a/xbmc/cores/VideoPlayer/DVDInputStreams/DVDInputStreamBluray.h:29:30: fatal error: libbluray/bluray.h: No such file or directory
compilation terminated.`
I didn't go to much into details, but the file
`/home/OpenELEC.tv/build.OpenELEC-RPi2.arm-8.0-devel/toolchain/armv7ve-openelec-linux-gnueabi/sysroot/usr/include/libbluray/bluray.h` is missing, if libbluray was not compiled at least once. Kodi depends on that package either with disabled blu ray support.
E.g. adding `PKG_DEPENDS_TARGET="$PKG_DEPENDS_TARGET libbluray"` to line 131
[https://github.com/OpenELEC/OpenELEC.tv/blob/master/packages/mediacenter/kodi/package.mk#L131](https://github.com/OpenELEC/OpenELEC.tv/blob/master/packages/mediacenter/kodi/package.mk#L131) works.

[...]

PR #4889, opened 2016-03-30
#4873 Compile errors with missing libbluray

hoblins:


sraue:
this makes no sense to build with libbluray if bluray support is disabled, i will check this later, there must be a fix in kodi

hoblins:
Of course you're right. That was too quick and dirty.
The flags `--disable-libbluray` and `--enable-libbluray` seem not to be used in OpenELEC.

PR #4897, opened 2016-04-02
device type for cec is now Player instead of Recorder. This way it is…

gdachs:
… matching with the announcement of u-boot. Thanks to kszaq!

### DIFF
--- a/projects/WeTek_Core/patches/kodi/0200-use-cec-device-type-player.patch
+++ b/projects/WeTek_Core/patches/kodi/0200-use-cec-device-type-player.patch
@@ -1,0 +1,12 @@
+diff -Nur a/system/peripherals.xml b/system/peripherals.xml
+--- a/system/peripherals.xml	2016-03-21 16:42:19.000000000 +0100
++++ b/system/peripherals.xml	2016-04-01 22:57:26.253146302 +0200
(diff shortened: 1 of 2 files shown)


kszaq:
I don't think this should be changed for WeTek Play as it doesn't have CEC enabled in u-boot and if we keep "Player" device type, more buttons can be sent from TV's remote, e.g. numbers and CH-/+. This is desirable considering the fact that most users use WP with DVB tuners and want to change channel numbers using TV's remote.

I don't have a WeTek Core to comment on it.

gdachs:
you mean to keep **recorder** device type on the play? Yes, I can change this back. I didn't now that it is not enabled in the play.

kszaq:
Yes, I think it's better to keep it as **recorder**. As for WeTek Core - I don't know, don't have a device. Does it offer wakeup over CEC?

gdachs:
Shit, I have thought you told me that there is a problem with cec wakeup because the device type is different between u-boot and kodi. Have forgotten who it was then. Sorry, I am old. 
But I know for sure that it was a core, I only changed it for the play because of symmetry ;)

kszaq:
It was me. :) I actually told you of the change needed for S805 devices to make your driver fully work according to my needs.

PR #4945, opened 2016-05-06
samba: move samba-defaults from systemd into the samba-config script

escalade:
Once a samba.conf has been generated, the samba-defaults unit will fail on boot and cause systemd to output to console, messing up the quiet boot. To avoid this we can create the default configuration in the samba-config script instead.


sraue:
added in https://github.com/OpenELEC/OpenELEC.tv/commit/daeeb57d1bfded1ef438dc384380278f3898a941 thank you very much!

escalade:
My first and last PR to OE as my authorship was removed.

sraue:
what for a auhorship? maybe the adress from the commiter.... but this happens if the patch must be reworked/changed - and if i must do manually like i have done and it was needed because it could not be merged from here. other then that most of the code was already 1:1 there , its only moved to match the coding style. and for this code the authorship was not you.
i feel the reason you complain is a different one.

escalade:
No need for conspiracy theories :) I simply wanted to contribute out of respect for your great product and put my small fingerprint on it. If you wanted something changed or reworked, you could simply have requested it as is normal in PR's. No biggie of course, it's a minor change, but anyway thought that was odd.

PR #4955, opened 2016-06-04
merging/updating with Openelec.tv master

ajw107:
merging/updating with Openelec.tv master


ajw107:
Sorry, meant to update my own repo, not the other way round, yikes.

PR #4959, opened 2016-06-08
fix auto-sense in lirc_rpi driver

IngoKresse:
On a Raspberry Pi 2, the lirc_rpi driver might receive spurious
interrupts and change it's low-active / high-active setting.
When this happens, the IR remote control stops working.

This patch disables this auto-detection if the 'sense' parameter
was set in the device tree, making the driver robust to such
spurious interrupts.


sraue:
included in newer RPi kernels. thanks much

PR #4974, opened 2016-10-06
CMake: patch to disable SSLv3 support, was removed in libressl curren…

jerome-benoit:
was removed in libressl current version. 


jerome-benoit:
I'm afraid it's not enough to support the latest changes in libressl, I'm still working on it.

PR #4979, opened 2016-10-21
pkg-config: source url update

jerome-benoit:
And a comestic indentation fix while at it.

### DIFF
--- a/packages/devel/pkg-config/package.mk
+++ b/packages/devel/pkg-config/package.mk
@@ -22,7 +22,7 @@ PKG_REV="1"
 PKG_ARCH="any"
 PKG_LICENSE="GPL"
 PKG_SITE="http://www.freedesktop.org/software/pkgconfig/"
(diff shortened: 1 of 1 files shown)


packages/devel/pkg-config/package.mk
```diff
@@ -38,5 +38,5 @@ PKG_CONFIGURE_OPTS_HOST="--disable-silent-rules \
 
 post_makeinstall_host() {
   mkdir -p $SYSROOT_PREFIX/usr/share/aclocal
-    cp pkg.m4 $SYSROOT_PREFIX/usr/share/aclocal
+  cp pkg.m4 $SYSROOT_PREFIX/usr/share/aclocal
```
jerome-benoit: It's a coding style convention. I simply fail to understand the other way around: either you consider there is a hierarchy between mkdir and cp for whatever reason, either you don't and you want to do the cp only if the mkdir don't fail and switch to a conditional if...fi statement around the cp. 

Anyway, it's not a very important PR, so do it as you like. 

Regards.

PR #4980, opened 2016-10-23
u-boot: properly add GCC 6.X support

jerome-benoit:


jerome-benoit:
Yes, upstream Linux kernel or u-boot add the specific bits relative to GCC each time a new stable major version is released. 

For GCC 5 and 6 there wasn't much change but in case there's an bug in GCC, it's handled here. 

Regards.

PR #4983, opened 2016-10-26
glibc: update the source url to make use of the GNU mirroiring system

jerome-benoit:
Signed-off-by: Jérôme Benoit jerome.benoit@piment-noir.org


sraue:
added in https://github.com/OpenELEC/OpenELEC.tv/commit/fa6844f2241ddd085091f0ba5badabd9ef357e1c
thanks much

PR #4984, opened 2016-10-30
LibreSSL: Fix CMake handling of pkgconfig files

jerome-benoit:
Add a patch to handle the pkgconfig files properly during
the build process

Signed-off-by: Jérôme Benoit jerome.benoit@piment-noir.org


sraue:
thanks much, will test here and report back... is it from you? it should be committed upstream

jerome-benoit:
Le 31/10/2016 à 14:34, Stephan Raue a écrit :
>
> thanks much, will test here and report back... is it from you? it
> should be committed upstream
>
>

Yep but it's not upstreamable, it must be more polished, it lack some
features that autoconf and friends offer but as a Linux only
distribution patch, it is good enough and work.

Regards,

-- 
Jérôme Benoit aka fraggle
Piment Noir - http://piment-noir.org
OpenPGP Key ID : 9FE9161D
Key fingerprint : 9CA4 0249 AF57 A35B 34B3 AC15 FAA0 CB50 9FE9 161D

PR #4996, opened 2016-12-23
Create Bluetooth pairing button via GPIO

Alphascrypt:
Create Bluetooth pairing button via GPIO. Then the it doesnt need ssh access.


lrusak:
probably not what you wanted to do

PR #5008, opened 2017-01-24
Update package.mk - http://samba.org

Andy-ABTec:
samba.org now appears to be https only...
Lines 24 and 25 of package.mk updated accordingly


Baldflamingo:
This error also appears to be present within the samba/package.mk file as well. Both ccache & samba files will need the same amendments.

Andy-ABTec:
Thanks, missed that one...

sraue:
fixed in https://github.com/OpenELEC/OpenELEC.tv/commit/30a5dcbaced4e52aede161e3b15df984387442f6, thanks much and sorry for not using your PR

PR #5009, opened 2017-01-26
Update samba package.mk

Andy-ABTec:
samba.org appears to be an https only site now, lines 24 and 25 updated accordingly


sraue:
thanks, changed in https://github.com/OpenELEC/OpenELEC.tv/commit/332e91ecb5a21d7e5cd17aa36d1dbe1884b2742e sorry for not using your PR :-)

PR #5018, opened 2017-03-02
ftp.kernel.org not working anymore

danielkucera:
see more here: https://kernel.org/shutting-down-ftp-services.html

use www.kernel.org instead


danielkucera:
Programmers are mean...

MilhouseVH:
Did you mean to close this?

PR #5042, opened 2017-05-02
gcc: add libatomic option

malaterre:
Signed-off-by: Mathieu Malaterre <malat@debian.org>


malaterre:
Should be all fixed now.

sraue:
thanks much!

PR #5048, opened 2017-05-19
ci20 mariadb

malaterre:
Fix building of mariadb on mipsel (Creator CI20)


malaterre:
I've setup my patches so that `mips` folder now correspond to setting `GCC_LIBATOMIC_SUPPORT=yes` which is a required option to compile `mariadb` anyway on `mips`.

